PR #5, opened 2020-04-21
Integrated cerberus for checking cluster health

### LINKED ISSUE #2: Integrate Cerberus to check the cluster health
Cerberus ( https://github.com/openshift-scale/cerberus ) needs to be run after each scenario specified in the config to monitor the cluster under test and the aggregated go/no-go signal generated by it needs to be consumed by Kraken to determine pass/fail i.e make sure the Kubernetes/OpenShift cluster is healthy after the failure injection, not just the targeted component.

yashashreesuresh:
Fixes #2


yashashreesuresh:
Please have a look @chaitanyaenr

run_kraken.py
```diff
@@ -38,6 +40,17 @@ def main(cfg):
...
         logging.info("\n%s%s" % (cluster_version, cluster_info))
 
+        if install_and_enable_cerberus:
+            try:
+                runcommand.invoke("podman ps | grep -w cerberus || podman ps -a | grep -w cerberus")
+                runcommand.invoke("podman rm -f cerberus")
+            except:
+                pass
+            runcommand.invoke("podman pull quay.io/openshift-scale/cerberus")
+            runcommand.invoke("podman run --name=cerberus --net=host -v ~/.kube/config:/root/.kube/config:Z -v ~/cerberus.yaml:/root/cerberus/config/config.yaml:Z -d quay.io/openshift-scale/cerberus:latest")
```
chaitanyaenr, line 50: The kubeconfig path can be picked from the kraken config. This will use the default cerberus config, think we need to allow the user to tweak and mount the desired config into the container. We can add a cerberus config in kraken but we need to keep updating it in kraken whenever there's a new change in cerberus repo which is not really feasible. Maybe it's better to let the user take care of installing cerberus for now? Thoughts?

run_kraken.py
```diff
@@ -46,6 +59,18 @@ def main(cfg):
                 logging.info("Scenario: %s has been successfully injected!" %(scenario))
                 logging.info("Waiting for the specified duration: %s" %(wait_duration))
                 time.sleep(wait_duration) 
+                if enable_cerberus or install_and_enable_cerberus:
+                    if enable_cerberus:
+                        http_server_cerberus = config["cerberus"]["http_server_cerberus"]
+                        if not http_server_cerberus:
+                            logging.error("Provide the http server url of cerberus")
```
chaitanyaenr, line 66: Maybe fail/exit when cerberus is enabled and the user doesn't provide the url?

run_kraken.py
```diff
@@ -46,6 +59,18 @@ def main(cfg):
...
+                    else:
+                        http_server_cerberus = "http://0.0.0.0:8080"
+                    cerberus_status = runcommand.invoke("curl " +  http_server_cerberus)
```
chaitanyaenr, line 69: Maybe use python requests module to get the cerberus_status?

PR #6, opened 2020-04-22
Added support to kill prometheus pods

yashashreesuresh:
Added a scenario to kill prometheus pods.

### DIFF
--- a/scenarios/post_action_prometheus.yml
+++ b/scenarios/post_action_prometheus.yml
@@ -1,0 +1,22 @@
+config:
+  runStrategy:
+    runs: 1
+    maxSecondsBetweenRuns: 10
+    minSecondsBetweenRuns: 1
+scenarios:
+  - name: "check 2 pods are in namespace with selector: prometheus"
+    steps:
+    - podAction:
+        matches:
+          - labels:
+              namespace: "openshift-monitoring"
+              selector: "app=prometheus"
+        filters:
+          - property:
+              name: "state"
+              value: "Running"
+        # The actions will be executed in the order specified
+        actions:
+          - checkPodCount:
+              count: 2
+ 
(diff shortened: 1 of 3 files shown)


yashashreesuresh:
Please have a look @chaitanyaenr

paigerube14:
I would add this scenario to the config file so it gets run by default, but everything else looks good to me!

scenarios/prometheus.yml
```diff
@@ -0,0 +1,23 @@
+config:
```
chaitanyaenr, line 1: Maybe also add a post action to check the pod count i.e 2 similar to https://github.com/openshift-scale/kraken/blob/823cecad6e801d3ee86580f1bc480c7a8ba5d237/scenarios/post_action_openshift-kube-apiserver.yml#L19?

yashashreesuresh: Fixed it

scenarios/post_action_prometheus.yml
```diff
@@ -0,0 +1,22 @@
...
+          - labels:
+              namespace: "openshift-monitoring"
+              selector: "app=prometheus"
+        filters:
+          - property:
+              name: "state"
+              value: "Running"
+        # The actions will be executed in the order specified
+        actions:
+          - checkPodCount:
```
chaitanyaenr, line 20: Is there a way to mention a timeout for the check pod count action? cc: @paigerube14

paigerube14: Not that I can see, I will keep looking though

chaitanyaenr: We might want to open an issue in powerfulseal to add support to specify the retries as this will avoid maintaining two configs for the same scenario. Kill can be the first action and checkPodCount can be the next action with retries in place. Thoughts?

chaitanyaenr:
Merging this as we can revisit it after the issue that @paigerube14 opened ( https://github.com/bloomberg/powerfulseal/issues/295 ) to add retries logic to avoid having two configs to inject failures and check if the targeted component recovered. Thanks.

Issue #8, opened 2020-04-28
Node chaos scenarios


chaitanyaenr:
It's important to check how Kubernetes/OpenShift is dealing with the node failures, the failures can be a node termination or stop/shut down. The nodes can be terminated by hitting the cloud platform API on top of which OpenShift has been deployed ( AWS, GCP, Azure e.t.c ).

Kraken can accept a config from the user with the details including role/selector of the node to kill ( Master/node-role.kubernetes.io/master, Worker/node-role.kubernetes.io/worker, Infra/node-role.kubernetes.io/infra), number of nodes to kill and wait between each kill and act accordingly.

mffiedler:
Some potential node failure scenarios:
- remove instance via cloud API
- "normal" node reboot (e.g. after applying patch for kernel CVE)
- extended (1 hour+) node shutdown for lab maintenance
- stop kubelet
- node crash (fork bomb or "echo c > /proc/sysrq-trigger")
- node SDN isolation from cluster

PR #10, opened 2020-04-30
Added node chaos scenarios

yashashreesuresh:
This commit:
- Adds a node scenario to stop and start an instance
- Adds a node scenario to terminate an instance
- Adds a node scenario to reboot an instance
- Adds a node scenario to stop the kubelet
- Adds a node scenario to crash the node

Fixes: #8


config/config.yaml
```diff
@@ -4,6 +4,15 @@ kraken:
         -    scenarios/etcd.yml
         -    scenarios/openshift-kube-apiserver.yml                           
         -    scenarios/openshift-apiserver.yml
+    node_scenarios:
+        node_start_stop_scenario:
+            enable_scenario:                               # Set to True to enable the scenario
+            selector:                                      # This enables the nodes with the specified selector to be selected for chaos injection
+            number_of_kills:                               # Number of times to inject the scenario
+        node_termination_scenario:
```
chaitanyaenr, line 12: Maybe combine the start_stop scenario and terminate into one since they both need the same details ( selector, number_of_kills )? Something like:
node_scenarios:
    action: start/stop/terminate
    selector:
    number_of_kills:

Thoughts?

config/config.yaml
```diff
@@ -4,6 +4,15 @@ kraken:
...
+            enable_scenario:
+            selector:
+            number_of_kills:
```
chaitanyaenr, line 15: Maybe rename number_of_kills to instance_count?

node_scenarios.py
```diff
@@ -0,0 +1,79 @@
...
+import time
+import json
+import boto3
+import random
+import logging
+import kraken.kubernetes.client as kubecli
+import kraken.invoke.command as runcommand
+
+
+def node_start_stop_scenario(node_scenario_config):
```
chaitanyaenr, line 11: Maybe add a package for node scenarios and import it in the main file i.e run_kraken.py? Thoughts?

node_scenarios.py
```diff
@@ -0,0 +1,79 @@
...
+            if not nodes: raise Exception("Nodes with the provided selector do not exist")
+            logging.info("Nodes with the selector %s: %s" % (selector, nodes))
+            number_of_nodes = len(nodes)
+            node = nodes[random.randint(0, number_of_nodes - 1)]
+            instance_id = client.describe_instances(Filters=[{'Name': 'private-dns-name', 'Values': [node]}])['Reservations'][0]['Instances'][0]['InstanceId']
+            logging.info("Stopping the node %s with instance ID: %s " % (node, instance_id))
+            client.stop_instances(InstanceIds=[instance_id])
+            instance.wait_until_stopped(InstanceIds=[instance_id])
+            logging.info("Node with instance ID: %s is in stopped state" % (instance_id))
+            time.sleep(30)
```
chaitanyaenr, line 31: Let's move the time out to config after changing this to poll instead of sleep. This way the user can set the time out after which the test can fail in case the node is not in the expected state.

node_scenarios.py
```diff
@@ -0,0 +1,79 @@
+import sys
+import time
+import json
+import boto3
```
chaitanyaenr, line 4: boto3 module needs to be listed as a dependency by adding it to requirements.txt file.

node_scenarios.py
```diff
@@ -0,0 +1,79 @@
...
+        except Exception as e:
+            logging.error("Failed to start node instance. Encountered following exception: %s. Test Failed" % (e))
+            logging.error("node_start_stop_scenario injection failed!")
+            sys.exit(1)
+
+def node_termination_scenario(node_scenario_config):
+    selector = node_scenario_config["selector"]
+    number_of_kills = node_scenario_config["number_of_kills"]
+
+    client = boto3.client('ec2')
```
chaitanyaenr, line 55: Let's make the boto3 client definition global as we use it in other functions related to the node scenarios as well.

node_scenarios.py
yashashreesuresh: We get boto3 automatically with AWS CLI installation.

node_scenarios.py
yashashreesuresh: Yeah, let's do that.

config/config.yaml
yashashreesuresh: instance_kill_count?

node_scenarios.py
chaitanyaenr: Maybe add awscli to the requirements.txt then?

node_scenarios/node_start_stop_scenario.yml
```diff
@@ -0,0 +1,3 @@
+label_selector:                                              # This enables the nodes with the specified label selector to be selected for chaos injection
+instance_kill_count:                                         # Number of times to inject the scenario in the nodes
+timeout:                                                     # Set the time out after which the test can fail in case the node is not in the expected state after chaos scenario injection
```
chaitanyaenr, line 3: Think the config should accept a list of node scenarios ( start, stop, restart, terminate ) similar to how powerfulseal does it - https://github.com/bloomberg/powerfulseal/blob/master/tests/policy/example_config2.yml except that we will ask user for list of actions ( start, stop, restart, terminate ) instead of ip address's/hostnames. Thoughts?

kraken/node_actions/node_scenarios.py
```diff
@@ -0,0 +1,117 @@
+import os
+import sys
+import time
+import json
+import yaml
+import boto3
```
mffiedler, line 6: Can we create an abstract node scenarios class and have cloud specific implementing classes for it that use the cloud specific libraries like boto3 (e.g. aws_node_scenarios)?  The config can contain an option for the cloud type to use and the client can just deal with generic scenarios like start/stop, reboot, termination.

yashashreesuresh: Yeah, this sounds good. I will make the changes to implement the same. Thank you for the feedback! @mffiedler

kraken/node_actions/aws_node_scenarios.py
```diff
@@ -0,0 +1,177 @@
...
+                self.aws.wait_until_stopped(instance_id)
+                logging.info("Node with instance ID: %s is in stopped state" % (instance_id))
+                wait_for_unknown_status(node, self.timeout)
+            except Exception as e:
+                logging.error("Failed to stop node instance. Encountered following exception: %s. Test Failed" % (e))
+                logging.error("node_stop_scenario injection failed!")
+                sys.exit(1)
+
+
+    def node_start_stop_scenario(self):
```
chaitanyaenr, line 112: The node_start_scenario and stop functions can be reused by passing the node_name/selector to act on instead of repeating the logic. Thoughts?

node_scenarios/node_scenarios_example.yml
```diff
@@ -0,0 +1,16 @@
...
+    node_name:
+    label_selector: node-role.kubernetes.io/worker
+    instance_kill_count: 1
+    timeout: 120
+    cloud_type: aws
+  - actions:
+    - node_reboot_scenario
+    node_name:
+    label_selector: node-role.kubernetes.io/infra
+    instance_kill_count: 1
```
chaitanyaenr, line 14: Maybe rename instance_kill_count to instance_count since the action can also be a reboot/stop?

node_scenarios/node_scenarios_example.yml
```diff
@@ -0,0 +1,16 @@
+node_scenarios:
+  - actions:
```
paigerube14, line 2: It might be nice to add a name to each node scenario here, similar to pod kill scenarios

yashashreesuresh: The items under `actions` refer to the node scenarios to be applied on a selected node.

paigerube14: Sorry, I meant above that 

node_scenarios:
  - name: Stop and Start Scenario
    actions: 
    - node_stop_scenario
    - node_start_scenario
    node_name:
    label_selector: node-role.kubernetes.io/worker
    instance_kill_count: 1
    timeout: 120
    cloud_type: aws
  - name: Reboot Scenario
    actions:
    - node_reboot_scenario
    node_name:
    label_selector: node-role.kubernetes.io/infra
    instance_kill_count: 1
    timeout: 120
    cloud_type: aws

yashashreesuresh: In the above case, if 5 different scenarios have to be applied on the same node one after another, the same configurations have to be specified 5 times which is redundant.

yashashreesuresh:
The current implementation uses abstract class; class for each cloud type inherits the abstract class. With this, the node scenario functions can to be called only after creating an object of the class. The advantage with using abstract class is that it makes sure all the cloud types support all the scenarios. But I think it would be easier to directly call functions without creating an object. Thoughts? @mffiedler @paigerube14

node_scenarios/node_scenarios_example.yml
paigerube14: I think my location of this comment is making it confusing and I think we are talking about two different things here. I am just suggesting that each node scenario in the list has a name, not each action and no need for redundancy 

Here is an example from the etcd scenario. Here the scenario itself has a name and then you can perform multiple steps/actions
```
scenarios:
  **- name: "delete etcd pods"**
    steps:
    - podAction:
        matches:
          - labels:
              namespace: "openshift-etcd"
              selector: "k8s-app=etcd"
```


Similarly, I am suggesting a name to the top of each node scenario you want to perform on a node. 
```
node_scenarios:
  **- name: Stop and Start Scenario**
    actions: 
    - node_stop_scenario
    - node_start_scenario
    node_name:
    label_selector: node-role.kubernetes.io/worker
    instance_kill_count: 1
    timeout: 120
    cloud_type: aws
```

yashashreesuresh: Suppose we want to perform stop, start, terminate, restart actions on node X. The actions would include node_stop_scenario, node_start_scenario, node_terminate_scenario, node_restart_scenario. In this case would the name be Start, Stop, Terminate and Restart scenario?

paigerube14: Yeh I see your point now, I agree it's redundant

mffiedler:
I think the use of the abstract class to allow for cloud-specific scenario implementations where the cloud instance APIs are required is the correct approach.    We also need to allow for scenarios which are not cloud specific (see the node crash scenarios which use oc debug node in the PR @paigerube14 started in #18 for an example).   Where possible, having a default/common implementation of the scenario would be beneficial and it can be overridden by a cloud-specific implementation where needed.   Hope that makes sense.

paigerube14:
I completely agree with Mike, I think the abstract might actually end up being easier for all the cloud providers to have a common layout of function. 

@yashashreesuresh did you want me to add my 2  functions (kubelet_action and crash_node) for the oc debug parts that Mike is talking about to your branch? How did you want to do that?

yashashreesuresh:
> I completely agree with Mike, I think the abstract might actually end up being easier for all the cloud providers to have a common layout.
> 
> @yashashreesuresh did you want me to add my 2 functions (kubelet_action and crash_node) for the oc debug parts that Mike is talking about to your branch? How did you want to do that?

@paigerube14 I have added the kubelet scenario as https://github.com/openshift-scale/kraken/pull/10/files#diff-d7c8b18dc1f8fd23bc1e62bdee0ccdceL8 subprocess.check_output doesn't finish to completion because once the kubelet is stopped, debug pod never terminates. Therefore, I had to use https://github.com/openshift-scale/kraken/pull/10/files#diff-d7c8b18dc1f8fd23bc1e62bdee0ccdceR15 subprocess.run command. 
I was trying your node_crash scenario and I observed many times the command gets executed but the node status doesn't change from Ready->NotReady->Ready. And sometimes the command pauses and doesn't finish to completion and I have to use Ctrl+C. However, I have added your command here https://github.com/openshift-scale/kraken/pull/10/files#diff-54651a20a8f57ef1c067dbaac8fdc0a6R63. It needs to be changed.

yashashreesuresh:
All the scenarios are added, PTAL @chaitanyaenr

paigerube14:
> > I completely agree with Mike, I think the abstract might actually end up being easier for all the cloud providers to have a common layout.
> > @yashashreesuresh did you want me to add my 2 functions (kubelet_action and crash_node) for the oc debug parts that Mike is talking about to your branch? How did you want to do that?
> 
> @paigerube14 I have added the kubelet scenario as https://github.com/openshift-scale/kraken/pull/10/files#diff-d7c8b18dc1f8fd23bc1e62bdee0ccdceL8 subprocess.check_output doesn't finish to completion because once the kubelet is stopped, debug pod never terminates. Therefore, I had to use https://github.com/openshift-scale/kraken/pull/10/files#diff-d7c8b18dc1f8fd23bc1e62bdee0ccdceR15 subprocess.run command.
> I was trying your node_crash scenario and I observed many times the command gets executed but the node status doesn't change from Ready->NotReady->Ready. And sometimes the command pauses and doesn't finish to completion and I had to use Ctrl+C. However, I have added your command here https://github.com/openshift-scale/kraken/pull/10/files#diff-54651a20a8f57ef1c067dbaac8fdc0a6R63. It needs to be changed.

I did have to use subprocess.Popen to get the stop kubelet to return properly. I'm not sure about the node_crash scenario. I'm not sure if the node will go to the notReady state. I was going off of the commands Mike had laid out in the original issue. I think for the node_crash scenario I also had to use subprocess.Popen to get a proper response. I think I thought either subprocess function worked so I didn't add it to my pull request, sorry about that. I tried the node_crash scenario from my branch and the Popen properly returns

yashashreesuresh:
> I did have to use subprocess.Popen to get the stop kubelet to return properly. I'm not sure about the node_crash scenario. I'm not sure if the node will go to the notReady state. I was going off of the commands Mike had laid out in the original issue. I think for the node_crash scenario I also had to use subprocess.Popen to get a proper response. I think I thought either subprocess function worked so I didn't add it to my pull request, sorry about that. I tried the node_crash scenario from my branch and the Popen properly returns

I tried with Popen, it's not working as I expected. The subprocess.Popen("oc debug node/" + node + " -- chroot /host dd if=/dev/urandom of=/proc/sysrq-trigger") command finishes. Only after few seconds the node goes to NotReady state. I expected the node to go to NotReady state while the Popen command is in execution because the node when to NotReady state after the debug pod is removed. Few times, the node state doesn't change at all.
I am not sure what's happening.

Anyways, I have removed the node_crash_scenario. You can add your code here https://github.com/openshift-scale/kraken/pull/10/files#diff-54651a20a8f57ef1c067dbaac8fdc0a6R58 once this PR is merged as it worked right for you.

yashashreesuresh:
All the changes are made in the code. PTAL, @chaitanyaenr @paigerube14 @mffiedler

mffiedler:
For node crash scenarios, nodes will not show immediately as NotReady in the api.   They have to miss a readiness check interval from the API server perspective.    If a node is in trouble with memory or disk pressure but still running, it can self report the issue immediately.  But for a crashed node, the cluster needs to notice it is gone to mark it NotReady.

yashashreesuresh:
> For node crash scenarios, nodes will not show immediately as NotReady in the api. They have to miss a readiness check interval from the API server perspective. If a node is in trouble with memory or disk pressure but still running, it can self report the issue immediately. But for a crashed node, the cluster needs to notice it is gone to mark it NotReady.

Added back the node crash scenario that was removed.

config/config.yaml
```diff
@@ -4,6 +4,8 @@ kraken:
         -    scenarios/etcd.yml
         -    scenarios/openshift-kube-apiserver.yml                           
         -    scenarios/openshift-apiserver.yml
+    node_scenarios:                                        # List of chaos node scenarios to load
+        -   scenarios/node_scenarios_example.yml
```
paigerube14, line 8: It might be also nice to show an example of the node_crash and stop_kubelet in a separate scenario or also in this one

paigerube14:
Is it expected that we set up the boto3 region and some other attributes before running? I did not set up of boto3 and I am getting an error` botocore.exceptions.NoRegionError: You must specify a region.` This might be good to add to the README or it might be nice to either specify those in the config or in the scenario itself where you define the cloud type to aws

yashashreesuresh:
> Is it expected that we set up the boto3 region and some other attributes before running? I did not set up of boto3 and I am getting an error` botocore.exceptions.NoRegionError: You must specify a region.` This might be good to add to the README or it might be nice to either specify those in the config or in the scenario itself where you define the cloud type to aws

boto3 client is created in the `default` region specified in the `~/.aws/config` file. This file is created when awscli is installed. Which version of AWS CLI are use using?

paigerube14:
> > Is it expected that we set up the boto3 region and some other attributes before running? I did not set up of boto3 and I am getting an error` botocore.exceptions.NoRegionError: You must specify a region.` This might be good to add to the README or it might be nice to either specify those in the config or in the scenario itself where you define the cloud type to aws
> 
> boto3 client is created in the `default` region specified in the `~/.aws/config` file. This file is created when awscli is installed. Which version of AWS CLI are use using?

Ah I didn't have an aws cli installed. I definitely think this should be put in the read me as an extra set up steps because that is not clear. Maybe just specify if using the specific node scenarios you'll need to set up aws cli with certain config. I followed https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-install.html if you want to use this in the documentation

yashashreesuresh:
> Ah I didn't have an aws cli installed. I definitely think this should be put in the read me as an extra set up steps because that is not clear. Maybe just specify if using the specific node scenarios you'll need to set up aws cli with certain config. I followed https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-install.html if you want to use this in the documentation

Yeah, I will add it in the README. Thank you!

config/config.yaml
paigerube14: In addition, it could be nice to maybe have a separate README defining how to set up a node scenario and all the possible actions that you can have.

yashashreesuresh: I have made changes in the README, PTAL.

paigerube14: LGTM thanks for adding that!

mffiedler:
/hold
I need some time to review and test this.  It has gotten pretty big.   In the future I have the suggestion that we add 1 scenario per PR.

yashashreesuresh:
> /hold
> I need some time to review and test this. It has gotten pretty big. In the future I have the suggestion that we add 1 scenario per PR.

Sure.

README.md
```diff
@@ -56,6 +58,43 @@ The report is generated in the run directory and it contains the information abo
 #### Cerberus to help with cluster health checks
 [Cerberus](https://github.com/openshift-scale/cerberus) can be used to monitor the cluster under test and the aggregated go/no-go signal generated by it can be consumed by Kraken to determine pass/fail. This is to make sure the Kubernetes/OpenShift environments are healthy on a cluster level instead of just the targeted components level. It is highly recommended to turn on the Cerberus health check feature avaliable in Kraken after installing and setting up Cerberus. To do that, set cerberus_enabled to True and cerberus_url to the url where Cerberus publishes go/no-go signal in the config file.
 
+### Kubernetes/OpenShift node chaos scenarios supported
+Following node chaos scenarios are supported:
```
chaitanyaenr, line 62: We might want to mention that the cloud related scenarios like start/stop/reboot are supported only for AWS as of now.

kraken/node_actions/abstract_node_scenarios.py
```diff
@@ -0,0 +1,68 @@
...
+                sys.exit(1)
+
+    # Node scenario to stop and start the kubelet
+    def stop_start_kubelet_scenario(self, instance_kill_count, node, timeout):
+        logging.info("Starting stop_start_kubelet_scenario injection")
+        self.stop_kubelet_scenario(instance_kill_count, node, timeout)
+        self.node_reboot_scenario(instance_kill_count, node, timeout)
+        logging.info("stop_start_kubelet_scenario has been successfully injected!")
+
+    # Node scenario to crash the node
```
chaitanyaenr, line 55: We might want to add a note in the readme that the node needs to be rebooted to recover from node crash and that it can be added as an action in the node scenarios config.

README.md
```diff
@@ -56,6 +58,43 @@ The report is generated in the run directory and it contains the information abo
...
+    node_name:
+    label_selector: node-role.kubernetes.io/infra
+    instance_kill_count: 1
+    timeout: 120
+    cloud_type: aws
+```
+
+**NOTE**: With aws as the cloud type, make sure [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html) is installed.
+
 ### Kubernetes/OpenShift chaos scenarios supported
```
chaitanyaenr, line 98: We might want to move the node scenarios under "Kubernetes/OpenShift chaos scenarios supported" section.

README.md
yashashreesuresh: Added it.

kraken/node_actions/abstract_node_scenarios.py
yashashreesuresh: Added it.

README.md
yashashreesuresh: Done.

chaitanyaenr:
@paigerube14 @mffiedler Thoughts?

README.md
```diff
@@ -57,7 +59,51 @@ The report is generated in the run directory and it contains the information abo
...
+
+**NOTE**: node_start_scenario, node_stop_scenario, node_stop_start_scenario, node_termination_scenario, node_reboot_scenario and stop_start_kubelet_scenario are supported only on AWS as of now.
+
+**NOTE**: With AWS as the cloud type, make sure [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html) is installed.
+
+Node scenarios can be injected by placing the node scenarios config files under node_scenarios option in the kraken config. Refer to [node_scenarios_example](https://github.com/openshift-scale/kraken/blob/master/scenarios/node_scenarios_example.yml) config file.
+
+
+```
+node_scenarios:
```
paigerube14, line 86: Should we make this in a separate readme for the node scenarios?

yashashreesuresh: I think it would be better to place everything in the main Readme as the user gets to know about different scenarios supported by Kraken.

mffiedler:
I think we should merge this and do any follow up work needed as a separate PR

LGTM

PR #17, opened 2020-06-03
Iterations

paigerube14:
This is for issue: https://github.com/openshift-scale/kraken/issues/12

Adding a loop to be able to run Kraken in multiple iterations or daemon/continuous mode

The iteration/daemon mode configuration will be similarly set up to Cerberus in the tunings area of the config.

I did not add an additional wait parameter like Mike mentioned in the original issue. I have the wait time still in between each scenario not between each iteration as well. Do we want an additional wait after each iteration injects all scenarios?

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -11,3 +11,5 @@ cerberus:
 
 tunings:
     wait_duration: 60                                      # Duration to wait between each chaos scenario
+    iterations: 1                                          # Number of times to execute the scenarios
+    daemon_mode: False                                     # Iterations are set to infinity which means that the cerberus will monitor the resources forever
--- a/run_kraken.py
+++ b/run_kraken.py
@@ -26,6 +26,8 @@ def main(cfg):
         scenarios = config["kraken"]["scenarios"]
         cerberus_enabled = config["cerberus"]["cerberus_enabled"]
         wait_duration = config["tunings"]["wait_duration"]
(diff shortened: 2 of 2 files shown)


mffiedler:
No additional "outer sleep" is required.  The normal scenario interval sleep is sufficient (until proven otherwise).

chaitanyaenr (approved):
LGTM

config/config.yaml
```diff
@@ -11,3 +11,5 @@ cerberus:
 
 tunings:
     wait_duration: 60                                      # Duration to wait between each chaos scenario
+    iterations: 2                                          # Number of times to execute the scenarios
```
chaitanyaenr, line 14: Maybe set the number of iterations to 1?

mffiedler: +1 to 1 iterations as a default

paigerube14: @chaitanyaenr  @mffiedler Is it ok to have the default for the daemon mode variable to True? or should I set it to false to only run 1 iteration by default?

mffiedler: I think daemon=false and 1 iteration is the least amount of surprise for someone running kraken for the first time.   Almost "demo" mode.

chaitanyaenr: +1 for setting daemon_mode to false.

mffiedler (approved):

PR #20, opened 2020-06-16
Adding updated scenarios for 3.0 powerfulseal version

paigerube14:
This is for: https://github.com/openshift-scale/kraken/issues/19

The scenarios should run the exact same way as before, the syntax is just a little different with 3.0 version of powerfulseal


chaitanyaenr (approved):
LGTM

PR #21, opened 2020-06-18
Check if killed component(s) recovered

paigerube14:
This is to cover: https://github.com/openshift-scale/kraken/issues/1

This adds the ability to run a post action scenario to check if the certain pod (nodes to come) has recovered after the certain scenario was executed. If the post action scenario does not complete properly it will be added to a list to rerun after the execution of the next scenario. 

The post action scenarios can be either a yaml/yml file that will be executed using powerfulseal or a custom script to validate the expected outputs. See [example](https://github.com/paigerube14/kraken/blob/9f3304594cea07d5ae3a9163a39afccfc8096d29/postActions/etcd-example.sh) for a custom bash script to validate 3 pods are running in the etcd namespace. This could be in place of the etcd.yml file also in the postAction folder

### DIFF
--- a/run_kraken.py
+++ b/run_kraken.py
@@ -12,6 +12,109 @@ import kraken.invoke.command as runcommand
 import pyfiglet
 
 
+# Get cerberus status
+def cerberus_integration(config):
+    cerberus_status = True
+    if config["cerberus"]["cerberus_enabled"]:
+        cerberus_url = config["cerberus"]["cerberus_url"]
+        if not cerberus_url:
+            logging.error("url where Cerberus publishes True/False signal is not provided.")
+            sys.exit(1)
+        cerberus_status = requests.get(cerberus_url).content
+        cerberus_status = True if cerberus_status == b'True' else False
+        if not cerberus_status:
+            logging.error("Received a no-go signal from Cerberus, looks like "
+                          "the cluster is unhealthy. Please check the Cerberus "
+                          "report for more details. Test failed.")
+            sys.exit(1)
+        else:
+            logging.info("Received a go signal from Ceberus, the cluster is healthy. "
+                         "Test passed.")
+    return cerberus_status
+
+
+# Function to publish kraken status to cerberus
+def publish_kraken_status(config, failed_post_scenarios):
+    cerberus_status = cerberus_integration(config)
+    if not cerberus_status:
+        if failed_post_scenarios:
+            if config['kraken']['exit_on_failure']:
+                logging.info("Cerberus status is not healthy and post action scenarios "
+                             "are still failing, exiting kraken run")
+                sys.exit(1)
+            else:
+                logging.info("Cerberus status is not healthy and post action scenarios "
+                             "are still failing")
+    else:
+
+        if failed_post_scenarios:
+            if config['kraken']['exit_on_failure']:
+                logging.info("Cerberus status is healthy but post action scenarios "
+                             "are still failing, exiting kraken run")
+                sys.exit(1)
+            else:
+                logging.info("Cerberus status is healthy but post action scenarios "
+                             "are still failing")
+
+
+def run_post_action(kubeconfig_path, scenario, pre_action_output=""):
+
+    if scenario.endswith(".yaml") or scenario.endswith(".yml"):
+        action_output = runcommand.invoke("powerfulseal autonomous "
+                                          "--use-pod-delete-instead-of-ssh-kill"
+                                          " --policy-file %s --kubeconfig %s --no-cloud"
+                                          " --inventory-kubernetes --headless"
+                                          % (scenario, kubeconfig_path))
+        # read output to make sure no error
+        if "ERROR" in action_output:
+            action_output.split("ERROR")[1].split('\n')[0]
+            if not pre_action_output:
+                logging.info("Powerful seal pre action check failed for " + str(scenario))
+            return False
+        else:
+            logging.info(scenario + " post action checks passed")
+
+    elif scenario.endswith(".py"):
+        action_output = runcommand.invoke("python3 " + scenario).strip()
+        if pre_action_output:
+            if pre_action_output == action_output:
+                logging.info(scenario + " post action checks passed")
+            else:
+                logging.info(scenario + ' post action response did not match pre check output')
+                return False
+    else:
+        # invoke custom bash script
+        action_output = runcommand.invoke(scenario).strip()
+        if pre_action_output:
+            if pre_action_output == action_output:
+                logging.info(scenario + " post action checks passed")
+            else:
+                logging.info(scenario + ' post action response did not match pre check output')
+                return False
+
+    return action_output
+
+
+# Perform the post scenario actions to see if components recovered
+def post_actions(kubeconfig_path, scenario, failed_post_scenarios, pre_action_output):
+
+    for failed_scenario in failed_post_scenarios:
+        post_action_output = run_post_action(kubeconfig_path,
+                                             failed_scenario[0], failed_scenario[1])
+        if post_action_output is not False:
+            failed_post_scenarios.remove(failed_scenario)
+        else:
+            logging.info('Post action scenario ' + str(failed_scenario) + "is still failing")
+
+    # check post actions
+    if len(scenario) > 1:
+        post_action_output = run_post_action(kubeconfig_path, scenario[1], pre_action_output)
+        if post_action_output is False:
+            failed_post_scenarios.append([scenario[1], pre_action_output])
+
+    return failed_post_scenarios
+
+
 # Main function
 def main(cfg):
     # Start kraken
@@ -24,7 +127,6 @@ def main(cfg):
             config = yaml.full_load(f)
         kubeconfig_path = config["kraken"]["kubeconfig_path"]
         scenarios = config["kraken"]["scenarios"]
-        cerberus_enabled = config["cerberus"]["cerberus_enabled"]
         wait_duration = config["tunings"]["wait_duration"]
         iterations = config["tunings"]["iterations"]
         daemon_mode = config["tunings"]['daemon_mode']
@@ -59,41 +161,34 @@ def main(cfg):
                          % str(iterations))
             iterations = int(iterations)
 
+        failed_post_scenarios = []
         # Loop to run the chaos starts here
         while (int(iteration) < iterations):
             # Inject chaos scenarios specified in the config
+            logging.info("Executing scenarios for iteration " + str(iteration))
             try:
                 # Loop to run the scenarios starts here
                 for scenario in scenarios:
-                    logging.info("Injecting scenario: %s" % (scenario))
+                    pre_action_output = run_post_action(kubeconfig_path, scenario[1])
                     runcommand.invoke("powerfulseal autonomous --use-pod-delete-instead-of-ssh-kill"
                                       " --policy-file %s --kubeconfig %s --no-cloud"
                                       " --inventory-kubernetes --headless"
-                                      % (scenario, kubeconfig_path))
-                    logging.info("Scenario: %s has been successfully injected!" % (scenario))
+                                      % (scenario[0], kubeconfig_path))
 
-                    if cerberus_enabled:
-                        cerberus_url = config["cerberus"]["cerberus_url"]
-                        if not cerberus_url:
-                            logging.error("url where Cerberus publishes True/False signal "
-                                          "is not provided.")
-                            sys.exit(1)
-                        cerberus_status = requests.get(cerberus_url).content
-                        cerberus_status = True if cerberus_status == b'True' else False
-                        if not cerberus_status:
-                            logging.error("Received a no-go signal from Cerberus, looks like the"
-                                          " cluster is unhealthy. Please check the Cerberus report"
-                                          " for more details. Test failed.")
-                            sys.exit(1)
-                        else:
-                            logging.info("Received a go signal from Ceberus, the cluster is "
-                                         "healthy. Test passed.")
+                    logging.info("Scenario: %s has been successfully injected!" % (scenario[0]))
                     logging.info("Waiting for the specified duration: %s" % (wait_duration))
                     time.sleep(wait_duration)
+                    failed_post_scenarios = post_actions(kubeconfig_path, scenario,
+                                                         failed_post_scenarios, pre_action_output)
+                    publish_kraken_status(config, failed_post_scenarios)
             except Exception as e:
                 logging.error("Failed to run scenario: %s. Encountered the following exception: %s"
-                              % (scenario, e))
+                              % (scenario[0], e))
             iteration += 1
+            logging.info("")
+        if failed_post_scenarios:
+            logging.error("Post scenarios are still failing at the end of all iterations")
+            sys.exit(1)
     else:
         logging.error("Cannot find a config at %s, please check" % (cfg))
         sys.exit(1)
--- a/scenarios/post_action_etcd_example.sh
+++ b/scenarios/post_action_etcd_example.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+pods="$(oc get pods -n openshift-etcd | grep -c Running)"
+echo "$pods"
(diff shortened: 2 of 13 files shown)


postActions/etcd.yml
```diff
@@ -0,0 +1,21 @@
+config:
+  runStrategy:
+    runs: 1
```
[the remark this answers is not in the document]
paigerube14: Not sure, will give that a try today!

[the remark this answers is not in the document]
paigerube14: I think the way I have it coded, the multiple runs won't work properly. I can definitely change it if necessary. But the output returns all the output from all runs so if we do 10 runs and and the first 5 fail and then we finally get 3 pods; it will still return "ERROR" in the response so the current implementation will think it failed. Thoughts on what to do here?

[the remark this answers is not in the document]
chaitanyaenr: Maybe keep checking the status and break once the condition is met and if not then timeout and fail after the set number of runs.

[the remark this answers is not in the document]
chaitanyaenr: Also think we need exit 1 to report the scenario failure and skip running the next scenarios since that might have a dependency on the failed component or it might affect the cluster more.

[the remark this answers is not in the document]
mffiedler: I can see two "modes" for Kraken:
1. Run a list of scenarios.  At the end, is the cluster healthy or no?
2. Run Kraken scenarios in daemon mode and see if the cluster can heal itself and stay healthy over time.   This would be useful in our reliability clusters (/cc: @wabouhamad )

Should we add a config option for exit/no exit?   An option to just keep dealing damage even if the cluster is in a suspect state?

run_kraken.py
```diff
@@ -12,6 +12,74 @@
...
+            if failed_post_scenarios:
+                logging.info("Cerberus status is not healthy and post action scenarios "
+                             "are still failing")
+            sys.exit(1)
+        else:
+            logging.info("Received a go signal from Ceberus, the cluster is "
+                         "healthy. Test passed.")
+
+            if failed_post_scenarios:
+                logging.info("Cerberus status is healthy but post action scenarios "
```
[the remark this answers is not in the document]
paigerube14: How would we know how many retries to do? An option in the config like "statusRetries"? Would this be fixed by Mike's suggestion with the exit/no exit option? Or be separate?

postActions/etcd.yml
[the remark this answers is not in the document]
paigerube14: I agree with the 2 scenarios and I feel like I have one covered currently. 
For the first scenarios I'll keep track of the failed post action scenarios and at the end give a status. For the second one I could either use the exit/no exit option in addition to the failed scenarios. For the no exit option I will just keep track of the failed scenarios and print them out so the user knows the failures during the daemon run? 

I definitely like the idea of exit/no exit option similar to the cerberus go/no go status. Would this be part of this pull request or should we make it separate? I think this would help us decide on some of the decisions we are talking about

run_kraken.py
```diff
@@ -12,6 +12,88 @@
...
+    else:
+        if config['kraken']['exit_on_failure']:
+            if failed_post_scenarios:
+                logging.info("Post action scenarios are failing, exiting kraken run")
+                sys.exit(1)
+
+
+# Perform the post scenario actions to see if components recovered
+def post_actions(kubeconfig_path, scenario, failed_post_scenarios):
+    for failed_scenario in failed_post_scenarios:
```
mffiedler, line 59: Just double checking my understanding of the logic.
    - Each time a scenario is run, call post_actions() after running the scenario itself
    - If there are any failed post-actions remaining from previous scenarios, run those post-actions again to see if they have cleared
    - If the current scenario array has a second entry, it is treated as a post-action and should either be yaml (powerfulseal) or custom bash.   If it fails, it goes on the list.

paigerube14: Yes that is correct, is there a different way that this could be accomplished?

mffiedler: That was for my understanding.  I am fine with the approach and the code looks good to me.   I did not run through all of the status publishing scenarios but did test the post actions.   @chaitanyaenr thoughts on the approach?

postActions/etcd.yml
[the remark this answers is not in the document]
chaitanyaenr: Cerberus signal is going to false leading to Kraken exiting 1 at the end of the iteration even in daemon mode if the components haven't recovered after the time specified in the scenario. Think that should be the behavior of Kraken in both non daemon and daemon mode for us to know when things are failing, also it might drive the cluster down the cluster down the hill if kraken keeps injecting failures even when things don't recover but that would be an interesting thing to do as well I guess.

run_kraken.py
chaitanyaenr: Sounds good to me. The issue might not be relevant to this PR but think we need to separate and print the pass/failures at the end of the iteration. Looks like the iterations are combined together, output from the current run:
```2020-06-26 13:02:40,479 [INFO] Daemon mode not enabled, will run through 2 iterations
2020-06-26 13:02:41,695 [INFO] Scenario: scenarios/etcd.yml has been successfully injected!
2020-06-26 13:02:41,695 [INFO] Waiting for the specified duration: 0
2020-06-26 13:02:42,909 [INFO] scenarios/post_action_etcd.yaml post action checks passed
2020-06-26 13:02:44,136 [INFO] Scenario: scenarios/openshift-kube-apiserver.yml has been successfully injected!
2020-06-26 13:02:44,137 [INFO] Waiting for the specified duration: 0
2020-06-26 13:02:45,354 [INFO] scenarios/post_action_openshift-kube-apiserver.yml post action checks passed
2020-06-26 13:02:46,572 [INFO] Scenario: scenarios/openshift-apiserver.yml has been successfully injected!
2020-06-26 13:02:46,572 [INFO] Waiting for the specified duration: 0
2020-06-26 13:02:47,787 [INFO] scenarios/post_action_openshift-apiserver.yml post action checks passed
2020-06-26 13:02:49,005 [INFO] Scenario: scenarios/etcd.yml has been successfully injected!
2020-06-26 13:02:49,005 [INFO] Waiting for the specified duration: 0
2020-06-26 13:02:50,215 [INFO] scenarios/post_action_etcd.yaml post action checks passed
2020-06-26 13:02:51,435 [INFO] Scenario: scenarios/openshift-kube-apiserver.yml has been successfully injected!
2020-06-26 13:02:51,435 [INFO] Waiting for the specified duration: 0
2020-06-26 13:02:52,653 [INFO] scenarios/post_action_openshift-kube-apiserver.yml post action checks passed
2020-06-26 13:02:53,861 [INFO] Scenario: scenarios/openshift-apiserver.yml has been successfully injected!
2020-06-26 13:02:53,861 [INFO] Waiting for the specified duration: 0
2020-06-26 13:02:55,080 [INFO] scenarios/post_action_openshift-apiserver.yml post action checks passed
```

scenarios/post_action_etcd_example.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+pods="$(oc get pods -n openshift-etcd | grep -c Running)"
```
yashashreesuresh, line 2: I think it would be nice to include a function where the user can place multiple checks within rather than a single command. Multiple checks can be performed and the function return value can be aggregated true/false values from each check.

run_kraken.py
```diff
@@ -12,6 +12,88 @@
...
+                                               " --inventory-kubernetes --headless"
+                                               % (failed_scenario, kubeconfig_path))
+        if "ERROR" not in post_action_output:
+            failed_post_scenarios.remove(failed_scenario)
+        else:
+            logging.info('Post action scenario ' + str(failed_scenario) + "is still failing")
+            error_message = post_action_output.split("ERROR")[1].split('\n')[0]
+            logging.info("Error message was " + str(error_message))
+    # check post actions
+    if len(scenario) > 1:
```
yashashreesuresh, line 72: I think it would be nice to allow the user to include multiple post_actions for each pod_scenario. Thoughts?

run_kraken.py
paigerube14: I added a new line at the end of each iteration and printed out the iteration number at the top of each loop. The output will now look like this (I only ran the etcd scenario).... Thoughts?

```
2020-06-26 13:31:03,551 [INFO] Daemon mode not enabled, will run through 2 iterations
2020-06-26 13:31:03,551 [INFO] Executing scenarios for iteration 0
2020-06-26 13:31:12,288 [INFO] Scenario: scenarios/etcd.yml has been successfully injected!
2020-06-26 13:31:12,289 [INFO] Waiting for the specified duration: 0
2020-06-26 13:31:14,054 [INFO] scenarios/post_action_etcd.yaml post action checks passed
2020-06-26 13:31:14,054 [INFO] 
2020-06-26 13:31:14,055 [INFO] Executing scenarios for iteration 1
2020-06-26 13:31:39,194 [INFO] Scenario: scenarios/etcd.yml has been successfully injected!
2020-06-26 13:31:39,194 [INFO] Waiting for the specified duration: 0
2020-06-26 13:31:40,804 [INFO] scenarios/post_action_etcd.yaml post action checks passed
2020-06-26 13:31:40,805 [INFO] 
```

run_kraken.py
paigerube14: I think it would be nice to be able to do multiple checks I'm just not sure how often it would be used. I think the post action pod scenarios and writing your own custom script could cover more than one check in itself and return a failure if any part of it faills. @mffiedler @chaitanyaenr what are your thoughts on this?

chaitanyaenr: +1, think the custom script option should cover scenarios having multiple checks and we can also mention a list of actions in the same config in case of multiple things to check.

run_kraken.py
```diff
@@ -12,6 +12,88 @@
...
+                                                   % (scenario[1], kubeconfig_path))
+            # read output to make sure no error
+            if "ERROR" in post_action_output:
+                error_message = post_action_output.split("ERROR")[1].split('\n')[0]
+                logging.info("Error message was " + str(error_message))
+                failed_post_scenarios.append(scenario[1])
+            else:
+                logging.info(scenario[1] + " post action checks passed")
+        else:
+            # invoke custom bash script
```
yashashreesuresh, line 87: I think it would be nice to provide the user with an option to provide post actions using a Python file in addition to bash script as this project is based on Python and the user might be familiar with Python over shell script. Thoughts?

paigerube14: Yes I agree, I can add in a check to see if the given post action scenario is a python file and run accordingly.

paigerube14:
I added a post actions scenario for the new openshift regex scenario. I am not sure about the exact pod count for this. Can someone take a look to see if they get the same count back (146?) Thanks

scenarios/post_action_regex_openshift_pod_kill.yml
```diff
@@ -0,0 +1,18 @@
...
+    - podAction:
+        matches:
+          - namespace: "openshift-.*"
+        filters:
+          - property:
+             name: "state"
+             value: "Running"
+        actions:
+          - checkPodCount:
+              count: 146
```
mffiedler, line 18: This is going to be a moving target and depends on what optional or configuration dependent operators and associated pods are present.  Example openshift-logging will add n+4 pods where n is the number of nodes.   The real comparison should be before/after counts.  Not sure if powerfulseal can do that, maybe a custom script?   We might have to think about how pre/post checks work together.

paigerube14: I do not think powerfulseal can do a changing counts based on a before count. I think they have to be hard coded in the yaml when it gets executed

Here are a couple of solutions I could think of: 

- I could overwrite the post action yaml counts using 2 different possible paths

1. I could add another item in the double array in the config for a pre action yaml or custom script that the response could help fill in the post action yaml counts
2. I could also parse the matches and filters section of the scenario yaml and run a more general kubectl/oc command to get the expected counts to fill in the yaml

- In a lot of the scenarios I could parse the output from powerfulseal running a certain scenario because it spits out how many pods it had to pick from. For the regex scenario I have it picking 3 pods so it doesn't print out the number of pods it had.
I could change the scenario to run 3 times based on the top configuration so I could parse it like the other scenarios and maybe get the number of pods it had to pick from in that output.

- Or like Mike said I could do a custom script with adding options for some of the most common deployed operators and how many pods they create. Here's what I'm thinking. 

1. Set the initial amount of pods in the "openshift-*" namespace (This would be hard coded to whatever number we find, maybe 146?) 
2. Check if certain operators are deployed (find number of pods for that operator)
3. Add to existing amount of pods to initial count 
4. Repeat for as many operators as we want to check for
5. do a final get of all pods in "openshift-*" namespace and make sure they match the number we calculated

Thoughts?

chaitanyaenr: Think pre action and post action might work. We can run the pre action script before the scenario which returns the pod count in the given namespaces ( openshift* for random pod killing scenario ) and then the post action script ( same script as pre action ) after the scenario. We can then compare the values to determine pass/fail without having to modify/specify the pod count check in the powerfulseal config provided there are no changes to the cluster like increase in the node count leading to more pods ( in case of daemonsets ).

The pre/post action scripts are supplied on the fly and can be changed before kicking off the kraken, so user can specify the namespace to monitor depending on what is specified in the powerfulseal config file.

We can also make this simple by asking user to set the expected pod count in the yaml by calculating it manually with an assumption that the env doesn't change during the run.

run_kraken.py
```diff
@@ -12,6 +12,96 @@
 import pyfiglet
 
 
+# Function to publish kraken status to cerberus
+def publish_kraken_status(config, failed_post_scenarios):
+
+    if config["cerberus"]["cerberus_enabled"]:
+        cerberus_url = config["cerberus"]["cerberus_url"]
```
yashashreesuresh, line 19: Let’s make cerberus_integration a function as we will need to consume True/False signals after every chaos scenario (pod scenario, node scenario, cluster shutdown scenario, etc). I have added a [cerberus_integration](https://github.com/openshift-scale/kraken/pull/10/files#diff-cfec6e969439720ae75ffa7156ac298bR17) function in the node scenarios PR to achieve the same, it can be modified to return True/False signals additionally to fit in with the pod scenarios as well.

paigerube14:
I've updated this pull request to do a pre and post check for all types of files. If it is not a yaml file it uses the returned output from the pre check for if the post check is correct.

paigerube14:
@chaitanyaenr @yashashreesuresh @mffiedler have any of you had time to take a look at this? Thanks

run_kraken.py
yashashreesuresh: It would be great if this could be done.

mffiedler:
@paigerube14 will test/review today.

run_kraken.py
paigerube14: ah yes ok sorry I missed this

paigerube14: @yashashreesuresh we have pretty much the same function here. I'm a little unclear on what you wanted me to change here? Just the name of the function?

yashashreesuresh: A function to be defined which when called returns Cerberus true/false signal alone. In the current implementation of _publish_kraken_status_, retrieval of status from Cerberus is part of the entire code. I was telling about separating the retrieval of Cerberus status part.

yashashreesuresh: Because post actions works only with pod scenarios for now. In the above function, _failed_post_scenarios_ is included which is for pod scenarios and is not applicable for node scenarios and hence we need a separate function for obtaining true/false from cerberus.

paigerube14: changed it up! what do you think?

yashashreesuresh: Looks good! Thanks for the change!

yashashreesuresh (approved):
LGTM!

mffiedler:
Tested and reviewed.  LGTM.  Thanks for the PR.  If needed, we can refine in follow-up work.

Issue #9, opened 2020-04-29
Run containerized kraken as a kubernetes/openshift deployment


mffiedler:
Further improve the ease of use of containerized kraken by providing an image on quay.io and creating a kubernetes or openshift deployment to run it.   Config can be provided via configmap.

chaitanyaenr:
When we run Kraken/Cerberus as a Kubernetes/OpenShift object, there might be cases where the pod which is running the tool might get disrupted unless we add a way to blacklist it during the run. For instance, we might not get the results from the pod running Kraken or Cerberus when the API iteself is down. When we run them external to OpenShift, we can get a signal and catch test failures even when the API is down. Thoughts?

But agree that we might need to support both - running it internal/external to the cluster to improve the ease of use and users can pick the option depending on the use case.

chaitanyaenr:
We can do something similar to https://github.com/openshift-scale/cerberus/tree/master/containers keeping in mind to avoid killing anything in namespace/node with kraken running.

chaitanyaenr:
@paigerube14 do you want to pick this?

chaitanyaenr:
Fixed by https://github.com/openshift-scale/kraken/pull/24. Thanks @paigerube14.

PR #24, opened 2020-06-23
Adding kraken containerization with instructions

paigerube14:
This is to cover: https://github.com/openshift-scale/kraken/issues/9

Very similar to how cerberus is deployed. I didn't think that the service/route was necessary here, correct?

### DIFF
--- a/containers/kraken.yml
+++ b/containers/kraken.yml
@@ -1,0 +1,42 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: kraken-deployment
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      tool: Kraken
+  template:
+    metadata:
+      labels:
+        tool: Kraken
+    spec:
+      serviceAccountName: useroot
+      containers:
+        - name: kraken
+          securityContext:
+            privileged: true
+          image: quay.io/openshift-scale/kraken
+          command: ["/bin/sh", "-c"]
+          args: ["python3 run_kraken.py -c config/config.yaml"]
+          ports:
+            - containerPort: 8080
+          volumeMounts:
+            - mountPath: "/root/.kube"
+              name: config
+            - mountPath: "/root/kraken/config"
+              name: kraken-config
+            - mountPath: "/root/kraken/scenarios"
+              name: scenarios-config
+      volumes:
+        - name: config
+          configMap:
+            name: kube-config
+        - name: kraken-config
+          configMap:
+            name: kraken-config
+        - name: scenarios-config
+          configMap:
+            name: scenarios-config
--- a/kraken/kubernetes/client.py
+++ b/kraken/kubernetes/client.py
@@ -1,6 +1,10 @@
 from kubernetes import client, config
 from kubernetes.client.rest import ApiException
 import logging
+import kraken.invoke.command as runcommand
+import json
+
+kraken_node_name = ""
 
 
 # Load kubeconfig and initialize kubernetes python client
@@ -22,6 +26,21 @@ def list_nodes():
     return nodes
 
 
+# List nodes in the cluster that can be killed
+def list_killable_nodes():
+    nodes = []
+    try:
+        ret = cli.list_node(pretty=True)
+    except ApiException as e:
+        logging.error("Exception when calling CoreV1Api->list_node: %s\n" % e)
+    for node in ret.items:
+        if kraken_node_name != node.metadata.name:
+            for cond in node.status.conditions:
+                if str(cond.type) == "Ready" and str(cond.status) == "True":
+                    nodes.append(node.metadata.name)
+    return nodes
+
+
 # List pods in the given namespace
 def list_pods(namespace):
     pods = []
@@ -32,6 +51,14 @@ def list_pods(namespace):
                        CoreV1Api->list_namespaced_pod: %s\n" % e)
     for pod in ret.items:
         pods.append(pod.metadata.name)
+    return pods
+
+
+def get_all_pods():
+    pods = []
+    ret = cli.list_pod_for_all_namespaces(pretty=True)
+    for pod in ret.items:
+        pods.append([pod.metadata.name, pod.metadata.namespace])
     return pods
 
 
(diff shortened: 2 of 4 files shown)


mffiedler:
> This is to cover: #9
> 
> Very similar to how cerberus is deployed. I didn't think that the service/route was necessary here, correct?

Service/route not required.  Kraken does not (yet, at least) expose an API or endpoint.

containers/kraken.yml
```diff
@@ -0,0 +1,37 @@
...
+          command: ["/bin/sh", "-c"]
+          args: ["python3 run_kraken.py -c config/config.yaml"]
+          ports:
+            - containerPort: 8080
+          volumeMounts:
+            - mountPath: "/root/.kube"
+              name: config
+            - mountPath: "/root/kraken/config"
+              name: kraken-config
+      volumes:
```
yashashreesuresh, line 31: The `scenarios` directory has to be mounted as the user places all the scenarios here.
After the node scenarios PR is merged, the `node scenarios` directory has to mounted as well. And the code has to be modified to make sure the node chosen for node scenario injection doesn't have kraken pod running on it.
Instructions regarding modifying pod and node scenarios in `scenarios` and `node scenarios` directories has to be added in the README.md

paigerube14: I have added a mount for the scenarios directory. How do we want to proceed about the 2 coming stories that both add directories (recovered components and node failures scenarios)? Should we add the necessary mounts to each pull requests if this pull request gets merged first? 

Also, do I need to change anything on my side so the node kraken gets deployed on doesn't get killed during the scenario injections or will that be up to the user to properly specify nodes in their yaml files that don't include the one kraken is running on?

yashashreesuresh: You need to make change in this function: https://github.com/openshift-scale/kraken/pull/10/files#diff-54651a20a8f57ef1c067dbaac8fdc0a6R62 to make sure the selected node doesn't have kraken pod running on it.

chaitanyaenr: Maybe organize all the scenarios/post actions under scenarios directory, this way we can just mount one directory?

paigerube14: > Maybe organize all the scenarios/post actions under scenarios directory, this way we can just mount one directory?

Updated other pull request

paigerube14: > You need to make change in this function: https://github.com/openshift-scale/kraken/pull/10/files#diff-54651a20a8f57ef1c067dbaac8fdc0a6R62 to make sure the selected node doesn't have kraken pod running on it.

Added code for this in the kraken/kubernetes/client, please take a look

kraken/kubernetes/client.py
```diff
@@ -18,7 +24,9 @@ def list_nodes():
     except ApiException as e:
         logging.error("Exception when calling CoreV1Api->list_node: %s\n" % e)
     for node in ret.items:
-        nodes.append(node.metadata.name)
+        if kraken_node_name != node.metadata.name:
```
yashashreesuresh, line 27: I don't think it's a good idea to not append the node name to list of nodes if it's kraken_node_name as the purpose of this function is to obtain all the node names and this function might be called at other parts of the program as well.

paigerube14: I'll make a separate function similar to the list_nodes just to return all nodes but the one kraken is running on. I think you are going to have to update your get_node function to call it so you don't kill it, correct?

kraken/kubernetes/client.py
```diff
@@ -96,3 +112,30 @@ def monitor_component(iteration, component_namespace):
...
+        if "kraken-deployment" in pod[0]:
+            kraken_pod_name = pod[0]
+            kraken_project = pod[1]
+            break
+    # have to switch to proper project
+
+    if kraken_pod_name:
+        logging.info('pod name' + str(kraken_pod_name))
+        # get kraken-deployment pod, find node name
+        runcommand.invoke("oc project " + str(kraken_project))
```
yashashreesuresh, line 132: `oc project` isn't Kubernetes compatible.

kraken/kubernetes/client.py
yashashreesuresh: Yeah

paigerube14: @yashashreesuresh I have created a new function named list_killable_nodes to return the list of nodes that kraken is not running on. Can you please update your node_scenarios request to incorporate this? Thanks!

kraken/kubernetes/client.py
paigerube14: Fixed, please take another look

kraken/kubernetes/client.py
yashashreesuresh: Definitely!

paigerube14:
Are there any comments/updates I missed here that I can update so this can get merged?

chaitanyaenr (approved):
LGTM

mffiedler:
@yashashreesuresh Any further changes you would like to see?

kraken/kubernetes/client.py
```diff
@@ -1,13 +1,19 @@
...
+
+kraken_node_name = ""
 
 
 # Load kubeconfig and initialize kubernetes python client
 def initialize_clients(kubeconfig_path):
     global cli
     config.load_kube_config(kubeconfig_path)
     cli = client.CoreV1Api()
+    global kraken_node_name
```
yashashreesuresh, line 15: I think this line and the following line can be removed as kraken_node_name is initialized previously (line 7 of client.py)

kraken/kubernetes/client.py
```diff
@@ -22,6 +28,19 @@ def list_nodes():
     return nodes
 
 
+# List nodes in the cluster that can be killed
+def list_killable_nodes():
```
yashashreesuresh, line 32: It would be nice if this function could be modified to return killable `Ready` nodes. After the node chaos injection, the node might not come back to Ready state. In that case the node which is in NotReady state is not considered for next chaos scenario injection.

paigerube14: Added, please take a look

kraken/kubernetes/client.py
paigerube14: Took those lines out

yashashreesuresh (approved):
LGTM

PR #25, opened 2020-07-09
Added cluster shut down scenario

yashashreesuresh:
This commit adds a scenario to shut down all the nodes including the masters and restarts them after a specified duration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@ Kraken supports pod, node and time/date based scenarios.
 
 - [Time Scenarios](docs/time_scenarios.md)
 
+- [Cluster Shut Down Scenarios](docs/cluster_shut_down_scenarios.md)
+
 ### Kraken scenario pass/fail criteria and report
 It's important to make sure to check if the targeted component recovered from the chaos injection and also if the Kubernetes/OpenShift cluster is healthy as failures in one component can have an adverse impact on other components. Kraken does this by:
 - Having built in checks for pod and node based scenarios to ensure the expected number of replicas and nodes are up. It also supports running custom scripts with the checks.
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -18,6 +18,8 @@ kraken:
         -   litmus_scenarios:                              # List of litmus scenarios to load
             - - https://hub.litmuschaos.io/api/chaos/1.10.0?file=charts/generic/node-cpu-hog/rbac.yaml
               - scenarios/node_hog_engine.yaml
+        -   cluster_shut_down_scenario:
+            -   scenarios/cluster_shut_down_scenario.yml
 
 cerberus:
     cerberus_enabled: False                                # Enable it when cerberus is previously installed
--- a/run_kraken.py
+++ b/run_kraken.py
@@ -12,7 +12,7 @@ import kraken.kubernetes.client as kubecli
 import kraken.invoke.command as runcommand
 import kraken.litmus.common_litmus as common_litmus
 import kraken.node_actions.common_node_functions as nodeaction
-from kraken.node_actions.aws_node_scenarios import aws_node_scenarios
+from kraken.node_actions.aws_node_scenarios import AWS, aws_node_scenarios
 from kraken.node_actions.general_cloud_node_scenarios import general_node_scenarios
 from kraken.node_actions.gcp_node_scenarios import gcp_node_scenarios
 import kraken.time_actions.common_time_functions as time_actions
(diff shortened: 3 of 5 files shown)


yashashreesuresh:
This PR is based on node_scenarios PR: #[10](https://github.com/openshift-scale/kraken/pull/10) as I am reusing the AWS node stop and start functions.

kraken/invoke/command.py
```diff
@@ -5,8 +5,17 @@
...
+    except Exception as e:
+        logging.error("Failed to run %s, error: %s" % (command, e))
+    return out
+
+
+def run(command):
+    try:
+        subprocess.run(command, shell=True, universal_newlines=True, timeout=45)
     except Exception:
-        logging.error("Failed to run %s" % (command))
```
paigerube14, line 11: I think you should keep a logging statement here and not just pass

kraken/kubernetes/client.py
```diff
@@ -15,10 +15,13 @@ def initialize_clients(kubeconfig_path):
 
 
 # List nodes in the cluster
-def list_nodes():
+def list_nodes(label_selector=None):
     nodes = []
     try:
-        ret = cli.list_node(pretty=True)
+        if label_selector:
+            ret = cli.list_node(pretty=True, label_selector=label_selector)
```
paigerube14, line 22: Since you are giving label_selector=None, does just using line 22 work instead of the if statement? 

Also, I do not see anywhere where you call list_nodes and pass it a label_selector. Is that correct?

paigerube14:
I have just a few little comments but I was able to run this cluster shut down the other day and it worked perfectly!

run_kraken.py
```diff
@@ -1,15 +1,111 @@
...
+    node_id = {}
+    for node in nodes:
+        node_id[node] = cloud_object.get_instance_id(node)
+
+    for _ in range(runs):
+        logging.info("Starting cluster_shut_down scenario injection")
+        for node in nodes:
+            cloud_object.stop_instances(node_id[node])
+        logging.info("Waiting for 250s to shut down all the nodes")
+        time.sleep(250)
```
paigerube14, line 88: Are we able to have the user set this in their config or use 250 as a default? Is there a specific reason we chose 250 seconds here?

run_kraken.py
```diff
@@ -1,15 +1,111 @@
...
+                try:
+                    cloud_object.start_instances(node_id[node])
+                    restarted_nodes.add(node)
+                except Exception:
+                    time.sleep(10)
+                    continue
+            stopped_nodes = nodes - restarted_nodes
+        logging.info("Waiting for 250s to allow cluster component initilization")
+        time.sleep(250)
+        logging.info("Successfully injected cluster_shut_down scenario!")
```
paigerube14, line 106: Are we able to add in a verification that the nodes are all back up and ready here? Is that too much for kraken that it can just be handled in cerberus? Thoughts?

kraken/kubernetes/client.py
yashashreesuresh: Yeah, previously in [line](https://github.com/openshift-scale/kraken/pull/25/files#diff-1da5ecddeb0574d357ad50edac640d84R15), list_nodes was used instead of list_killable_node.

run_kraken.py
yashashreesuresh: There's no specific reason, it took around 2 mins on 10 node cluster. So I chose 250 seconds to accommodate clusters of bigger sizes. [start_instance](https://github.com/openshift-scale/kraken/pull/25/files#diff-cfec6e969439720ae75ffa7156ac298bR98) function can be called on a node only when it is in stopped state else it throws error. However I have added a [try except](https://github.com/openshift-scale/kraken/pull/25/files#diff-cfec6e969439720ae75ffa7156ac298bR97) condition such that we sleep for 10 additional seconds when a node isn't in stopped state even after 250 seconds.

run_kraken.py
yashashreesuresh: I think this part would be handled by cerberus. With cerberus intergration, when we receive a true, kraken proceeds with next scenario indicating all the nodes are ready but with false, we terminate kraken indicating some components aren't healthy. But it can be explicitly specified after [line](https://github.com/openshift-scale/kraken/pull/25/files#diff-cfec6e969439720ae75ffa7156ac298bR107) when cerberus integration is enabled if needed. Thoughts?

kraken/invoke/command.py
yashashreesuresh: This function was added just for the stop_kubelet node scenario, when this command is executed once, the kubelet stops even if the command exceeds the timeout (here set to 45s) and exception indicating 'failed to run command' wouldn't be meaningful in that case.

chaitanyaenr:
@yashashreesuresh Can we rebase the code please?

yashashreesuresh:
> @yashashreesuresh Can we rebase the code please?

@chaitanyaenr I have rebased the code. But I was not able to test completely as I don't have a cluster.
The code needs to be tested.

After this PR gets merged, [cluster_shut_down](https://github.com/cloud-bulldozer/kraken/pull/25/files#diff-2219b910ebc424a2f189a7be011d52196e2082631eb36cbecc66e853fbbfaea7R280) can be moved to a separate file where cluster_shut_down scenario for other cloud types can be added as well.

chaitanyaenr:
@yashashreesuresh Thanks for rebasing the code. No worries about the cluster, will test the scenario and let you know how it goes.

README.md
```diff
@@ -41,4 +43,4 @@ We are always looking for more enhancements, fixes to make it better, any contri
 ### Community
 Key Members(slack_usernames): paigerube14, rook, mffiedler, mohit, dry923, rsevilla, ravi
 * [**#sig-scalability on Kubernetes Slack**](https://kubernetes.slack.com)
-* [**#forum-perfscale on CoreOS Slack**](https://coreos.slack.com)
+* [**#forum-perfscale on CoreOS Slack**](https://coreos.slack.com)
```
paigerube14, line 46: is this an open slack channel that anyone can get on?

config/config.yaml
```diff
@@ -18,6 +18,8 @@ kraken:
         -   litmus_scenarios:                              # List of litmus scenarios to load
             - - https://hub.litmuschaos.io/api/chaos/1.10.0?file=charts/generic/node-cpu-hog/rbac.yaml
               - scenarios/node_hog_engine.yaml
+        -   cluster_shut_down_scenario:
```
paigerube14, line 21: The scenario type in run_kraken has an "s" at the end of the scenario. Just need to add an s here to be: **cluster_shut_down_scenarios**

run_kraken.py
```diff
@@ -277,6 +277,57 @@ def litmus_scenarios(scenarios_list, config, litmus_namespaces, litmus_uninstall
     return litmus_namespaces
 
 
+# Inject the cluster shut down scenario
+def cluster_shut_down(shut_down_config, config):
+    runs = shut_down_config["runs"]
+    shut_down_duration = shut_down_config["shut_down_duration"]
+    cloud_type = shut_down_config["cloud_type"]
+    if cloud_type == "aws":
```
paigerube14, line 285: Can we add in the other cloud types that have been added? 
In addition, could we add in an else case for if the cloud type is not supported and to not run/exit with an error message

paigerube14:
Based most of my coding on this PR and got it merged for AWS, Azure, Openstack and GCP
https://github.com/cloud-bulldozer/kraken/pull/102

PR #28, opened 2020-08-18
Adding Kraken to PerfScale Pipeline

yashashreesuresh:
This commit adds Kraken to CI pipeline and thereby enabling chaos
scenarios to be injected on specified jump host. Link to the [Jenkins job](http://perf-sm5039-4-8.perf.lab.eng.rdu2.redhat.com:8080/job/karken_jjb/) to achieve the same.

### DIFF
--- a/CI/scenarios/etcd.yml
+++ b/CI/scenarios/etcd.yml
@@ -1,0 +1,20 @@
+config:
+  runStrategy:
+    runs: 1
+    maxSecondsBetweenRuns: 30
+    minSecondsBetweenRuns: 1
+scenarios:
+  - name: "delete etcd pods"
+    steps:
+    - podAction:
+        matches:
+          - labels:
+              namespace: "openshift-etcd"
+              selector: "k8s-app=etcd"
+        filters:
+          - randomSample:
+              size: 1
+        actions:
+          - kill:
+              probability: 1
+              force: true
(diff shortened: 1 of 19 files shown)


yashashreesuresh:
Please have a look @chaitanyaenr @mffiedler @paigerube14

chaitanyaenr:
@paigerube14 @mffiedler CI directory will store the configs being used by the PerfScale CI pipeline, the plan is to tweak them accordingly instead of changing the ones in the base dir of the Kraken.

paigerube14:
Are we able to update the config to point to the scenarios under the CI folder so we won't have to keep up with changing the scenarios in both places? Thoughts?

chaitanyaenr:
> Are we able to update the config to point to the scenarios under the CI folder so we won't have to keep up with changing the scenarios in both places? Thoughts?

The CI configs will be tailored to our cluster setup vs the defaults which is why we have a separate config and scenarios vs re-using the existing ones. The plan is to tweak the CI configs to meet our needs accordingly as we finalize what to run as part of the PerfScale CI. Thoughts?

paigerube14:
> > Are we able to update the config to point to the scenarios under the CI folder so we won't have to keep up with changing the scenarios in both places? Thoughts?
> 
> The CI configs will be tailored to our cluster setup vs the defaults which is why we have a separate config and scenarios vs re-using the existing ones. The plan is to tweak the CI configs to meet our needs accordingly as we finalize what to run as part of the PerfScale CI. Thoughts?

That makes sense!

PR #30, opened 2020-09-11
Retry powerfulseal for post action checks

paigerube14:
Updating scenarios to use retries that has been added to the master branch of powerfulseal. Need to figure out good timeout in number of seconds or a retry count that will catch a real issue 

Should I delete the post action scenario files that will no longer be needed?

Not sure when the next time the powerfulseal pip version will be updated so want to use the github master branch for now. Will want to update later


chaitanyaenr:
@paigerube14 This is great, let's remove the post action files which were used check the pod count before wherever possible. Using the master branch till they release a new version sounds good.

paigerube14:
> @paigerube14 This is great, let's remove the post action files which were used check the pod count before wherever possible. Using the master branch till they release a new version sounds good.

I took out some of the post action files, but left one etcd one as an example.

I noticed the prometheus.yml file is not in the config. Is this on purpose? Should I add it into the config for this PR?

chaitanyaenr:
> > @paigerube14 This is great, let's remove the post action files which were used check the pod count before wherever possible. Using the master branch till they release a new version sounds good.
> 
> I took out some of the post action files, but left one etcd one as an example.
Sounds good.
> 
> I noticed the prometheus.yml file is not in the config. Is this on purpose? Should I add it into the config for this PR?
We wanted to show the way user can add scenarios by specifying couple of usual scenarios we run to the config and also users might loose the prometheus data when killed/restarted when it's not configured to use a persistent volume, so should be okay to leave it out of config IMHO. Thoughts?

paigerube14:
> > > @paigerube14 This is great, let's remove the post action files which were used check the pod count before wherever possible. Using the master branch till they release a new version sounds good.
> > 
> > 
> > I took out some of the post action files, but left one etcd one as an example.
> > Sounds good.
> > I noticed the prometheus.yml file is not in the config. Is this on purpose? Should I add it into the config for this PR?
> > We wanted to show the way user can add scenarios by specifying couple of usual scenarios we run to the config and also users might loose the prometheus data when killed/restarted when it's not configured to use a persistent volume, so should be okay to leave it out of config IMHO. Thoughts?

Yes that makes sense!

chaitanyaenr (approved):
LGTM

paigerube14:
@mffiedler whenever you get a chance are you able to review this PR? Thanks!

mffiedler (approved):
LGTM - thanks @paigerube14

PR #33, opened 2020-10-07
Litmus integration with node cpu hog experiment

paigerube14:
This PR is for https://github.com/openshift-scale/kraken/issues/31

This will install litmus, apply any experiments and check the status of them and then tear down litmus with each kraken run. 

The scenario I was able to run and am adding the scenario files here are for the node cpu hog. For this scenario you have to add the node you want to hog the cpu of in both the scenario/node_hog.yaml as well as scenario/node_hog_engine.yaml

Other scenarios that we can add can be found here: 
https://hub.litmuschaos.io/

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1,7 +1,9 @@
 kraken:
     kubeconfig_path: /root/.kube/config                    # Path to kubeconfig
     exit_on_failure: False                                 # Exit when a post action scenario fails
-    chaos_scenarios:                                         # List of policies/chaos scenarios to load
+    litmus_version: v1.10.0                                # Litmus version to install
+    litmus_uninstall: False                                # If you want to uninstall litmus if failure
+    chaos_scenarios:                                       # List of policies/chaos scenarios to load
         -   pod_scenarios:                                 # List of chaos pod scenarios to load
             - -    scenarios/etcd.yml
             - -    scenarios/regex_openshift_pod_kill.yml
@@ -13,6 +15,9 @@ kraken:
             - -    scenarios/openshift-kube-apiserver.yml
         -   time_scenarios:                                # List of chaos time scenarios to load
             - scenarios/time_scenarios_example.yml
+        -   litmus_scenarios:                              # List of litmus scenarios to load
+            - - https://hub.litmuschaos.io/api/chaos/1.10.0?file=charts/generic/node-cpu-hog/rbac.yaml
+              - scenarios/node_hog_engine.yaml
 
 cerberus:
     cerberus_enabled: False                                # Enable it when cerberus is previously installed
--- a/docs/litmus_scenarios.md
+++ b/docs/litmus_scenarios.md
@@ -1,0 +1,41 @@
+### Litmus Scenarios
+Kraken consumes [Litmus](https://github.com/litmuschaos/litmus) under the hood for some infrastructure, pod, and node scenarios
+ 
+Official Litmus documentation and to read more information on specifics of Litmus resources can be found [here](https://docs.litmuschaos.io/docs/next/getstarted/)
+
+
+#### Litmus Chaos Custom Resources
+There are 3 custom resources that are created during each Litmus scenario. Below is a description of the resources:
+* ChaosEngine: A resource to link a Kubernetes application or Kubernetes node to a ChaosExperiment. ChaosEngine is watched by Litmus' Chaos-Operator which then invokes Chaos-Experiments
+* ChaosExperiment: A resource to group the configuration parameters of a chaos experiment. ChaosExperiment CRs are created by the operator when experiments are invoked by ChaosEngine.
+* ChaosResult : A resource to hold the results of a chaos-experiment. The Chaos-exporter reads the results and exports the metrics into a configured Prometheus server.
+
+### Understanding Litmus Scenarios  
+
+To run Litmus scenarios we need to apply 3 different resources/yaml files to our cluster
+1. **Chaos experiments** contain the actual chaos details of a scenario
+
+    i. This is installed automatically by Kraken (does not need to be specified in kraken scenario configuration)
+    
+2. **Service Account**: should be created to allow chaosengine to run experiments in your application namespace. Usually sets just enough permissions to a specific namespace to be able to run the experiment properly 
+
+    i. This can be defined using either a link to a yaml file or a downloaded file in the scenarios folder
+    
+3. **Chaos Engine** connects the application instance to a Chaos Experiment. This is where you define the specifics of your scenario; ie: the node or pod name you want to cause chaos within 
+
+    i. This is a downloaded yaml file in the scenarios folder, full list of scenarios can be found [here](https://hub.litmuschaos.io/)
+
+**NOTE**: By default all chaos experiments will be installed based on the version you give in the config file. 
+
+Adding a new Litmus based scenario is as simple as adding references to 2 new yaml files (the Service Account and Chaos engine files for your scenario ) in the Kraken config.
+
+### Current Scenarios
+
+Following are the start of scenarios for which a chaos scenario config exists today. 
+
+Component                | Description                                                                                        | Working
+------------------------ | ---------------------------------------------------------------------------------------------------| ------------------------- |
+Node CPU Hog             | Chaos scenario that hogs up the CPU on a defined node for a specific amount of time                | :heavy_check_mark:        |
+
+
+
--- a/kraken/litmus/common_litmus.py
+++ b/kraken/litmus/common_litmus.py
@@ -1,0 +1,103 @@
+import kraken.invoke.command as runcommand
+import logging
+import time
+import sys
+
+
+# Install litmus and wait until pod is running
+def install_litmus(version):
+    runcommand.invoke("kubectl apply -f "
+                      "https://litmuschaos.github.io/litmus/litmus-operator-%s.yaml" % version)
+
+    runcommand.invoke("oc patch -n litmus deployment.apps/chaos-operator-ce --type=json --patch ' "
+                      "[ { \"op\": \"add\", \"path\": \"/spec/template/spec/containers/0/env/-\", "
+                      "\"value\": { \"name\": \"ANALYTICS\", \"value\": \"FALSE\" } } ]'")
+
+    runcommand.invoke("oc wait deploy -n litmus chaos-operator-ce --for=condition=Available")
+
+
+def deploy_all_experiments(version_string):
+
+    if not version_string.startswith("v"):
+        logging.error("Incorrect version string for litmus, needs to start with 'v' "
+                      "followed by a number")
+        sys.exit(1)
+    version = version_string[1:]
+
+    runcommand.invoke("kubectl apply -f "
+                      "https://hub.litmuschaos.io/api/chaos/%s?file=charts/generic/experiments.yaml"
+                      % version)
+
+
+def delete_experiments():
+    runcommand.invoke("kubectl delete chaosengine --all")
+
+
+# Check status of experiment
+def check_experiment(engine_name, experiment_name, namespace):
+    chaos_engine = runcommand.invoke("kubectl get chaosengines/%s -n %s -o jsonpath="
+                                     "'{.status.engineStatus}'" % (engine_name, namespace))
+    engine_status = chaos_engine.strip()
+    max_tries = 30
+    engine_counter = 0
+    while engine_status.lower() != "running" and engine_status.lower() != "completed":
+        time.sleep(10)
+        logging.info("Waiting for engine to start running.")
+        chaos_engine = runcommand.invoke("kubectl get chaosengines/%s -n %s -o jsonpath="
+                                         "'{.status.engineStatus}'" % (engine_name, namespace))
+        engine_status = chaos_engine.strip()
+        if engine_counter >= max_tries:
+            logging.error("Chaos engine took longer than 5 minutes to be running or complete")
+            return False
+        engine_counter += 1
+        # need to see if error in run
+        if "notfound" in engine_status.lower():
+            logging.info("Chaos engine was not found")
+            return False
+
+    if not chaos_engine:
+        return False
+    chaos_result = runcommand.invoke("kubectl get chaosresult %s"
+                                     "-%s -n %s -o "
+                                     "jsonpath='{.status.experimentstatus.verdict}'"
+                                     % (engine_name, experiment_name, namespace))
+    result_counter = 0
+    status = chaos_result.strip()
+    while status == "Awaited":
+        logging.info("Waiting for chaos result to finish, sleeping 10 seconds")
+        time.sleep(10)
+        chaos_result = runcommand.invoke("kubectl get chaosresult %s"
+                                         "-%s -n %s -o "
+                                         "jsonpath='{.status.experimentstatus.verdict}'"
+                                         % (engine_name, experiment_name, namespace))
+        status = chaos_result.strip()
+        if result_counter >= max_tries:
+            logging.error("Chaos results took longer than 5 minutes to get a final result")
+            return False
+        result_counter += 1
+        if "notfound" in status.lower():
+            logging.info("Chaos result was not found")
+            return False
+
+    if status == "Pass":
+        return True
+    else:
+        chaos_result = runcommand.invoke("kubectl get chaosresult %s"
+                                         "-%s -n %s -o jsonpath="
+                                         "'{.status.experimentstatus.failStep}'" %
+                                         (engine_name, experiment_name, namespace))
+        logging.info("Chaos result failed information: " + str(chaos_result))
+        return False
+
+
+# Delete all chaos engines in a given namespace
+def delete_chaos(namespace):
+    runcommand.invoke("kubectl delete chaosengine --all -n " + str(namespace))
+    runcommand.invoke("kubectl delete chaosexperiment --all -n " + str(namespace))
+    runcommand.invoke("kubectl delete chaosresult --all -n " + str(namespace))
+
+
+# Uninstall litmus operator
+def uninstall_litmus(version):
+    runcommand.invoke("kubectl delete -f "
+                      "https://litmuschaos.github.io/litmus/litmus-operator-%s.yaml" % version)
--- a/run_kraken.py
+++ b/run_kraken.py
@@ -10,6 +10,7 @@ import requests
 import pyfiglet
 import kraken.kubernetes.client as kubecli
 import kraken.invoke.command as runcommand
+import kraken.litmus.common_litmus as common_litmus
 import kraken.node_actions.common_node_functions as nodeaction
 from kraken.node_actions.aws_node_scenarios import aws_node_scenarios
 from kraken.node_actions.general_cloud_node_scenarios import general_node_scenarios
@@ -229,6 +230,53 @@ def time_scenarios(scenarios_list, config):
                 publish_kraken_status(config, not_reset)
 
 
+def litmus_scenarios(scenarios_list, config, litmus_namespaces, litmus_uninstall):
+    # Loop to run the scenarios starts here
+    for l_scenario in scenarios_list:
+        try:
+            for item in l_scenario:
+                runcommand.invoke("kubectl apply -f %s" % item)
+                if "http" in item:
+                    f = requests.get(item)
+                    yaml_item = list(yaml.safe_load_all(f.content))[0]
+                else:
+                    with open(item, "r") as f:
+                        logging.info("opened yaml" + str(item))
+                        yaml_item = list(yaml.safe_load_all(f))[0]
+
+                if yaml_item['kind'] == "ChaosEngine":
+                    engine_name = yaml_item['metadata']['name']
+                    namespace = yaml_item['metadata']['namespace']
+                    litmus_namespaces.append(namespace)
+                    experiment_names = yaml_item['spec']['experiments']
+                    for expr in experiment_names:
+                        expr_name = expr['name']
+                        experiment_result = common_litmus.check_experiment(engine_name,
+                                                                           expr_name,
+                                                                           namespace)
+                        if experiment_result:
+                            logging.info("Scenario: %s has been successfully injected!"
+                                         % item)
+                        else:
+                            logging.info("Scenario: %s was not successfully injected!"
+                                         % item)
+                            if litmus_uninstall:
+                                for l_item in l_scenario:
+                                    logging.info('item ' + str(l_item))
+                                    runcommand.invoke("kubectl delete -f %s" % l_item)
+            if litmus_uninstall:
+                for item in l_scenario:
+                    logging.info('item ' + str(item))
+                    runcommand.invoke("kubectl delete -f %s" % item)
+            cerberus_integration(config)
+            logging.info("Waiting for the specified duration: %s" % wait_duration)
+            time.sleep(wait_duration)
+        except Exception as e:
+            logging.error("Failed to run litmus scenario: %s. Encountered "
+                          "the following exception: %s" % (item, e))
+    return litmus_namespaces
+
+
 # Main function
 def main(cfg):
     # Start kraken
@@ -242,6 +290,8 @@ def main(cfg):
         global kubeconfig_path, wait_duration
         kubeconfig_path = config["kraken"].get("kubeconfig_path", "")
         chaos_scenarios = config["kraken"].get("chaos_scenarios", [])
+        litmus_version = config['kraken'].get("litmus_version", 'v1.9.1')
+        litmus_uninstall = config['kraken'].get("litmus_uninstall", False)
         wait_duration = config["tunings"].get("wait_duration", 60)
         iterations = config["tunings"].get("iterations", 1)
         daemon_mode = config["tunings"].get("daemon_mode", False)
@@ -277,6 +327,8 @@ def main(cfg):
             iterations = int(iterations)
 
         failed_post_scenarios = []
+        litmus_namespaces = []
+        litmus_installed = False
         # Loop to run the chaos starts here
         while (int(iteration) < iterations):
             # Inject chaos scenarios specified in the config
@@ -298,9 +350,23 @@ def main(cfg):
                         # Inject time skew chaos scenarios specified in the config
                         elif scenario_type == "time_scenarios":
                             time_scenarios(scenarios_list, config)
+                        elif scenario_type == "litmus_scenarios":
+                            if not litmus_installed:
+                                common_litmus.install_litmus(litmus_version)
+                                common_litmus.deploy_all_experiments(litmus_version)
+                                litmus_installed = True
+                            litmus_namespaces = litmus_scenarios(scenarios_list, config,
+                                                                 litmus_namespaces,
+                                                                 litmus_uninstall)
 
             iteration += 1
             logging.info("")
+        if litmus_uninstall and litmus_installed:
+            for namespace in litmus_namespaces:
+                common_litmus.delete_chaos(namespace)
+            common_litmus.delete_experiments()
+            common_litmus.uninstall_litmus(litmus_version)
+
         if failed_post_scenarios:
             logging.error("Post scenarios are still failing at the end of all iterations")
             sys.exit(1)
--- a/scenarios/node_hog_engine.yaml
+++ b/scenarios/node_hog_engine.yaml
@@ -1,0 +1,25 @@
+apiVersion: litmuschaos.io/v1alpha1
+kind: ChaosEngine
+metadata:
+  name: nginx-chaos
+  namespace: default
+spec:
+  # It can be true/false
+  annotationCheck: 'false'
+  # It can be active/stop
+  engineState: 'active'
+  chaosServiceAccount: node-cpu-hog-sa
+  monitoring: false
+  # It can be delete/retain
+  jobCleanUpPolicy: 'delete'
+  experiments:
+    - name: node-cpu-hog
+      spec:
+        components:
+          env:
+            # set chaos duration (in sec) as desired
+            - name: TOTAL_CHAOS_DURATION
+              value: '60'
+            # ENTER THE COMMA SEPARATED TARGET NODES NAME
+            - name: TARGET_NODES
+              value: ''


scenarios/node_hog_engine.yaml
```diff
@@ -0,0 +1,26 @@
...
+      spec:
+        components:
+          env:
+            # set chaos duration (in sec) as desired
+            - name: TOTAL_CHAOS_DURATION
+              value: '60'
+
+             # ENTER THE NAME OF THE APPLICATION NODE
+            - name: APP_NODE
+              value: 'ip-10-0-139-222.us-east-2.compute.internal'
```
mffiedler, line 26: suggestion: replace with something generic like node_name

scenarios/node_hog.yaml
```diff
@@ -0,0 +1,69 @@
...
+    - name: TOTAL_CHAOS_DURATION
+      value: '60'
+
+    # Period to wait before and after injection of chaos in sec
+    - name: RAMP_TIME
+      value: '20'
+
+    # ENTER THE NAME OF THE APPLICATION NODE
+    - name: APP_NODE
+      value: 'ip-10-0-139-222.us-east-2.compute.internal'
```
mffiedler, line 57: suggestion: replace with something generic like node_name

scenarios/node_hog_engine.yaml
paigerube14: This is one of the given scenarios from litmus. Not sure if I can change it

mffiedler: That's fine - we should consume their scenarios as-is

kraken/litmus/common_litmus.py
```diff
@@ -0,0 +1,80 @@
...
+            return False
+
+    if not chaos_engine:
+        return False
+    chaos_result = runcommand.invoke("kubectl describe chaosresult %s"
+                                     "-%s -n %s | grep Verdict" %
+                                     (engine_name, experiment_name, namespace))
+    status = chaos_result.split(": ")[1].strip()
+    while status == "Awaited":
+        logging.info("Waiting for chaos result to finish waiting, sleeping 10 seconds")
```
mffiedler, line 56: Not a big deal, but this is chatty on the console for longer tests - get this repeatedly.    Maybe something to follow up on as we see how other Litmus scenarios run in kraken.

kraken/litmus/common_litmus.py
```diff
@@ -0,0 +1,80 @@
...
+            logging.error("Litmus pod had error creating")
+            sys.exit(1)
+
+
+# Check status of experiment
+def check_experiment(engine_name, experiment_name, namespace):
+    chaos_engine = runcommand.invoke("kubectl describe chaosengines/%s -n %s | grep "
+                                     "Status" % (engine_name, namespace))
+    engine_status = chaos_engine.split(": ")[-1].strip()
+    while engine_status != "Running" and engine_status != "Completed":
```
mffiedler, line 39: We should have a max_retry count here.  I've seen cases where the experiment or engine are mis-configured and we just wait for ever.

run_kraken.py
```diff
@@ -246,9 +252,51 @@ def main(cfg):
...
+                        publish_kraken_status(config, failed_post_scenarios)
+                    except Exception as e:
+                        logging.error("Failed to run litmus scenario: %s. Encountered "
+                                      "the following exception: %s" % (item, e))
             iteration += 1
             logging.info("")
+        if litmus_scenarios:
+            for namespace in litmus_namespaces:
+                common_litmus.delete_experiments(namespace)
+            common_litmus.uninstall_litmus()
```
mffiedler, line 298: can we catch SIGINT (ctrl-c) and clean up Litmus?   I think we're doing that in Cerberus, possibly for slack.   This would help when a) things don't go well with a scenario and it is stuck and b) when running in daemon mode.    We could do this as a follow on PR - not a must to merge this one.

kraken/litmus/common_litmus.py
```diff
@@ -0,0 +1,92 @@
...
+    yaml_pod = yaml.full_load(pods_status)
+    logging.info('len of pod items' + str(len(yaml_pod['items'])))
+    for pod in yaml_pod['items']:
+        return pod['status']['phase']
+
+
+# Install litmus and wait until pod is running
+def install_litmus():
+    runcommand.invoke("kubectl apply -f "
+                      "https://litmuschaos.github.io/litmus/litmus-operator-v1.8.2.yaml")
```
machacekondra, line 19: Not a need now, but it would be nice to have this somehow parametrized, because updates are quite often in litmus.

kraken/litmus/common_litmus.py
```diff
@@ -0,0 +1,92 @@
...
+        logging.info("ready status for pod" + str(ready_status))
+        ready_status = get_pod_ready_status()
+        if "Error" in ready_status:
+            logging.error("Litmus pod had error creating")
+            sys.exit(1)
+
+
+# Check status of experiment
+def check_experiment(engine_name, experiment_name, namespace):
+    chaos_engine = runcommand.invoke("kubectl describe chaosengines/%s -n %s | grep "
```
machacekondra, line 36: Not sure how backward compatible the describe output is, but wouldn't it be better to use something like:
```
$ kubectl get chaosengine my-chaos -o jsonpath='{.status.engineStatus}'
Completed
```

kraken/litmus/common_litmus.py
```diff
@@ -0,0 +1,92 @@
...
+    engine_status = chaos_engine.split(": ")[-1].strip()
+    max_tries = 2
+    engine_counter = 0
+    while engine_status != "Running" and engine_status != "Completed":
+        time.sleep(10)
+        logging.info("Sleeping waiting for engine to start running")
+        chaos_engine = runcommand.invoke("kubectl describe chaosengines/%s -n %s | grep "
+                                         "Status" % (engine_name, namespace))
+        engine_status = chaos_engine.split(": ")[-1].strip()
+        if engine_counter >= max_tries:
```
machacekondra, line 47: Not sure if 30seconds is enough for every experiment. Would be nice to have it configurable in future.

kraken/litmus/common_litmus.py
```diff
@@ -0,0 +1,92 @@
...
+        return True
+    else:
+        return False
+
+
+# Delete all chaos engines in a given namespace
+def delete_experiments(namespace):
+    runcommand.invoke("kubectl delete chaosengine --all -n " + str(namespace))
+    runcommand.invoke("kubectl delete chaosexperiment --all -n " + str(namespace))
+    runcommand.invoke("kubectl delete chaosresult --all -n " + str(namespace))
```
machacekondra, line 86: I think you forgot to cleanup service accounts here.

kraken/litmus/common_litmus.py
```diff
@@ -0,0 +1,92 @@
...
+                                         (engine_name, experiment_name, namespace))
+        status = chaos_result.split(": ")[1].strip()
+        if result_counter >= max_tries:
+            logging.error("Chaos results took longer than 5 minutes to get a final result")
+            return False
+        result_counter += 1
+        if "not found" in status:
+            return False
+
+    if status == "Pass":
```
machacekondra, line 76: It would be nice to take into consideration/results also litmus probes state, in future.

kraken/litmus/common_litmus.py
paigerube14: That's my bad, I was testing out the max_retries and forgot to reset. The max_retries should be 30. Updating

kraken/litmus/common_litmus.py
paigerube14: The service accounts should be cleaned up here:  https://github.com/openshift-scale/kraken/blob/4b83f67379b058446b8fce7f5040ed61cbd43476/run_kraken.py#L285

Is that correct?

run_kraken.py
paigerube14: Yes definitely think that this would be nice to add. I might do it separate from this PR just to be able to get it in quickly

kraken/litmus/common_litmus.py
paigerube14: Updated

kraken/litmus/common_litmus.py
paigerube14: Thoughts on adding the version of litmus into the configuration file to be able to update on your own but leave a working version in the master branch.

kraken/litmus/common_litmus.py
paigerube14: Updated

config/config.yaml
```diff
@@ -10,12 +11,16 @@ kraken:
           -    scenarios/post_action_regex.py
     node_scenarios:                                        # List of chaos node scenarios to load
         -   scenarios/node_scenarios_example.yml
-
+    litmus_scenarios:
```
chaitanyaenr, line 14: We might want to add documentation around litmus integration.

scenarios/node_hog.yaml
```diff
@@ -0,0 +1,69 @@
...
+          - "update"
+          - "delete"
+      - apiGroups:
+          - ""
+        resources:
+          - "nodes"
+        verbs:
+          - "get"
+          - "list"
+    image: "litmuschaos/go-runner:1.8.2"
```
machacekondra, line 39: It would be better to use the experiments from the github, because this version may outdate and not equal to version of litmus installed. So maybe would be better to install experiments with litmus from: https://hub.litmuschaos.io/api/chaos/1.9.0?file=charts/generic/experiments.yaml

paigerube14: Took a look at this yesterday. I agree that this would be better to install the experiments from the versioned url's. 

From my understanding I have to edit the experiment once I apply it to give it the correct node name for this specific experiment I have been working on. Is that correct?

docs/litmus_scenarios.md
```diff
@@ -0,0 +1,18 @@
+### Litmus Scenarios
+Kraken consumes [Litmus](https://github.com/litmuschaos/litmus) under the hood for some infrastructure, pod, and node scenarios 
+
+#### Litmus chaos scenarios
+There are 3 yaml files needed to properly execute each litmus scenario. Below is a description of each file needed
```
masayag, line 5: ChaosResult is not needed to be created - it will be created by litmus itself and updated by litmus.
The only responsibility will be to inspect the reported status and to clear the resource.

kraken/litmus/common_litmus.py
```diff
@@ -0,0 +1,97 @@
+import kraken.invoke.command as runcommand
+import yaml
+import logging
+import time
+import sys
+
+
+def get_pod_ready_status():
+    pods_status = runcommand.invoke("kubectl get pods -n litmus -o yaml")
```
masayag, line 9: I'd check the readiness of litmus by querying the deployment status since it is deployed as an operator, e.g.:
```
oc wait deploy -n litmus chaos-operator-ce --for=condition=Available --timeout=30s
```

run_kraken.py
```diff
@@ -223,6 +262,7 @@ def main(cfg):
         global kubeconfig_path, wait_duration
         kubeconfig_path = config["kraken"].get("kubeconfig_path", "")
         chaos_scenarios = config["kraken"].get("chaos_scenarios", [])
+        litmus_version = config['kraken'].get("litmus_version", 'v1.8.2')
```
masayag, line 265: v1.9.0 ?

scenarios/node_hog.yaml
masayag: the node name can be provided via `APP_NODE` env var in the ChaosEngine resource
See an example of the ChaosEngine that provides it (with blank value in the example): https://docs.litmuschaos.io/docs/node-cpu-hog/#sample-chaosengine-manifest

Once you set it via the ChaosEngine, it will override the value in the ChaosExperiment and be used.
That will allow you to rely on the formal released experiments which is aligned with litmus operator version.

masayag (changes requested):

config/config.yaml
```diff
@@ -1,6 +1,7 @@
 kraken:
-    kubeconfig_path: /root/.kube/config                    # Path to kubeconfig
+    kubeconfig_path: /Users/prubenda/.kube/config                    # Path to kubeconfig
     exit_on_failure: False                                 # Exit when a post action scenario fails
+    litmus_version: v1.9.0
```
jordigilh, line 4: s/v1.9.0/v.1.9.1 perhaps? I see a reference to v1.9.1 in line 18

docs/litmus_scenarios.md
```diff
@@ -0,0 +1,20 @@
+### Litmus Scenarios
+Kraken consumes [Litmus](https://github.com/litmuschaos/litmus) under the hood for some infrastructure, pod, and node scenarios 
+
+#### Litmus chaos scenarios
+There are 3 custom resources that are created during each litmus scenario. Below is a description of the resources:
+* ChaosEngine: A resource to link a Kubernetes application or Kubernetes node to a ChaosExperiment. ChaosEngine is watched by Litmus' Chaos-Operator which then invokes Chaos-Experiments
+* ChaosExperiment: A resource to group the configuration parameters of a chaos experiment. ChaosExperiment CRs are created by the operator when experiments are invoked by ChaosEngine.
+* ChaosResult : A resource to hold the results of a chaos-experiment. The Chaos-exporter reads the results and exports the metrics into a configured Prometheus server.
+
+The
```
jordigilh, line 10: I think you might want to merge this line and the following one.

run_kraken.py
```diff
@@ -223,6 +268,7 @@ def main(cfg):
         global kubeconfig_path, wait_duration
         kubeconfig_path = config["kraken"].get("kubeconfig_path", "")
         chaos_scenarios = config["kraken"].get("chaos_scenarios", [])
+        litmus_version = config['kraken'].get("litmus_version", 'v1.9.0')
```
jordigilh, line 271: s/v.1.9.0/v.1.9.1/
In line [18](https://github.com/openshift-scale/kraken/pull/33/files?diff=split&w=1#diff-38df760413887d0cc4ee5707919e6390456a863d326f6080644cc3d9c72e4bbaR18) you are applying version v.1.9.1

scenarios/node_hog.yaml
```diff
@@ -0,0 +1,69 @@
...
+      value: 'ip-10-0-139-222.us-east-2.compute.internal'
+
+    # PROVIDE THE LIB HERE
+    # ONLY LITMUS SUPPORTED
+    - name: LIB
+      value: 'litmus'
+
+    # provide lib image
+    - name: LIB_IMAGE
+      value: 'litmuschaos/go-runner:1.8.2'
```
jordigilh, line 66: upgrade to 1.9.1?

run_kraken.py
```diff
@@ -210,6 +212,49 @@ def time_scenarios(scenarios_list, config):
...
+                                         % item)
+                        else:
+                            logging.info("Scenario: %s was not successfully injected!"
+                                         % item)
+            for item in l_scenario:
+                logging.info('item ' + str(item))
+                runcommand.invoke("kubectl delete -f %s" % item)
+            cerberus_integration(config)
+            logging.info("Waiting for the specified duration: %s" % wait_duration)
+            time.sleep(wait_duration)
```
jordigilh, line 251: Why not wait for a condition rather than wait for a fix amount of time? Otherwise how do you know if the condition has been met after the wait_duration?

kraken/litmus/common_litmus.py
```diff
@@ -0,0 +1,98 @@
...
+
+# Install litmus and wait until pod is running
+def install_litmus(version):
+    runcommand.invoke("kubectl apply -f "
+                      "https://litmuschaos.github.io/litmus/litmus-operator-%s.yaml" % version)
+
+    runcommand.invoke("oc wait deploy -n litmus chaos-operator-ce --for=condition=Available --timeout=30s")
+
+
+def perform_experiment(version_string):
```
jordigilh, line 16: More than perform, I'd say deploy or install since you're adding the generic experiment CRs, but there is no action taken that could be interpreted as an execution of the experiment.

kraken/litmus/common_litmus.py
```diff
@@ -0,0 +1,98 @@
...
+        logging.info("Waiting for chaos result to finish waiting, sleeping 10 seconds")
+        time.sleep(10)
+        chaos_result = runcommand.invoke("kubectl get chaosresult %s"
+                                         "-%s -n %s -o "
+                                         "jsonpath='{.status.experimentstatus.verdict}'"
+                                         % (engine_name, experiment_name, namespace))
+        status = chaos_result.strip()
+        if result_counter >= max_tries:
+            logging.error("Chaos results took longer than 5 minutes to get a final result")
+            return False
```
jordigilh, line 71: I wonder if it's safe to exit here and leave the test running. Cleaning up before starting the next tests might be worth considering, since the failed test could impact the outcome of the subsequent tests (ongoing node disruption, for instance). WDYT?

kraken/litmus/common_litmus.py
```diff
@@ -0,0 +1,98 @@
...
+import yaml
+import logging
+import time
+import sys
+
+
+# Install litmus and wait until pod is running
+def install_litmus(version):
+    runcommand.invoke("kubectl apply -f "
+                      "https://litmuschaos.github.io/litmus/litmus-operator-%s.yaml" % version)
```
jordigilh, line 11: you might want to consider disabling the [telemetry](https://docs.litmuschaos.io/docs/faq-general/#does-litmus-track-any-usage-metrics-on-the-test-clusters) that is being collected for the operator:
`  oc patch -n litmus deployment.apps/chaos-operator-ce --type=json --patch '
   [
     {
       "op": "add",
       "path": "/spec/template/spec/containers/0/env/-",
       "value": {
         "name": "ANALYTICS",
         "value": "FALSE"
       }
     }
   ]'`

kraken/litmus/common_litmus.py
```diff
@@ -0,0 +1,98 @@
...
+
+    runcommand.invoke("oc wait deploy -n litmus chaos-operator-ce --for=condition=Available --timeout=30s")
+
+
+def perform_experiment(version_string):
+
+    if version_string.startswith("v"):
```
jordigilh, line 18: why not stick to a version format and fail otherwise? it will give more clarity to have a single version format rather than expect multiple ones.

scenarios/node_hog.yaml
jordigilh, line 66: s/1.8.2/1.9.1

config/config.yaml
```diff
@@ -1,6 +1,7 @@
 kraken:
-    kubeconfig_path: /root/.kube/config                    # Path to kubeconfig
+    kubeconfig_path: /Users/prubenda/.kube/config                    # Path to kubeconfig
```
chaitanyaenr, line 2: s/prubenda/root or $HOME

kraken/litmus/common_litmus.py
```diff
@@ -0,0 +1,98 @@
...
+# Check status of experiment
+def check_experiment(engine_name, experiment_name, namespace):
+    chaos_engine = runcommand.invoke("kubectl get chaosengines/%s -n %s -o jsonpath="
+                                     "'{.status.engineStatus}'" % (engine_name, namespace))
+    engine_status = chaos_engine.strip()
+    max_tries = 30
+    engine_counter = 0
+    while engine_status.lower() != "running" and engine_status.lower() != "completed":
+        time.sleep(10)
+        logging.info("Sleeping waiting for engine to start running.")
```
chaitanyaenr, line 40: s/Sleeping waiting for engine to start running/Waiting for the ChaosEngine to start running.

docs/litmus_scenarios.md
```diff
@@ -0,0 +1,20 @@
+### Litmus Scenarios
+Kraken consumes [Litmus](https://github.com/litmuschaos/litmus) under the hood for some infrastructure, pod, and node scenarios 
+
+#### Litmus chaos scenarios
+There are 3 custom resources that are created during each litmus scenario. Below is a description of the resources:
+* ChaosEngine: A resource to link a Kubernetes application or Kubernetes node to a ChaosExperiment. ChaosEngine is watched by Litmus' Chaos-Operator which then invokes Chaos-Experiments
+* ChaosExperiment: A resource to group the configuration parameters of a chaos experiment. ChaosExperiment CRs are created by the operator when experiments are invoked by ChaosEngine.
+* ChaosResult : A resource to hold the results of a chaos-experiment. The Chaos-exporter reads the results and exports the metrics into a configured Prometheus server.
```
chaitanyaenr, line 8: Thoughts on using the in-cluster prometheus by default to ship the metrics in case of OpenShift?

run_kraken.py
```diff
@@ -279,9 +327,22 @@ def main(cfg):
...
+                                litmus_installed = True
+                            litmus_namespaces = litmus_scenarios(scenarios_list, config,
+                                                                 litmus_namespaces)
 
             iteration += 1
             logging.info("")
+        if litmus_installed:
+            for namespace in litmus_namespaces:
+                common_litmus.delete_chaos(namespace)
+            common_litmus.delete_experiments()
```
chaitanyaenr, line 343: We might want to make this optional with the default set to False, it's useful to keep them around for debugging. Thoughts?

config/config.yaml
```diff
@@ -10,9 +11,12 @@ kraken:
             -   scenarios/node_scenarios_example.yml
         -   pod_scenarios:
             - -    scenarios/openshift-apiserver.yml
-            - -    scenarios/openshift-kube-apiserver.yml
+             - -    scenarios/openshift-kube-apiserver.yml
         -   time_scenarios:                                # List of chaos time scenarios to load
             - scenarios/time_scenarios_example.yml
+        -   litmus_scenarios:
+            - - https://hub.litmuschaos.io/api/chaos/1.9.1?file=charts/generic/node-cpu-hog/rbac.yaml
```
chaitanyaenr, line 18: Maybe consume the rbac file local to scenarios dir to be able to modify the namespace? Thoughts?

scenarios/node_hog.yaml
```diff
@@ -0,0 +1,69 @@
+apiVersion: litmuschaos.io/v1alpha1
+description:
+  message: |
+    Give a cpu spike on a node belonging to a deployment
+kind: ChaosExperiment
```
masayag, line 5: This CR is being installed by https://github.com/openshift-scale/kraken/pull/33/files#diff-1f4248c56fcac8c6a639e9812e8fd51cb324a8cbc3aa79b34a7d38aa90451775R23 which installs all of the experiments.
This file can be removed and the env vars should be set via the ChaosEngine resource.

run_kraken.py
paigerube14: This is just a wait time between scenarios to give the cluster a break

run_kraken.py
paigerube14: Yeh I think that is a good idea, adding an extra flag in the config to delete or not based on the run status

config/config.yaml
paigerube14: I'm not sure which is better here. Might be good to keep it has the github link to keep it up to date but could see it both ways

kraken/litmus/common_litmus.py
paigerube14: good idea! I had added this incase we wanted to use "latest" as the version. What are your thoughts here? Should I give the option to use latest

scenarios/node_hog.yaml
paigerube14: Going to remove this file since we are deploying the experiments using the github link

run_kraken.py
jordigilh: what do you mean by a break (cpu utilization, network IO)? Is there any way you can synchronize this wait to a signal (deployment of a pod completed, or something else)?
Waiting is best to link it to a condition, otherwise it just delays the execution and can render flaky tests results.

kraken/litmus/common_litmus.py
jordigilh: Using latest could lead to a situation where we test against a major release that render some existing tests invalid due to changes in the CRD schema, for instance. Controlling which version is deployed can also help to reduce false positives by removing litmus code from the potential changes in the test execution.

scenarios/node_hog.yaml
paigerube14: Removed file

kraken/litmus/common_litmus.py
paigerube14: I agree. I added a cleanup of the scenario if the test fails

kraken/litmus/common_litmus.py
paigerube14: Update to "deploy_all_experiments". Thoughts here?

run_kraken.py
paigerube14: If you take a look at the configuration (tunings-> wait_duration (https://github.com/paigerube14/kraken/blob/320c9531bf7b4b0a3df8202a4290abbf12a1a208/config/config.yaml#L26)) we want to wait in-between executions of separate scenarios. This should have no effect on the test results

kraken/litmus/common_litmus.py
paigerube14: Added after the installation of litmus

docs/litmus_scenarios.md
paigerube14: I think this is built into litmus. Would have to be on their side? Would need some clarification on this part

paigerube14:
I think I have updated this PR with all comments, @jordigilh are you able to take another look?

jordigilh:
LGTM

Thanks!

/Jordi

config/config.yaml
mffiedler: Sorry for late comment.   GitHub link has disadvantage of not working in disconnected clusters.   Probably not a common scenario and not a blocker for this PR.   But does using the link make the node_hog_rbac.yaml file superfluous?  Or we should keep it around?

scenarios/node_hog_rbac.yaml
```diff
@@ -0,0 +1,37 @@
+---
```
mffiedler, line 1: Is this file still needed since the config is now using the URL?

scenarios/node_hog_engine.yaml
```diff
@@ -0,0 +1,26 @@
...
+    - name: node-cpu-hog
+      spec:
+        components:
+          env:
+            # set chaos duration (in sec) as desired
+            - name: TOTAL_CHAOS_DURATION
+              value: '60'
+
+             # ENTER THE NAME OF THE APPLICATION NODE
+            - name: APP_NODE
```
mffiedler, line 25: I think this env var needs to be TARGET_NODES (see: https://docs.litmuschaos.io/docs/node-cpu-hog/#prepare-chaosengine)   @jordigilh Can you confirm?   Even when I use TARGET_NODES, though, it is not honored and the load runs on a different node.

mffiedler (changes requested):

scenarios/node_hog_engine.yaml
paigerube14: It looks like the key was updated between versions 1.9 and 1.10. In 1.9 documentation it shows APP_NODES still (https://docs.litmuschaos.io/docs/1.9.0/node-cpu-hog/) 

Seems like the versions are updated quite often. Do we want to just pick a version and keep using it until we have a need to update. Should we use 1.10 now since it's the latest?

scenarios/node_hog_rbac.yaml
paigerube14: Should we keep the experiment file (had previously deleted from this PR) and this file for the disconnected clusters situation you were talking about in your previous comment? Or better to just call it out in the documentation that you'll have to download and use the files of whatever version they want so it doesn't become super out of date?

mffiedler: That was my confusion, I needed to down load the experiment file and add it to config.yaml (it is gone from there too).

-  wget https://hub.litmuschaos.io/api/chaos/1.9.1?file=charts/generic/node-cpu-hog/experiment.yaml -O scenarios/experiment.yaml
- edit config/config.yml and add experiment.yaml after node_hog_engine.yaml
- TARGET_NODES worked fine for 1.9.1 - that was in the sample experiment file 

Our sample config.yaml should probably work OOTB and include a local experiment file and local engine file since they are most likely to be updated.  If not, we need to include instructions in the Litmus README for downloading scenarios and adding them to the config.   My opinion, anyways.    @jordigilh @chaitanyaenr other opinions.

scenarios/node_hog_engine.yaml
jordigilh: @paigerube14 confirmed it should be [named](https://github.com/litmuschaos/litmus-go/blob/master/pkg/generic/node-cpu-hog/environment/environment.go#L32)  `TARGET_NODES`.
They also changed the structure of the probes in the schema. Since you don't use probes in this case, you should be good to move to 1.10.0.

scenarios/node_hog_rbac.yaml
paigerube14: From my understanding, the code will install all experiments based on the version you specify in the overarching config file. So no need to do the wget or have the experiment files locally. 
Setting the TARGET_NODES in ChaosEngine file should set an environment variable to properly set the variables for the specific experiment file with the given target node name.

Working on validating this and updating the documentation file for litmus to be a little more clear on the steps needed.

jordigilh: Yes, you can keep this locally for disconnected clusters, but then you'll have to make sure that the spec matches the litmus go runner version, particularly when you upgrade the litmus version.
You only need to apply the `chaosengine` that will be used for this particular experiment, and discard the others.

paigerube14: Removed the rbac file for the node_hog scenario

scenarios/node_hog_engine.yaml
paigerube14: Updated config file to use the 1.10.0 versions.

scenarios/node_hog_rbac.yaml
jordigilh: I think you'll need the rbac file to define the sa and the role an role binding. Unless you are handling that elsewhere :)

paigerube14: I meant removed the downloaded rbac file from the scenarios folder. Still using the hub.litmus link in the config to set the sa, role and role binding

jordigilh: ok sorry, my bad :)

mffiedler:
@paigerube14 LGTM.  Let's squash and merge and iterate in subsequent PRs/Issue

PR #37, opened 2020-10-21
Time skew

paigerube14:
PR for https://github.com/openshift-scale/kraken/issues/22

This PR allows the user to reset the date or time on nodes and pods and validates that the time gets reset properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,12 +16,13 @@ Instructions on how to setup the config and the options supported can be found a
 
 
 ### Kubernetes/OpenShift chaos scenarios supported
-Kraken supports pod and node based scenarios.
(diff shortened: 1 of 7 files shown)


mffiedler (approved):
LGTM - code looks good + tested both pod and node scenarios and the worked well.

docs/time_skew.rd
```diff
@@ -0,0 +1,29 @@
+###  Time/Date Skew Scenarios
```
chaitanyaenr, line 1: we need to rename this file to time_scenarios.md to match the one being used in README.md

paigerube14: Updated.

chaitanyaenr (approved):
LGTM

PR #38, opened 2020-10-21
Adds the ability to run a specific type of scenario multiple times

yashashreesuresh:
With the current implementation, all the scenarios of specific type
(for example, pod scenario) has to be executed together. All
pod_scenarios are followed by node_scenarios and so on.
(pod_scenarios -> node_scenarios -> pod_scenarios is not possible)
This commit enables the user to run a specific type of scenario
multiple times. For example, few pod_scenarios followed by
node_scenarios followed by few pod_scenarios.

### DIFF
--- a/run_kraken.py
+++ b/run_kraken.py
@@ -152,6 +152,61 @@ def post_actions(kubeconfig_path, scenario, failed_post_scenarios, pre_action_ou
     return failed_post_scenarios
 
 
+def pod_scenarios(scenarios_list, config, failed_post_scenarios):
+    try:
+        # Loop to run the scenarios starts here
+        for pod_scenario in scenarios_list:
+            if len(pod_scenario) > 1:
+                pre_action_output = run_post_action(kubeconfig_path, pod_scenario[1])
+            else:
+                pre_action_output = ''
+            runcommand.invoke("powerfulseal autonomous --use-pod-delete-instead-of-ssh-kill"
+                              " --policy-file %s --kubeconfig %s --no-cloud"
+                              " --inventory-kubernetes --headless"
+                              % (pod_scenario[0], kubeconfig_path))
+
+            logging.info("Scenario: %s has been successfully injected!" % (pod_scenario[0]))
+            logging.info("Waiting for the specified duration: %s" % (wait_duration))
+            time.sleep(wait_duration)
+
+            failed_post_scenarios = post_actions(kubeconfig_path, pod_scenario,
+                                                 failed_post_scenarios, pre_action_output)
+            publish_kraken_status(config, failed_post_scenarios)
+    except Exception as e:
+        logging.error("Failed to run scenario: %s. Encountered the following "
+                      "exception: %s" % (pod_scenario[0], e))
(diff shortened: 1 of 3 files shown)


yashashreesuresh:
cc: @chaitanyaenr

paigerube14:
This might be a good time to also separate out each different type of scenario into its own function. This might make it easier to read and add the many more types of scenarios that are being worked on as well. Thoughts?? @yashashreesuresh @chaitanyaenr @mffiedler

mffiedler:
> This might be a good time to also separate out each different type of scenario into its own function. This might make it easier to read and add the many more types of scenarios that are being worked on as well. Thoughts?? @yashashreesuresh @chaitanyaenr @mffiedler

You read my mind.   With PR #37  and this one, run_kraken.py main() is  getting noisy.   I don't think it has to be done in this PR and is probably better done after #37 merges and this merges when it is ready.   I've opened issue #39 for this.

paigerube14:
LGTM

yashashreesuresh:
@paigerube14 @mffiedler @chaitanyaenr 
Hey! I have updated the code and I have also separated different scenarios into their own functions.
However, I was not able to test the scenarios completely after the code change as I do not have a cluster.
The code needs to be tested, sorry for the inconvenience.

paigerube14:
I was able to run this and it works well! It looks like you might have pushed the kraken.report file, should we delete that?

run_kraken.py
```diff
@@ -152,6 +152,60 @@ def post_actions(kubeconfig_path, scenario, failed_post_scenarios, pre_action_ou
...
+                              " --inventory-kubernetes --headless"
+                              % (pod_scenario[0], kubeconfig_path))
+
+            logging.info("Scenario: %s has been successfully injected!" % (pod_scenario[0]))
+            logging.info("Waiting for the specified duration: %s" % (wait_duration))
+            time.sleep(wait_duration)
+
+            failed_post_scenarios = post_actions(kubeconfig_path, pod_scenario,
+                                                 failed_post_scenarios, pre_action_output)
+            publish_kraken_status(config, failed_post_scenarios)
```
paigerube14, line 174: Currently you an error when running the first: `scenarios/etcd.yml. Encountered the following exception: local variable 'failed_post_scenarios' referenced before assignment`
I think here we need to return the failed_post_scenarios so that they can be rerun later if they do fail. This would only have the last failure I believe. 

failed_post_scenarios also needs to be added as an argument and appended to instead of resetting.

yashashreesuresh: Ah, sorry. I have updated the code, PTAL. Thank you!

chaitanyaenr:
@yashashreesuresh tried running pod-scenarios -> node-scenarios ->pod-scenarios and it worked perfectly.

chaitanyaenr (approved):
LGTM

paigerube14:
LGTM

run_kraken.py
paigerube14: Looks awesome thanks so much!!

PR #40, opened 2020-10-27
Az nodes

paigerube14:
Adding node scenarios for clusters on Azure

I had to reorganize part of _inject_node_scenario_. It seemed that when azure VM's are stopped they are in a different state than for AWS so for the stop start scenario the Vm's would never restart.

### DIFF
--- a/docs/node_scenarios.md
+++ b/docs/node_scenarios.md
@@ -39,6 +39,24 @@ The supported node level chaos scenarios on an OPENSTACK cloud are `node_stop_st
 
 To execute the scenario, ensure the value for `ssh_private_key` in the node scenarios config file is set with the correct private key file path for ssh connection to the helper node. Ensure passwordless ssh is configured on the host running Kraken and the helper node to avoid connection errors.
 
+#### Azure
+
+**NOTE**: For Azure node killing scenarios, make sure [Azure CLI](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest) is installed
+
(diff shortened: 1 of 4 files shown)


docs/node_scenarios.md
```diff
@@ -13,11 +13,13 @@ Following node chaos scenarios are supported:
 
 **NOTE**: If the node doesn't recover from the node_crash_scenario injection, reboot the node to get it back to Ready state.
 
-**NOTE**: node_start_scenario, node_stop_scenario, node_stop_start_scenario, node_termination_scenario, node_reboot_scenario and stop_start_kubelet_scenario are supported only on AWS as of now.
+**NOTE**: node_start_scenario, node_stop_scenario, node_stop_start_scenario, node_termination_scenario, node_reboot_scenario and stop_start_kubelet_scenario are supported only on AWS and Azure as of now.
```
mffiedler, line 16: Before the list of NOTEs, document that the valid values for cloud_type are now aws, gcp and az

mffiedler:
https://github.com/powerfulseal/powerfulseal/pull/314 has merged so I will re-test this with latest upstream PowerfulSeal

paigerube14:
@mffiedler retested the start stop and reboot scenarios today on an Azure cluster. Things looked good to me, let me know what you think.

mffiedler:
@paigerube14 please fix the last doc nit and squash and I think this is good to go.   My tests worked with a clean virtualenv.   I had to dnf install the python3-devel package for one of the powerfulseal prereqs to install, but that can be a separate doc issue for the prereqs for Kraken.

mffiedler:
Thanks @paigerube14

PR #42, opened 2020-10-29
Display the pod scenarios logs

chaitanyaenr:
The current implementation just reports the pass/fail when
running the pod scenarios. This commit prints the pod scenarios logs
to stdout to give users the ability to understand the actions run by
powerfulseal under the hood - number of pods matching the filter,
pods picked for killing, post checks etc. This also helps in
debugging when the scenarios fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ dist: xenial
 
 langauge: python
 
-python: "3.6"
+python: "3.6.8"
 
 before_install:
   - sudo apt-get update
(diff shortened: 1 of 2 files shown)


paigerube14:
LGTM (other than the check that is failing)

chaitanyaenr:
@paigerube14 there seems to be some dependency issues with travis environment causing the CI to fail.

paigerube14:
LGTM

PR #50, opened 2020-11-11
Adding a general cloud provider file for kubelet reset and fork bomb on all clouds

paigerube14:
This PR will allow a user to specify there correct cloud type for node scenarios to allow them to run the stop start kubelet scenario as well as the fork bomb scenario.

This is for https://github.com/openshift-scale/kraken/issues/46

### DIFF
--- a/docs/node_scenarios.md
+++ b/docs/node_scenarios.md
@@ -28,7 +28,9 @@ A google service account is required to give proper authentication to GCP for no
  
 After creating the service account you'll need to enable the account using the following: ```export GOOGLE_APPLICATION_CREDENTIALS="<serviceaccount.json>"```
 
-**NOTE**: The `stop_start_kubelet_scenario` and `node_crash_scenario` scenarios are supported as they are independent of the cloud platform.
+**NOTE**: The `node_crash_scenario` and `stop_kubelet_scenario` scenario is supported independent of the cloud platform.
+
+Use 'generic' or do not add the 'cloud_type' key to your scenario if your cluster is not set up using one of the current supported cloud types
 
 
 Node scenarios can be injected by placing the node scenarios config files under node_scenarios option in the kraken config. Refer to [node_scenarios_example](https://github.com/openshift-scale/kraken/blob/master/scenarios/node_scenarios_example.yml) config file.
@@ -51,4 +53,10 @@ node_scenarios:
     instance_kill_count: 1
     timeout: 120
     cloud_type: aws
+  - actions:
+    - node_crash_scenario
+    node_name:
+    label_selector: node-role.kubernetes.io/infra
+    instance_kill_count: 1
+    timeout: 120
 ```
--- a/run_kraken.py
+++ b/run_kraken.py
@@ -39,6 +39,7 @@ def get_node_scenario_object(node_scenario):
 
 # Inject the specified node scenario
 def inject_node_scenario(action, node_scenario, node_scenario_object):
+    generic_cloud_scenarios = ("stop_kubelet_scenario", "node_crash_scenario")
     # Get the node scenario configurations
     instance_kill_count = node_scenario.get("instance_kill_count", 1)
     node_name = node_scenario.get("node_name", "")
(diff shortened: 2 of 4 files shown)


chaitanyaenr (approved):
LGTM

run_kraken.py
```diff
@@ -22,6 +23,8 @@ def get_node_scenario_object(node_scenario):
         return aws_node_scenarios()
     elif node_scenario['cloud_type'] == 'gcp':
         return gcp_node_scenarios()
+    else:
```
mffiedler, line 26: Nitpick: I think defaulting to 'generic' if cloud_type is not specified is fine, but let's not allow arbitrary values.   Something like:
if node_scenario['cloud_type'] == 'generic' or not 'cloud_type' in node_scenario:
   generic
else:
  bad cloud type

paigerube14: Think that is much better! Added some documentation to show this as well

run_kraken.py
```diff
@@ -12,16 +12,24 @@
...
 
 # Get the node scenarios object of specfied cloud type
 def get_node_scenario_object(node_scenario):
+    if "cloud_type" not in node_scenario.keys() or node_scenario['cloud_type'] == "generic":
+        return general_node_scenarios()
     if node_scenario['cloud_type'] == 'aws':
         return aws_node_scenarios()
     elif node_scenario['cloud_type'] == 'gcp':
         return gcp_node_scenarios()
+    else:
```
mffiedler, line 28: Still using arbitrary values as 'generic'.  Missed a file on the last commit?

paigerube14: Sorry about that, somehow missed the main file in my last commit! Should be there now

mffiedler:
I don't think we can use generic for stop_start_kubelet_scenario.   Starting the kubelet uses the reboot scenario which requires a cloud provider:

```
    def stop_start_kubelet_scenario(self, instance_kill_count, node, timeout):
        logging.info("Starting stop_start_kubelet_scenario injection")
        self.stop_kubelet_scenario(instance_kill_count, node, timeout)
        self.node_reboot_scenario(instance_kill_count, node, timeout)
        logging.info("stop_start_kubelet_scenario has been successfully injected!")
```

paigerube14:
> I don't think we can use generic for stop_start_kubelet_scenario. Starting the kubelet uses the reboot scenario which requires a cloud provider:
> 
> ```
>     def stop_start_kubelet_scenario(self, instance_kill_count, node, timeout):
>         logging.info("Starting stop_start_kubelet_scenario injection")
>         self.stop_kubelet_scenario(instance_kill_count, node, timeout)
>         self.node_reboot_scenario(instance_kill_count, node, timeout)
>         logging.info("stop_start_kubelet_scenario has been successfully injected!")
> ```

Ah yes. It's interesting because the doc listed stop kubelet in needing a specfici cloud but then also listed as generic. Removed it from the generic area. Do you think there is more that we should do to stop people from using certain node actions for cloud types that we have not set up or that are "generic" for now?

docs/node_scenarios.md
```diff
@@ -28,9 +28,9 @@ A google service account is required to give proper authentication to GCP for no
  
 After creating the service account you'll need to enable the account using the following: ```export GOOGLE_APPLICATION_CREDENTIALS="<serviceaccount.json>"```
 
-**NOTE**: The `stop_start_kubelet_scenario` and `node_crash_scenario` scenarios are supported as they are independent of the cloud platform.
+**NOTE**: The `node_crash_scenario` scenario is supported independent of the cloud platform.
```
mffiedler, line 31: Do you think we need to flag the scenario as not-applicable to the generic cloud_provider in the code.  i.e. treat it like reboot, etc?   Sorry to be so nitpicky about this

paigerube14: No worries! I was thinking abut that as well just to maybe inform the user of their incorrect scenario set up.

paigerube14: Added a lot of if statements here to log what is happening. Is this over kill?

paigerube14:
I noticed that the tox spacing was incorrect for the merge in GCP. Updated in this pr. Is this ok or should I undo it?

docs/node_scenarios.md
paigerube14: Found that the doc for generic scenarios should've said "stop_kubelet_scenario" instead of stop/start. Updated this

PR #54, opened 2020-11-17
Replace kubectl exec with kubernetes client python exec

paigerube14:
This is for: https://github.com/openshift-scale/kraken/issues/45

This utilizes the k8s python client instead of invoking a kubectl command every time. 

I also increased the sleep time between retries to test if the time was reset from 5 to 10 seconds

### DIFF
--- a/kraken/kubernetes/client.py
+++ b/kraken/kubernetes/client.py
@@ -1,4 +1,5 @@
 from kubernetes import client, config
+from kubernetes.stream import stream
 from kubernetes.client.rest import ApiException
 import logging
 import kraken.invoke.command as runcommand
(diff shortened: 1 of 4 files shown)


kraken/kubernetes/client.py
```diff
@@ -1,4 +1,5 @@
 from kubernetes import client, config
+from kubernetes.stream import stream
```
mffiedler, line 2: Add kubernetes and oauth2client to requirements.txt.  Interesting no one caught that before.

mffiedler (changes requested):
Verified time scenarios are working well using the client.

There are still a few usages of kubectl remaining, but none are in loops or performance sensitive spots.

kraken/kubernetes/client.py
paigerube14: It looks like they are installed from the powerfulseal installation. Should I also add to our requirements? They seem to be using a specific version of kubernetes (kubernetes==12.0.0a1)

mffiedler: On a RHEL 7.5 jump host I created a new virtualenv and did a pip install -r and they did not get installed.   Strange.  I had to install them explicitly.   I think it is OK if we add them to be save kubernetes>=12.0.0a1 and oauth2client>=whatever you have.   Mine is 4.1.3.

paigerube14: Added!

mffiedler (approved):
LGTM

PR #62, opened 2021-01-28
Add pod scenarios for custom app

juliemathew:
Signed-off-by: mjulie <mjulie@in.ibm.com>


config/config.yaml
```diff
@@ -8,6 +8,8 @@ kraken:
             - -    scenarios/etcd.yml
             - -    scenarios/regex_openshift_pod_kill.yml
               -    scenarios/post_action_regex.py
+            - -    scenarios/customapp_pod.yml
+              -    scenarios/post_action_customapp_pod.py
```
chaitanyaenr, line 12: @juliemathew we can definitely have the scenario part of the scenarios to consume in the config but we might want to delete the entry in the default config since the acme-air namespace might not exist on a default cluster. Thoughts?

juliemathew: @chaitanyaenr 
I agree to your suggestion. Removed the entry in the config file and modified the ReadMe for 'Pod Scenarios'.
Please review. Thanks.

paigerube14:
One small suggestion, you are kind of doing a double post action check. I would suggest picking just one. 

The 2 different types can be seen in the first pod_scenario section [here](https://github.com/cloud-bulldozer/kraken/blob/183313804d822b7a747db613b051ba68ba0524e0/config/config.yaml#L7) 

- the etcd scenario doesn't have a second item in the list and does the post action check in the yaml scenario [file](https://github.com/cloud-bulldozer/kraken/blob/183313804d822b7a747db613b051ba68ba0524e0/scenarios/etcd.yml#L21). 

- While the regex scenario has a python file in the list for that scenario. The regex scenario has varying number of pods so we use the python file to print the number of pods for that specific env. If you want to use a python file or bash script as the post action check, it will first run the script before the scenario to get the original number of pods before the pod killing and again after to make sure those numbers match. 
Also, if you have a varying number of pods you can print just the number of pods as seen [here](https://github.com/cloud-bulldozer/kraken/blob/183313804d822b7a747db613b051ba68ba0524e0/scenarios/post_action_regex.py#L68) and kraken will take care of the rest.

Please let me know if any of that doesn't make sense.

scenarios/customapp_pod.yaml
```diff
@@ -0,0 +1,32 @@
...
+        matches:
+          - labels:
+              namespace: "acme-air"
+              selector: ""
+        retries:
+          retriesTimeout:
+            timeout: 180
+
+        actions:
+          - checkPodCount:
```
chaitanyaenr, line 31: @juliemathew we can skip the "scenarios/post_action_customapp_pod.py" action in favor of this check I guess. Thoughts?

juliemathew: Yes. I agree. 
Made the changes. Please review. Thanks.

chaitanyaenr (approved):
LGTM, thanks @juliemathew.

docs/pod_scenarios.md
```diff
@@ -4,6 +4,7 @@ Kraken consumes [Powerfulseal](https://github.com/powerfulseal/powerfulseal) und
 
 #### Pod chaos scenarios
 Following are the components of Kubernetes/OpenShift for which a basic chaos scenario config exists today. Adding a new pod based scenario is as simple as adding a new config under scenarios directory and defining it in the config.
+For example, for adding a pod level scenario for a custom application, refer to the sample scenario (customapp_pod.yaml) provided in the scenarios directory.
```
paigerube14, line 7: Can we change this location to be under scenarios/customapp_pod.yaml or a link to the scenario in github like https://github.com/cloud-bulldozer/kraken/blob/master/scenarios/customapp_pod.yml

juliemathew: 👍

paigerube14:
LGTM

PR #72, opened 2021-02-08
Add support to deploy performance dashboards

chaitanyaenr:
This commit enables performance monitoring on the cluster when
running Kraken to be able to observe how cluster reacts to failures
as it's important to make sure the cluster is healthy in terms of
both recovery as well as performance.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -21,6 +21,8 @@ $ cd kraken
 $ pip3 install -r requirements.txt
 ```
 
+**NOTE**: Make sure python3-devel is installed on the system.
+
 #### Run
 ```
 $ python3 run_kraken.py --config <config_file_location>
--- a/README.md
+++ b/README.md
@@ -31,6 +31,9 @@ It's important to make sure to check if the targeted component recovered from th
 - Leveraging [Cerberus](https://github.com/openshift-scale/cerberus) to monitor the cluster under test and consuming the aggregated go/no-go signal to determine pass/fail. It is highly recommended to turn on the Cerberus health check feature avaliable in Kraken. Instructions on installing and setting up Cerberus can be found [here](https://github.com/openshift-scale/cerberus#installation). Once Cerberus is up and running, set cerberus_enabled to True and cerberus_url to the url where Cerberus publishes go/no-go signal in the Kraken config file.
 
 
+### Performance monitoring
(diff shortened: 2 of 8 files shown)


chaitanyaenr:
This is an initial step towards the automated goal of querying prometheus for performance metrics to determining pass/fail.

chaitanyaenr:
Fixes part of https://github.com/cloud-bulldozer/kraken/issues/7.

paigerube14:
Ravi, I am hitting an issue and I'm not sure if it's because of my set up or something wrong with the code. When the performance set up is running I keep hitting the following error

```
# export ./deploy.sh
-bash: export: `./deploy.sh': not a valid identifier
```

I tried this on both my mac and a fedora jump host. I did install a specific python3-devel package on my jump host and saw no difference. Any thoughts on a possible fix?

chaitanyaenr:
@paigerube14 I left an export hanging in there, my bad. Fixed it, do try it out when you get a chance. Thanks.

paigerube14:
LGTM

docs/installation.md
```diff
@@ -21,6 +21,8 @@ $ cd kraken
 $ pip3 install -r requirements.txt
 ```
 
+**NOTE**: Make sure python3-devel is installed on the system.
```
paigerube14, line 24: Might be nice to have a link on common ways to download python3-devel to different system types or how to know if it's installed

paigerube14:
This might be something to add to dittybopper and not here, but I hit an issue with my config file for kraken but dittybopper installed properly. When rerunning kraken, the dittybopper url is not printed out again because the dittybopper namespaces exists so had to scroll up to my previous run to find the URL. Not a huge problem to stop this PR just inconvenient in my case 

`Error from server (AlreadyExists): namespaces "dittybopper" already exists`

docs/installation.md
chaitanyaenr: Right, it depends on the linux distribution. Using the containerized version of the Kraken might be the way to go to avoid any issues with the dependencies.

chaitanyaenr:
@paigerube14 Nice catch, https://github.com/cloud-bulldozer/performance-dashboards/pull/11 should fix the issue on the dittybopper side. Thanks.

paigerube14:
LGTM

PR #78, opened 2021-03-18
Add pointer to the new blog

chaitanyaenr:
This commit:
- Adds a pointer to a new blog which emphasizes the importance
  of making chaos part of Perf/Scale test runs.
- Bumps up the allowed max-line-length for the linters.

### DIFF
--- a/run_kraken.py
+++ b/run_kraken.py
@@ -318,7 +318,7 @@ def main(cfg):
         iterations = config["tunings"].get("iterations", 1)
         daemon_mode = config["tunings"].get("daemon_mode", False)
         deploy_performance_dashboards = config["performance_monitoring"].get("deploy_dashboards", False)
-        dashboard_repo = config["performance_monitoring"].get("repo", "https://github.com/cloud-bulldozer/performance-dashboards.git")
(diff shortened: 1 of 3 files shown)


run_kraken.py
```diff
@@ -318,7 +318,7 @@ def main(cfg):
         iterations = config["tunings"].get("iterations", 1)
         daemon_mode = config["tunings"].get("daemon_mode", False)
         deploy_performance_dashboards = config["performance_monitoring"].get("deploy_dashboards", False)
-        dashboard_repo = config["performance_monitoring"].get("repo", "https://github.com/cloud-bulldozer/performance-dashboards.git")
+        dashboard_repo = config["performance_monitoring"].get("repo", "https://github.com/cloud-bulldozer/performance-dashboards.git") # noqa
```
paigerube14, line 321: did you mean to add this to this commit?

chaitanyaenr: Yes, it's causing the linters to fail since it exceeds the max allowed chars per line ( 134 vs 120 ) even after the change in tox configuration.

paigerube14:
LGTM

Issue #85, opened 2021-04-12
Trying to run openshift-apiserver scenario without success


kmlefebv:
I've set up a standalone cluster for Kraken and followed the steps provided in the Readme to get started. I made some changes in the Kraken config file to only include one scenario which would be the openshift-apiserver to isolate the problem I'm having. This is what I have in the kraken deployment pod log after starting the deployment:
```
2021-04-12 14:16:22,450 [INFO] Starting kraken
2021-04-12 14:16:22,455 [INFO] Initializing client to talk to the Kubernetes cluster
/usr/local/lib/python3.6/site-packages/urllib3/connectionpool.py:1020: InsecureRequestWarning: Unverified HTTPS request is being made to host 'api.kraken-test-04.cp.fyre.ibm.com'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  InsecureRequestWarning,
2021-04-12 14:16:23,455 [INFO] Fetching cluster info
2021-04-12 14:16:47,718 [INFO]
I0412 14:16:24.777729      13 request.go:668] Waited for 1.168597466s due to client-side throttling, not priority and fairness, request: GET:https://api.kraken-test-04.cp.fyre.ibm.com:6443/apis/policy/v1beta1?timeout=32s
I0412 14:16:34.778149      13 request.go:668] Waited for 11.168923785s due to client-side throttling, not priority and fairness, request: GET:https://api.kraken-test-04.cp.fyre.ibm.com:6443/apis/coordination.k8s.io/v1beta1?timeout=32s
NAME      VERSION   AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.6.16    True        False         21h     Cluster version is 4.6.16
I0412 14:16:36.903365      31 request.go:668] Waited for 1.155303512s due to client-side throttling, not priority and fairness, request: GET:https://api.kraken-test-04.cp.fyre.ibm.com:6443/apis/networking.k8s.io/v1?timeout=32s
I0412 14:16:47.103691      31 request.go:668] Waited for 11.353934532s due to client-side throttling, not priority and fairness, request: GET:https://api.kraken-test-04.cp.fyre.ibm.com:6443/apis/admissionregistration.k8s.io/v1beta1?timeout=32s
Kubernetes control plane is running at https://api.kraken-test-04.cp.fyre.ibm.com:6443

[...]

PR #86, opened 2021-04-13
Apiserver labels

paigerube14:
### Description
The label of the openshift apiserver pods has changed to have a "-a" at the end. 

### Fixes
No matching pods when trying to kill pods in the openshift apiserver namespace/scenario


mffiedler:
Verified and code LGTM.   Fixes #85

Issue #87, opened 2021-04-13
Dockerfile failing to build image


bmos299:
I trying to build an image to run in DEBUG mode but failing run I attempt to build in kraken/containers using the Dockerfile.  I am receiving a python error:
```
Collecting azure-keyvault-certificates~=4.1 (from azure-keyvault->-r requirements.txt (line 9))
  Downloading https://files.pythonhosted.org/packages/23/2b/e3219ae48391263b4657817b61eb5104ac1190d360e288e4546a18e61ca5/azure_keyvault_certificates-4.2.1-py2.py3-none-any.whl (235kB)
Collecting azure-keyvault-keys~=4.1 (from azure-keyvault->-r requirements.txt (line 9))
  Downloading https://files.pythonhosted.org/packages/a5/7f/a626704af6f0fb36bdc16397b49ab313a7e7a801ded5b43c6a74391f536a/azure_keyvault_keys-4.3.1-py2.py3-none-any.whl (248kB)
Collecting azure-keyvault-secrets~=4.1 (from azure-keyvault->-r requirements.txt (line 9))
  Downloading https://files.pythonhosted.org/packages/43/ae/4ad0c67e54f820d0ead249ce48e8cf498cfee42034dcd837d8f218ba9e76/azure_keyvault_secrets-4.2.0-py2.py3-none-any.whl (206kB)
Collecting azure-core<2.0.0,>=1.0.0 (from azure-identity->-r requirements.txt (line 10))
  Downloading https://files.pythonhosted.org/packages/19/18/21cfd7faf7ab24c35689c9f199179081cee8fec44668c7f090e1db61226d/azure_core-1.13.0-py2.py3-none-any.whl (133kB)
Collecting msal<2.0.0,>=1.6.0 (from azure-identity->-r requirements.txt (line 10))
  Downloading https://files.pythonhosted.org/packages/68/a9/b534f1158ffce8c551dea86d90981e9bd892f310c4c27d079d6b4b88849a/msal-1.11.0-py2.py3-none-any.whl (63kB)
Collecting msal-extensions~=0.3.0 (from azure-identity->-r requirements.txt (line 10))
  Downloading https://files.pythonhosted.org/packages/49/cb/c833ffa0f97c3098b146ac19bb2266c2d84b2119ffff83fdf001bb59d3ae/msal_extensions-0.3.0-py2.py3-none-any.whl
Collecting cryptography>=2.1.4 (from azure-identity->-r requirements.txt (line 10))
[...]

PR #88, opened 2021-04-13
Modifying the Dockerfile to add the needed pip3 update

bmos299:
### Description
Modifying the Dockerfile so images can be built
### Fixes
I just had to dd changes for the pip3 update

### DIFF
--- a/containers/Dockerfile
+++ b/containers/Dockerfile
@@ -14,6 +14,7 @@ COPY --from=origintests /usr/bin/kubectl /usr/bin/kubectl
 
 # Install dependencies
 RUN yum install -y git python36 python3-pip && \
+python3 -m pip install -U pip && \
 git clone https://github.com/openshift-scale/kraken /root/kraken && \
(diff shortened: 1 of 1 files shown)


containers/kraken.yml
```diff
@@ -18,7 +18,7 @@ spec:
         - name: kraken
           securityContext:
             privileged: true
-          image: quay.io/openshift-scale/kraken
+          image: bmos299/kraken-debug
```
chaitanyaenr, line 21: We might want to skip updating the image, an image build will be triggered once we merge this PR: https://quay.io/repository/openshift-scale/kraken.

chaitanyaenr:
Fixes #87

containers/kraken.yml
bmos299: absolutely...i didnt mean to commit this.

chaitanyaenr (approved):
LGTM

PR #97, opened 2021-05-05
Blank node name error message

paigerube14:
### Description
If the user leaves the node_name blank in the node scenario config file an error message is printed out saying it couldn't find the node. But in our default node scenario we are matching based on the label selector so this message can be confusing. 
This update should only print off the error message if a node name is given and does not match and ready node names


chaitanyaenr (approved):
LGTM

PR #102, opened 2021-06-15
Adding shut down scenario for gcp, az, aws, openstack

paigerube14:
### Description
Adding cluster shut down on the current cloud providers. Tried to add more useful logging and consistency of the functions across all the clouds we support

This is an updated version of [PR](https://github.com/cloud-bulldozer/kraken/pull/25) 

The one thing I didn't add is a retry count of how many retries we want to have for waiting for the vm's to be stopped or running. We have a timeout for one attempt at the vm getting to the specified status. Not sure if it's worth adding in a total retries counter to fail eventually if the vm's/ec instances never get to the wanted status. Not sure how often we will hit that but could happen. 

### Fixes
https://github.com/cloud-bulldozer/kraken/issues/63

### DIFF
--- a/scenarios/post_action_shut_down.py
+++ b/scenarios/post_action_shut_down.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python3
+import subprocess
+import logging
+import time
+import yaml
+
+
+def run(cmd):
+    out = ""
+    try:
+        output = subprocess.Popen(
+            cmd, shell=True, universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+        )
+        (out, err) = output.communicate()
+    except Exception as e:
+        logging.info("Failed to run %s, error: %s" % (cmd, e))
+    return out
+
+
+# Get cluster operators and return yaml
+def get_cluster_operators():
+    operators_status = run("kubectl get co -o yaml")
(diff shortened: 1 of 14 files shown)


paigerube14:
Thoughts on adding a post action here to validate all nodes are back and "Ready" after the extra wait period?

chaitanyaenr:
@paigerube14 Adding a post action to check if the nodes and API are back and serving the requests sounds good. Adding retries/timeout to wait for ~15 mins by default for the cluster/nodes to come back after restarting should be good enough I guess. Thoughts?

scenarios/post_action_shut_down.py
```diff
@@ -0,0 +1,31 @@
...
+        (out, err) = output.communicate()
+    except Exception as e:
+        logging.error("Failed to run %s, error: %s" % (cmd, e))
+    return out
+
+
+wait_duration = 10
+timeout = 900
+counter = 0
+not_ready = run("oc get nodes --no-headers | grep 'NotReady' | wc -l").rstrip()
```
chaitanyaenr, line 22: Maybe we should add checks for API reachability and cluster operators status to cover the case where cerberus is not deployed as part of the run? Thoughts?

paigerube14: Should be added now with doing `oc get co`

chaitanyaenr (approved):
LGTM, nice work @paigerube14.

PR #103, opened 2021-06-15
Add support to scrape and index metrics

### LINKED ISSUE #101: Scrape and store metrics long term
Kraken currently supports deploying mutable Grafana with pre-loaded dashboards to help with monitoring the cluster state as well as various performance metrics but the user needs to know the start and end time to narrow down the metrics for a specific run which makes it difficult especially in case of CI/automation runs where there won't be anyone monitoring the runs and the cluster will be terminated: https://github.com/cloud-bulldozer/kraken#performance-monitoring.

To solve the problem, the proposal is to add ability in Kraken to accept a profile which consists of metrics to scrape from the prometheus for the test run duration and store them in Elasticsearch, this way we can use Grafana to visualize them. Each test run can have a uuid associated with it to differentiate them.

Kube-burner indexing feature can be leveraged to achieve this: https://kube-burner.readthedocs.io/en/latest/indexers/.

chaitanyaenr:
### Description
This commit:
- Enables Kraken to leverage kube-burner to scrape metrics from
  Prometheus and index them into Elasticsearch. This way we can
  take a look at the metrics in Grafana long term even after the
  cluster is terminated.
- Enables separation of operations based on distribution with
  OpenShift as the default option. One of the use cases is to
  capture Prometheus instance details as it's installed by default
  while it's optional for Kubernetes.

### Fixes
#101


docs/metrics.md
```diff
@@ -0,0 +1,51 @@
+## Scraping and storing metrics for the run
+
+In addition to supporting tracking the metrics by deploying pre-loaded grafana dashboards on the live cluster, Kraken also supports capturing metrics for the duration of the scenarios defined in the config and indexes then into Elasticsearch to be able to store and evaluate the state of the runs long term. The indexed metrics can be visualized with the help of Grafana. It uses [Kube-burner](https://github.com/cloud-bulldozer/kube-burner) under the hood. The metrics to capture need to be defined in a metrics profile which Kraken consumes to query prometheus ( installed by default in OpenShift ) with the start and end timestamp of the run. Each run has a unique identifier ( uuid ) and all the metrics/documents in Elasticsearch will be associated with it. The uuid is generated automatially if not set in the config. This feature can be enabled in the [config](https://github.com/cloud-bulldozer/kraken/blob/master/config/config.yaml) by setting the following:
```
paigerube14, line 3: indexes the**m** into Elasticsearch (end of second line) 

Might just be me but the first part of this blob is a little hard to read, very wordy. Not sure how we can make it easier to read but still get all the information. Maybe split into 2 sentences?

chaitanyaenr: Sure thing.

chaitanyaenr: @paigerube14 I split the blob into two sentences like you suggested. PTAL.

paigerube14:
LGTM

paigerube14 (approved):
LGTM

PR #114, opened 2021-06-22
Pod docs

paigerube14:
### Description
Updated the pod killing scenario documentation to include specifics of how to edit/add a new scenario and what each section is looking for 

### Fixes
https://github.com/cloud-bulldozer/kraken/issues/112

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,13 +7,16 @@ Kraken injects deliberate failures into Kubernetes/OpenShift clusters to check i
 ![Kraken workflow](media/kraken-workflow.png)
 
 
-### Installation and usage
+### How to Get Started
 Instructions on how to setup, configure and run Kraken can be found at [Installation](docs/installation.md).
+
+See the [getting started doc](docs/getting_started.md) on support on how to get started with your own custom scenario or editing current scenarios for your specific usage
+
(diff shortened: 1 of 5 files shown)


chaitanyaenr (approved):
LGTM

paigerube14:
Anything else you can think to add here to clarify additions or edits to the configs?

chaitanyaenr:
@paigerube14 The content looks good, it will definitely help users understand how to tweak or create new scenario configs. Maybe mention that the scenario configs can be loaded by specifying them in the main config under https://github.com/cloud-bulldozer/kraken/blob/master/config/config.yaml#L8. Think it might also help if we add something like "Refer to the scenarios supported ( add link ) to get started" as part of the installation page - https://github.com/cloud-bulldozer/kraken/blob/master/docs/installation.md. Thoughts?

paigerube14:
Updated the doc to point to the kraken config line of how to add pod scenarios. I added the how to get started to the overall README, thoughts on having it there?

I also noticed that the pod_scenarios were running twice because the run_kraken.py file called the pod_scenarios run twice so I took that out. Can have a separate PR for that if needed

chaitanyaenr (approved):
LGTM

PR #115, opened 2021-06-23
Update the workflow

chaitanyaenr:
### Description
This commit modifies the workflow diagram to add pieces that are
leveraged to determine pass/fail for the chaos scenarios.


paigerube14:
Are you able to make the font's a little smaller just for the items in the bottom left so that they aren't as close together? 

The arrow directly to the right is a little wordy, what do you think of the below? 
"Injects deliberate failures through Kube and Cloud API's, then checks the recovery of the component under test"

chaitanyaenr:
@paigerube14 makes sense, will update the font and text.

paigerube14:
LGTM

paigerube14 (approved):
LGTM

PR #121, opened 2021-07-08
Increase the granularity of cerberus checks

chaitanyaenr:
This commit modifies the wait time from 60 seconds to 3 seconds between
each of the requests to the API to capture the components state at a more
granular level by default.


paigerube14 (approved):
LGTM

PR #126, opened 2021-07-14
Node mem and node io scenarios; Updating Litmus version

paigerube14:
### Description
Adding the node memory and node io stress scenario to litmus in the config
Updating litmus version to 1.13.6

### DIFF
--- a/scenarios/node_io_engine.yaml
+++ b/scenarios/node_io_engine.yaml
@@ -1,0 +1,35 @@
+apiVersion: litmuschaos.io/v1alpha1
+kind: ChaosEngine
+metadata:
+  name: nginx-chaos
+  namespace: default
+spec:
+  # It can be delete/retain
+  jobCleanUpPolicy: 'retain'
+  # It can be active/stop
+  engineState: 'active'
+  chaosServiceAccount: node-io-stress-sa
+  experiments:
+    - name: node-io-stress
+      spec:
+        components:
+          env:
+            # set chaos duration (in sec) as desired
+            - name: TOTAL_CHAOS_DURATION
+              value: '40'
+
+            ## specify the size as percentage of free space on the file system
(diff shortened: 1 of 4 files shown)


scenarios/node_io_engine.yaml
```diff
@@ -0,0 +1,35 @@
...
+            ## Number of core of CPU
+            - name: CPU
+              value: '1'
+
+            ## Total number of workers default value is 4
+            - name: NUMBER_OF_WORKERS
+              value: '3'
+
+             ## enter the comma separated target nodes name
+            - name: TARGET_NODES
```
chaitanyaenr, line 34: @paigerube14 The scenario is working perfectly without any issues when not targeting specific nodes but it seems to fail when trying to hog a particular node. We might also want a way to log the error. Thoughts?

chaitanyaenr: Sure thing.

scenarios/node_mem_engine.yaml
```diff
@@ -0,0 +1,28 @@
...
+        components:
+          env:
+            # set chaos duration (in sec) as desired
+            - name: TOTAL_CHAOS_DURATION
+              value: '30'
+
+            ## Specify the size as percent of total node capacity Ex: '30'
+            ## Note: For consuming memory in mebibytes change the variable to MEMORY_CONSUMPTION_MEBIBYTES
+            - name: MEMORY_CONSUMPTION_PERCENTAGE
+              value: '300'
```
chaitanyaenr, line 24: Maybe set this to 30% by default instead of 300?

chaitanyaenr (approved):
LGTM

scenarios/node_mem_engine.yaml
paigerube14: sure!

PR #127, opened 2021-07-19
Improve error handling

chaitanyaenr:
### Description
This commit:
- Adds timeout to avoid operations hanging for long durations.
- Improves exception handling and exits wherever needed.
- Sets KUBECONFIG env var globoally to access the cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,8 +20,8 @@ Kraken indexes the metrics specified in the profile into Elasticsearch in additi
 
 ```
 $ cd kraken
-$ podman-compose up or $ docker-compose up    # Spins up the containers specified in the docker-compose.yml file present in the run directory
-$ podman-compose down or $ docker-compose up  # Delete the containers installed
+$ podman-compose up or $ docker-compose up      # Spins up the containers specified in the docker-compose.yml file present in the run directory
+$ podman-compose down or $ docker-compose down  # Delete the containers installed
 ```
 This will manage the Cerberus and Elasticsearch containers on the host on which you are running Kraken.
 
--- a/kraken/invoke/command.py
+++ b/kraken/invoke/command.py
@@ -1,17 +1,16 @@
 import subprocess
 import logging
+import sys
 
 
 # Invokes a given command and returns the stdout
-def invoke(command):
+def invoke(command, timeout=None):
     try:
-        output = subprocess.Popen(
-            command, shell=True, universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
-        )
-        (out, err) = output.communicate()
+        output = subprocess.check_output(command, shell=True, universal_newlines=True, timeout=timeout)
     except Exception as e:
         logging.error("Failed to run %s, error: %s" % (command, e))
-    return out
+        sys.exit(1)
+    return output
 
 
 def run(command):
--- a/kraken/kubernetes/client.py
+++ b/kraken/kubernetes/client.py
@@ -13,8 +13,12 @@ kraken_node_name = ""
 # Load kubeconfig and initialize kubernetes python client
 def initialize_clients(kubeconfig_path):
     global cli
-    config.load_kube_config(kubeconfig_path)
-    cli = client.CoreV1Api()
+    try:
+        config.load_kube_config(kubeconfig_path)
+        cli = client.CoreV1Api()
+    except ApiException as e:
+        logging.error("Failed to initialize kubernetes client: %s\n" % e)
+        sys.exit(1)
 
 
 # List all namespaces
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -20,8 +20,8 @@ Kraken indexes the metrics specified in the profile into Elasticsearch in additi
 
 ```
 $ cd kraken
-$ podman-compose up or $ docker-compose up    # Spins up the containers specified in the docker-compose.yml file present in the run directory
-$ podman-compose down or $ docker-compose up  # Delete the containers installed
+$ podman-compose up or $ docker-compose up      # Spins up the containers specified in the docker-compose.yml file present in the run directory
+$ podman-compose down or $ docker-compose down  # Delete the containers installed
```
paigerube14, line 24: Good catch!

kraken/invoke/command.py
```diff
@@ -1,17 +1,16 @@
 import subprocess
 import logging
+import sys
 
 
 # Invokes a given command and returns the stdout
-def invoke(command):
+def invoke(command, timeout=None):
```
paigerube14, line 7: should we set this to 60 by default. Might be helpful for not having to find all invoke commands and it'll allow commands to not get stuck waiting for a response.

run_kraken.py
```diff
@@ -67,9 +67,9 @@ def main(cfg):
 
         # Cluster info
         logging.info("Fetching cluster info")
-        cluster_version = runcommand.invoke("kubectl get clusterversion")
+        cluster_version = runcommand.invoke("kubectl get clusterversion", 60)
```
paigerube14, line 70: I have hit an issue here before where I am running kraken on a different cluster than the one set using export KUBECONFIG... and it gives me the wrong clusterversion and cluster info
We should use --kubeconfig on all these commands to properly execute on the correct kubeconfig we are going to run kraken on

kraken/invoke/command.py
chaitanyaenr: We are using the invoke function for calling the pod scenarios i.e powerfulseal commands which have their own timeout set, so if we set the default to 60 here, it will kill those commands. Setting this to None will allow us to define timeouts explicitly for each of the commands depending on the use case. Thoughts?

run_kraken.py
chaitanyaenr: Nice catch @paigerube14, will fix it to explicitly use the config defined in the kraken config.

chaitanyaenr: We now have the KUBECONFIG env var set to make sure we are using the cluster specified in the kraken config. PTAL. Thanks.

kraken/invoke/command.py
paigerube14: Ah makes sense! Definitely don't want to add that for the powerfulseal calls

kraken/kubernetes/client.py
```diff
@@ -241,13 +245,17 @@ def find_kraken_node():
 
     if kraken_pod_name:
         # get kraken-deployment pod, find node name
-        runcommand.invoke("kubectl config set-context --current --namespace=" + str(kraken_project))
-        pod_json_str = runcommand.invoke("kubectl get pods/" + str(kraken_pod_name) + " -o json")
-        pod_json = json.loads(pod_json_str)
-        node_name = pod_json["spec"]["nodeName"]
+        try:
+            runcommand.invoke("kubectl config set-context --current --namespace=" + str(kraken_project))
```
paigerube14, line 249: Should also add kubeconfig here

paigerube14:
Also need to add kubeconfig param to https://github.com/cloud-bulldozer/kraken/blob/f051c1c30f8a24f1373a3614f3ac8b5d236fb000/kraken/pod_scenarios/setup.py#L92

kraken/kubernetes/client.py
chaitanyaenr: https://github.com/cloud-bulldozer/kraken/blob/346ce29b43dbffe19ef8e2eb4f2f280f3ecedf23/run_kraken.py#L64 will handle setting the kubeconfig context for all the commands running outside kubernetes python client

kraken/litmus/common_litmus.py
```diff
@@ -33,34 +33,43 @@ def run(scenarios_list, config, litmus_namespaces, litmus_uninstall, wait_durati
...
-                                    runcommand.invoke("kubectl delete -f %s" % l_item)
+                                    runcommand.invoke(
+                                        "kubectl delete -f %s --kubeconfig %s" % (l_item, kubeconfig_path)
+                                    )
+                            sys.exit(1)
             if litmus_uninstall:
                 for item in l_scenario:
                     logging.info("item " + str(item))
-                    runcommand.invoke("kubectl delete -f %s" % item)
+                    runcommand.invoke("kubectl delete -f %s --kubeconfig %s" % (item, kubeconfig_path))
```
paigerube14, line 47: think we can take these out now

paigerube14:
Tested with a cluster config located not at the default oc location, works great!

kraken/litmus/common_litmus.py
chaitanyaenr: Yup, fixed it. Thanks.

run_kraken.py
```diff
@@ -58,18 +58,22 @@ def main(cfg):
...
+        os.environ["KUBECONFIG"] = str(kubeconfig_path)
         kubecli.initialize_clients(kubeconfig_path)
 
         # find node kraken might be running on
         kubecli.find_kraken_node()
 
         # Cluster info
         logging.info("Fetching cluster info")
-        cluster_version = runcommand.invoke("kubectl get clusterversion")
+        cluster_version = runcommand.invoke("kubectl get clusterversion --kubeconfig %s" % kubeconfig_path, 60)
```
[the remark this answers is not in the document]
chaitanyaenr: Yup, fixed it. Thanks.

paigerube14 (approved):
LGTM

PR #129, opened 2021-07-21
Fix JSONDecodeError when running in OpenShift

koflerm:
In one of my issues (https://github.com/cloud-bulldozer/kraken/issues/128) I faced a problem where kraken was not able to run in OpenShift. Locally, with the same configuration files it works. The outcome was it is related to parsing the pod yaml of the kraken pod. With this code change, the broken parse is no longer needed and it saves also some lines of code because it uses kubectl directly to get the hostname.

### Fixes

https://github.com/cloud-bulldozer/kraken/issues/128

### DIFF
--- a/kraken/kubernetes/client.py
+++ b/kraken/kubernetes/client.py
@@ -3,7 +3,6 @@ from kubernetes.stream import stream
 from kubernetes.client.rest import ApiException
 import logging
 import kraken.invoke.command as runcommand
-import json
 import sys
 import re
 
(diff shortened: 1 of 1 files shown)


chaitanyaenr (approved):
LGTM

paigerube14:
I found [this issue](https://github.com/cloud-bulldozer/kraken/pull/127#discussion_r672324051) in kraken that I think could be effecting this issue. We are fixing in Ravi's PR https://github.com/cloud-bulldozer/kraken/pull/127

We should add an if statement to verify it is a proper node name is properly set and not set to `Error from server (NotFound): pods "*" not found`

chaitanyaenr:
> I found [this issue](https://github.com/cloud-bulldozer/kraken/pull/127#discussion_r672324051) in kraken that I think could be effecting this issue. We are fixing in Ravi's PR #127
> 
> We should add an if statement to verify it is a proper node name is properly set and not set to `Error from server (NotFound): pods "*" not found`. Just as an extra sanity check but this is definitely helpful

Right, PR #127 should help with setting the cluster context right and fails early on if not.

chaitanyaenr:
@koflerm Using jsonpath over loading and parsing the json is a great improvement. Looks like the checks are failing: https://github.com/cloud-bulldozer/kraken/pull/129/checks?check_run_id=3125295915, can we fix them when you get time please. We can run the pre-commit locally as well: $ pip3 install pre-commit $ pre-commit run --all-files. Do feel free to let us know in case of any questions. Thanks.

paigerube14 (approved):
LGTM, can you squash your commits

PR #130, opened 2021-07-23
Monitor user application availability during chaos

chaitanyaenr:
### Description
Current Kraken integration with Cerberus monitors the cluster as well
as the application health post chaos and pass/fails if they are not
healthy after chaos. This commit adds ability to monitor the user
application health during the chaos and fails the run in case of
downtime as it's potentially a downtime in case of customers environment
as well. It is especially useful in case of control plane failure
scenarios including API server, Etcd, Ingress etc.

### DIFF
--- a/run_kraken.py
+++ b/run_kraken.py
@@ -126,6 +126,7 @@ def main(cfg):
                             failed_post_scenarios = pod_scenarios.container_run(
                                 kubeconfig_path, scenarios_list, config, failed_post_scenarios, wait_duration
                             )
+
                         # Inject node chaos scenarios specified in the config
                         elif scenario_type == "node_scenarios":
                             logging.info("Running node scenarios")
@@ -135,6 +136,8 @@ def main(cfg):
                         elif scenario_type == "time_scenarios":
                             logging.info("Running time skew scenarios")
                             time_actions.run(scenarios_list, config, wait_duration)
+
+                        # Inject litmus based chaos scenarios
                         elif scenario_type == "litmus_scenarios":
                             logging.info("Running litmus scenarios")
                             if not litmus_installed:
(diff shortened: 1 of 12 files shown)


run_kraken.py
```diff
@@ -193,6 +197,18 @@ def main(cfg):
...
+        if check_applicaton_routes and cerberus_enabled:
+            status, unavailable_routes = cerberus.application_status(cerberus_url, start_time, end_time)
+            if not status:
+                logging.error(
+                    "Application routes: %s monitored by cerberus encountered downtime during the run, failing"
+                    % unavailable_routes
+                )
+                sys.exit(1)
+            else:
+                logging.info("Application routes being monited didn't encounter any downtime during the run!")
```
paigerube14, line 210: should be **monitored**

paigerube14:
When running the _get_status_ function at the end of each of the scenarios it exits if the cerberus status is false. In this case the cerberus status could be false because we are checking for a failed route, so it never gets to the application_status function. I think we need to take out the sys.exit(1) [here](https://github.com/cloud-bulldozer/kraken/blob/d69f0438089c9b076e050f7d4ac4cb1127343b64/kraken/cerberus/setup.py#L23) so that we can continue to check the application url

run_kraken.py
```diff
@@ -193,6 +197,18 @@ def main(cfg):
             logging.error("Post scenarios are still failing at the end of all iterations")
             sys.exit(1)
 
+        # Fail if the application routes monitered by cerberus experience downtime during the chaos
```
paigerube14, line 200: should be monitored

run_kraken.py
```diff
@@ -193,6 +197,18 @@ def main(cfg):
...
+        if check_applicaton_routes and cerberus_enabled:
```
paigerube14, line 201: thoughts on adding the call to this check into the get_status function so it gets called every time in the cerberus checks

chaitanyaenr: Nice idea, added the call to the cerberus get_status function. This will help with failing the run at the scenario level instead of the run level which can have multiple scenarios like you mentioned @paigerube14. PTAL.

paigerube14:
When setting cerberus to true and monitor routes to false I get the following error: 
`UnboundLocalError: local variable 'application_routes_status' referenced before assignment`

chaitanyaenr:
> When setting cerberus to true and monitor routes to false I get the following error:
> `UnboundLocalError: local variable 'application_routes_status' referenced before assignment`

Updated the PR to fix it. PTAL. Thanks.

paigerube14 (approved):
LGTM

PR #133, opened 2021-07-26
Avoid namespace context switch

### LINKED ISSUE #132: Kubectl set-context fails because .kube/config file is read-only
I am having a problem with starting kraken in our OpenShift 4.6 cluster:

error: open /root/.kube/config.lock: read-only file system
2021-07-26 13:58:48,953 [ERROR] Failed to run kubectl config set-context --current --namespace=openshift-kraken, error: Command 'kubectl config set-context --current --namespace=openshift-kraken' returned non-zero exit status 1.

The problem is according to your guide the .kube/config file is mounted via configmap to the correct folder. If you now execute a context switch in the code, kubectl tries to write this to the .kube/config file which is not possible as configmaps can only be mounted as read-only into containers. Do you have a solution for that?

chaitanyaenr:
### Description
There are cases where the kubeconfig can be read only like when running
Kraken as a kubernetes deployment. This commit fixes the instances to
use -n flag instead of a namespace context switch.

### Fixes
#132

### DIFF
--- a/kraken/kubernetes/client.py
+++ b/kraken/kubernetes/client.py
@@ -245,13 +245,13 @@ def find_kraken_node():
     if kraken_pod_name:
         # get kraken-deployment pod, find node name
         try:
-            runcommand.invoke("kubectl config set-context --current --namespace=" + str(kraken_project))
             node_name = runcommand.invoke(
-                "kubectl get pods/" + str(kraken_pod_name) + ' -o jsonpath="{.spec.nodeName}"'
+                "kubectl get pods/"
+                + str(kraken_pod_name)
+                + ' -o jsonpath="{.spec.nodeName}"'
+                + " -n"
+                + str(kraken_project)
             )
-
-            # Reset to the default project
-            runcommand.invoke("kubectl config set-context --current --namespace=default")
 
             global kraken_node_name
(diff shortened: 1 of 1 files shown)


chaitanyaenr:
@koflerm  PTAL.

kraken/kubernetes/client.py
```diff
@@ -245,14 +245,14 @@ def find_kraken_node():
     if kraken_pod_name:
         # get kraken-deployment pod, find node name
         try:
-            runcommand.invoke("kubectl config set-context --current --namespace=" + str(kraken_project))
             node_name = runcommand.invoke(
-                "kubectl get pods/" + str(kraken_pod_name) + ' -o jsonpath="{.spec.nodeName}"'
+                "kubectl get pods/"
+                + str(kraken_pod_name)
+                + ' -o jsonpath="{.spec.nodeName}"'
+                + "-n"
```
paigerube14: think you need a space at least before the - here

paigerube14 (approved):
LGTM

kraken/kubernetes/client.py
paigerube14: Ravi fixed this already

PR #134, opened 2021-07-28
Add more tunables for cpu hog scenario

chaitanyaenr:
### Description
This commit exposes the flags to tweak the number of cores and node
count to hog during the node-cpu-hog scenario.


paigerube14:
LGTM

PR #136, opened 2021-08-06
Run tasks in pod using Job object type

chaitanyaenr:
### Description
This commit switches the object type from Deployment to Job to be able
to display the status after executing all the scenarios specified in
the Kraken config instead of crashing which is expected in Deployments.

### Fixes
Fixes https://github.com/cloud-bulldozer/kraken/issues/135


paigerube14 (approved):
LGTM

PR #140, opened 2021-08-20
Container post action check

paigerube14:
### Description
Adding in check for container being ready in the pod once it gets killed

### DIFF
--- a/docs/container_scenarios.md
+++ b/docs/container_scenarios.md
@@ -14,4 +14,25 @@ scenarios:
   container_name: "<specific container name>"  # This is optional, can take out and will kill all containers in all pods found under namespace and label
   pod_names:  # This is optional, can take out and will select all pods with given namespace and label
   - <pod_name>
+  retry_wait: <number of seconds to wait for container to be running again> (defaults to 120seconds)
+```
+
+#### Post Action
+In all scenarios we do a post chaos check to wait and verify the specific component
+
+Here there are two options:
+1. Pass a custom script in the main config scenario list, that will run before the chaos and verify the output matches post chaos scenario
+
+See [scenarios/post_action_etcd_container.py](https://github.com/cloud-bulldozer/kraken/tree/master/scenarios/post_action_etcd_container.py) for an example
+```
+-   container_scenarios:                                 # List of chaos pod scenarios to load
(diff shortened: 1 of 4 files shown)


kraken/pod_scenarios/setup.py
```diff
@@ -59,18 +59,13 @@ def container_run(kubeconfig_path, scenarios_list, config, failed_post_scenarios
...
                 # capture start time
                 start_time = int(time.time())
-                container_killing_in_pod(cont_scenario)
+                killed_containers = container_killing_in_pod(cont_scenario)
                 logging.info("Waiting for the specified duration: %s" % (wait_duration))
                 time.sleep(wait_duration)
-                failed_post_scenarios = post_actions.check_recovery(
-                    kubeconfig_path, container_scenario_config, failed_post_scenarios, pre_action_output
-                )
+                failed_post_scenarios = check_failed_containers(killed_containers, cont_scenario.get("retry_wait", 120))
```
chaitanyaenr, line 67: Maybe move this check to line 65? This way the scenario waits for the container to recover and then waits for the usual time specified in the config. Thoughts?

paigerube14: In pod scenarios we do the wait duration first and then do the check post actions. I could see it being helpful both ways. I could change either to match. definitely think we should be consistent

docs/container_scenarios.md
```diff
@@ -14,4 +14,25 @@ scenarios:
...
+-   container_scenarios:                                 # List of chaos pod scenarios to load
+            - -    scenarios/container_etcd.yml
+              -    scenarios/post_action_etcd_container.py
+```
+
+2. Allow kraken to wait and check the killed containers become ready again. Kraken keeps a list of the specific
+containers that were killed as well as the namespaces and pods to verify all containers that were affected recover properly
+
+```
+retry_wait: <seconds to wait for container to recover
```
chaitanyaenr, line 37: Missing > in the end.

chaitanyaenr (approved):
Minor nit, other than that LGTM.

docs/container_scenarios.md
paigerube14: Added!

PR #148, opened 2021-09-03
Run all the litmus resources in a single namespace

chaitanyaenr:
### Description
    - This eases the usage and debuggability by running the fault injection pods in
      the same namespace as other resources of litmus. This will also ease the
      deletion process and ensure that there are no leftover objects on the cluster.
    
    - This commit also enables users to use the same rbac template for all the litmus
      scenarios without having to pull in a specific one for each of the scenarios.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -4,6 +4,7 @@ kraken:
     exit_on_failure: False                                 # Exit when a post action scenario fails
     litmus_version: v1.13.6                                # Litmus version to install
     litmus_uninstall: False                                # If you want to uninstall litmus if failure
+    litmus_namespace: litmus                               # Namespace to configure and run litmus based scenarios
     chaos_scenarios:                                       # List of policies/chaos scenarios to load
         -   container_scenarios:                                 # List of chaos pod scenarios to load
             - -    scenarios/container_etcd.yml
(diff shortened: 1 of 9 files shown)


paigerube14:
Would be nice to have a default cpu cores and nodes affected percentage in [node_cpu_hog_engine](https://github.com/chaitanyaenr/kraken-1/blob/litmus-rbac/scenarios/node_cpu_hog_engine.yaml) similar to [node_mem_engine](https://github.com/chaitanyaenr/kraken-1/blob/litmus-rbac/scenarios/node_mem_engine.yaml)

As well as a place holder for the node name 

ex) '<node_name>'

paigerube14 (commented):
Need to change service account to point to litmus-sa in both node_mem_engine.yaml and node_cpu_hog_engine.yaml

It is correct in node_io_engine.yaml

Once I edited this, it runs perfectly!

config/config.yaml
```diff
@@ -19,11 +20,11 @@ kraken:
         -   time_scenarios:                                # List of chaos time scenarios to load
             - scenarios/time_scenarios_example.yml
         -   litmus_scenarios:                              # List of litmus scenarios to load
-            - - https://hub.litmuschaos.io/api/chaos/1.13.6?file=charts/generic/node-cpu-hog/rbac.yaml
+            - - https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/templates/litmus-rbac.yaml
```
paigerube14, line 23: Would be nice to point to the local copy of that file: scenarios/templates/litmus-rbac.yaml

chaitanyaenr: Agreed, should we skip this change in favor of #149? Thoughts?

paigerube14: Sure! That works!

paigerube14 (approved):
LGTM

PR #150, opened 2021-09-13
Delete litmus crd's during the cleanup

chaitanyaenr:
This commit will ensure that the litmus resources installed on the
cluster get cleaned up and also creates the chaosengine in the
specified namespace.

### DIFF
--- a/kraken/litmus/common_litmus.py
+++ b/kraken/litmus/common_litmus.py
@@ -14,7 +14,7 @@ def run(scenarios_list, config, litmus_uninstall, wait_duration, litmus_namespac
         start_time = int(time.time())
         try:
             for item in l_scenario:
-                runcommand.invoke("kubectl apply -f %s" % item)
+                runcommand.invoke("kubectl apply -f %s -n %s" % (item, litmus_namespace))
                 if "http" in item:
                     f = requests.get(item)
                     yaml_item = list(yaml.safe_load_all(f.content))[0]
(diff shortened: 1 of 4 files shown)


paigerube14:
2 thoughts in general with all the litmus scenarios: 

* If the namespace that you set in the config doesn't match the namespace specified in the engine file, an error gets printed. I think this could be really problematic for the user copying over there own litmus scenarios and not properly setting the namespace in the engine file and config. Not 100% sure how to handle this situation though. **Thoughts**? 

`error: the namespace from the provided object "litmus-test" does not match the namespace "litmus". You must pass '--namespace=litmus-test' to perform this operation.`


* Should we put in an error if we read in <node_name> in any of the engine yaml files to remind the user to fill them in before running.

paigerube14 (approved):
This PR itself looks good to me!

chaitanyaenr:
> 2 thoughts in general with all the litmus scenarios:
> 
>     * If the namespace that you set in the config doesn't match the namespace specified in the engine file, an error gets printed. I think this could be really problematic for the user copying over there own litmus scenarios and not properly setting the namespace in the engine file and config. Not 100% sure how to handle this situation though. **Thoughts**?
> 
> 
> `error: the namespace from the provided object "litmus-test" does not match the namespace "litmus". You must pass '--namespace=litmus-test' to perform this operation.`
> 
>     * Should we put in an error if we read in <node_name> in any of the engine yaml files to remind the user to fill them in before running.

You are right @paigerube14, thinking more about this it might be better to not have the ability to configure the namespace given that we would need to create the namespace and manage it if it doesn't exist and ask the user to switch the namespace in the chaosengine to litmus in case of bringing their own configs like you mentioned. Latest commit should have the changes, please take a look and let me know if you think this helps address it. Thanks.

kraken/litmus/common_litmus.py
```diff
@@ -25,6 +25,13 @@ def run(scenarios_list, config, litmus_uninstall, wait_duration, litmus_namespac
                 if yaml_item["kind"] == "ChaosEngine":
                     engine_name = yaml_item["metadata"]["name"]
                     experiment_names = yaml_item["spec"]["experiments"]
+                    experiment_namespace = yaml_item["metadata"]["namespace"]
+                    if experiment_namespace != "litmus":
+                        logging.error(
+                            "Specified namespace in the scenario: %s is not supported, please switch it to litmus"
```
paigerube14, line 31: Can you add what file the incorrect namespace is in, might be easier to find that way

chaitanyaenr: Done. PTAL.

paigerube14 (approved):
LGTM

PR #151, opened 2021-09-21
Adding specific tag version of powerfulseal

paigerube14:
### Description
Adding specific version of powerfulseal to install


chaitanyaenr (approved):
LGTM

PR #152, opened 2021-09-21
Time spacing fix

paigerube14:
### Description
This fix should match on one or more spaces when checking the time/date has returned to the correct time 

I was not able to get the response to show the error that we had seen before but it should work
Format of date when error occured (2 spaces between Sep and 1) 
`Wed Sep  1 13:27:07 UTC 2021`

### Fixes
https://github.com/cloud-bulldozer/kraken/issues/146


chaitanyaenr (approved):
LGTM

PR #153, opened 2021-09-21
Prioritize filtering on namespace to improve performance

chaitanyaenr:
This will avoid querying all namespaces for pods matching the label_selector
if defined as shown in the sample scenario config. This commit also prints a
pointer to the report generated at the end of the run.


paigerube14 (approved):
LGTM!

PR #155, opened 2021-09-23
Pull images from quay instead of docker

chaitanyaenr:
This is needed to avoid getting rate limited. Build for reference -
https://recovery.quay.io/repository/openshift-scale/kraken/build/0cccc967-cfef-43d0-98ca-e3eccb698045.


paigerube14 (approved):
LGTM

PR #156, opened 2021-09-29
Add support to block traffic to an application

chaitanyaenr:
### Description
This commit enables users to simulate a downtime of an application
by blocking the traffic for the specified duration to see how
it/other components communicating with it behave in case of downtime.

### DIFF
--- a/docs/application_outages.md
+++ b/docs/application_outages.md
@@ -1,0 +1,17 @@
+### Application outages
+Scenario to block the traffic ( Ingress/Egress ) of an application matching the labels for the specified duration of time to understand the behavior of the service/other services which depend it during the downtime. This helps with the planning the requirements accordingly be it improving the timeouts or tweaking the alerts etc.
+
+##### Sample scenario config
+```
+application_outage:                                  # Scenario to create an outage of an application by blocking traffic
+  duration: 600                                      # Duration in seconds after which the routes will be accessible
+  namespace: <namespace-with-application>            # Namespace to target - all application routes will go inaccessible if pod selector is empty
+  pod_selector: {app=foo}                            # Pods to target
+  block: [Ingress, Egress]                           # It can be Ingress or Egress or Ingress, Egress
+```
+
+##### Debugging steps in case of failures
+Kraken creates a network policy blocking the ingress/egress traffic to create an outage, in case of failures before reverting back the network policy, you can delete it manually by executing the following commands to stop the outage:
+```
+$ oc delete networkpolicy/kraken-deny -n <targeted-namespace>
+```
--- a/kraken/application_outage/actions.py
+++ b/kraken/application_outage/actions.py
@@ -1,0 +1,55 @@
+import yaml
+import logging
+import time
+import kraken.cerberus.setup as cerberus
+from jinja2 import Template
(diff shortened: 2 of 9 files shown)


docs/application_outages.md
```diff
@@ -0,0 +1,17 @@
+### Application outages
+Scenario to block the traffic ( Ingress/Egress ) of an application matching the labels for the specified duration of time to understand the behavior of the service/other services which depend it during the downtime. This helps with the planning the requirements accordingly be it improving the timeouts or tweaking the alerts etc.
+
+##### Sample scenario config
+```
```
paigerube14, line 5: Would be nice to have a sample here with filled in parameters. The pod_selector variable is a little confusing and in a different format than normal so unsure what to put.

paigerube14:
On first run I get the error:
```
the path "kraken-network-policy.yaml" does not exist
2021-09-29 20:30:43,909 [ERROR] Failed to run kubectl create -f kraken-network-policy.yaml -n openshift-authentication, error: Command 'kubectl create -f kraken-network-policy.yaml -n openshift-authentication' returned non-zero exit status 1.

```

docs/application_outages.md
chaitanyaenr: Right, updated the docs and sample configuration. Thanks.

chaitanyaenr:
> On first run I get the error:
> 
> ```
> the path "kraken-network-policy.yaml" does not exist
> 2021-09-29 20:30:43,909 [ERROR] Failed to run kubectl create -f kraken-network-policy.yaml -n openshift-authentication, error: Command 'kubectl create -f kraken-network-policy.yaml -n openshift-authentication' returned non-zero exit status 1.
> ```

Should be fixed now. PTAL. Thanks.

kraken/application_outage/actions.py
```diff
@@ -0,0 +1,55 @@
...
+# block the traffic for the specified duration
+def run(scenarios_list, config, wait_duration):
+    failed_post_scenarios = ""
+    for app_outage_config in scenarios_list:
+        if len(app_outage_config) > 1:
+            with open(app_outage_config, "r") as f:
+                app_outage_config_yaml = yaml.full_load(f)
+                scenario_config = app_outage_config_yaml["application_outage"]
+                namespace = scenario_config["namespace"]
+                pod_selector = scenario_config["pod_selector"]
```
paigerube14, line 19: Can we add in a default value here in case the user doesn't want to set a pod selector like '{}'

paigerube14 (commented):
LGTM other than default for pod_selector

kraken/application_outage/actions.py
chaitanyaenr: Good idea, updated it. Thanks.

paigerube14: Are we able to do it like the parameters in the main run_kraken file? This way if the pod_selector key isn't found in the file 
ex.) chaos_scenarios = config["kraken"].get("chaos_scenarios", [])

I'm getting the below when commenting out the pod_selector line in the scenario file

```
    pod_selector = scenario_config["pod_selector"]
KeyError: 'pod_selector'
```

paigerube14 (approved):
LGTM

PR #164, opened 2021-10-14
Adding fix for openstack node name issue

paigerube14:
### Description
When setting the node name for the openstack cloud type. It was not properly getting the node name from the server

### Fixes
https://github.com/cloud-bulldozer/kraken/issues/161


chaitanyaenr (approved):
LGTM

PR #166, opened 2021-10-14
Multi node name

paigerube14:
### Description
Adding in the option to set multiple node names separated by a comma. Will update kraken-hub with the same comment in the documentation

### Fixes
https://github.com/cloud-bulldozer/kraken/issues/163


chaitanyaenr (approved):
LGTM

PR #172, opened 2021-11-03
Fixing killed container no reference issue

paigerube14:
### Description
In correct reference to a variable in container scenarios


chaitanyaenr (approved):
LGTM

PR #181, opened 2021-11-16
Adding check for if namespace exists

paigerube14:
### Description
Adding in a single check to make sure the namespace exists before trying to read the status of it

### Fixes
https://github.com/cloud-bulldozer/kraken/issues/176


chaitanyaenr (approved):
LGTM

PR #182, opened 2021-11-17
Added new scenario to fill up a given volumen

agullon:
### Description
Scenario to fill up a given PersistenVolumeClaim by creating a temp file of a given size from a pod that has the volume mounted in the given path. The purpose of this scenario is to fill up a volume to understand faults cause by the application using this volume.

### DIFF
--- a/CI/scenarios/volume_scenario.yaml
+++ b/CI/scenarios/volume_scenario.yaml
@@ -1,0 +1,44 @@
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: kraken-test-pv
+  labels:
+    type: local
+spec:
+  storageClassName: manual
+  capacity:
+    storage: 2Gi
+  accessModes:
+    - ReadWriteOnce
+  hostPath:
+    path: "/mnt/data"
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: kraken-test-pvc
+spec:
+  storageClassName: manual
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: kraken-test-pod
+spec:
+  volumes:
+    - name: kraken-test-pv
+      persistentVolumeClaim:
+        claimName: kraken-test-pvc
+  containers:
+    - name: kraken-test-container
+      image: 'image-registry.openshift-image-registry.svc:5000/openshift/httpd:latest'
+      volumeMounts:
+        - mountPath: "/home/krake-dir/"
+          name: kraken-test-pv
+      securityContext:
+        privileged: true
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -39,6 +39,8 @@ kraken:
             - scenarios/zone_outage.yaml
         -   application_outages:
             - scenarios/app_outage.yaml
+        -   pvc_scenarios:
+            - scenarios/pvc_scenario.yaml
 
 cerberus:
     cerberus_enabled: False                                # Enable it when cerberus is previously installed
--- a/docs/pvc_scenario.md
+++ b/docs/pvc_scenario.md
@@ -1,9 +1,9 @@
 ### PVC scenario
-Scenario to fill up a given PersistenVolumeClaim by creating a temp file of a given size from a pod that has the volume mounted in the given path. The purpose of this scenario is to fill up a volume to understand faults cause by the application using this volume. 
+Scenario to fill up a given PersistenVolumeClaim by creating a temp file of a given size from a pod that has the volume mounted in the given path. The purpose of this scenario is to fill up a volume to understand faults cause by the application using this volume.
 
 ##### Sample scenario config
 ```
-pvc_scenario:                        # Scenario to stop all the nodes for specified duration and restart the nodes
+pvc_scenario:
   pod_name: pod_name                 # Name of the pod with the PVC linked to
   namespace: namespace_name          # Namespace where the pod is
   mount_path: /path/to/pvc           # Path to the PVC in the default container pod
--- a/docs/pvc_scenario.md
+++ b/docs/pvc_scenario.md
@@ -1,21 +1,23 @@
 ### PVC scenario
-Scenario to fill up a given PersistenVolumeClaim by creating a temp file of a given size from a pod that has the volume mounted in the given path. The purpose of this scenario is to fill up a volume to understand faults cause by the application using this volume.
+Scenario to fill up a given PersistenVolumeClaim by creating a temp file on the PVC from a pod associated with it. The purpose of this scenario is to fill up a volume to understand faults cause by the application using this volume.
 
 ##### Sample scenario config
 ```
 pvc_scenario:
-  pod_name: pod_name                 # Name of the pod with the PVC linked to
-  namespace: namespace_name          # Namespace where the pod is
-  mount_path: /path/to/pvc           # Path to the PVC in the default container pod
-  file_size: 1024                    # Size of the file to be created in the PVC (block size is 1K)
-  duration: 60                       # Duration in seconds for the fault
+  pvc_name: <pvc_name>          # Name of the target PVC
+  namespace: <namespace_name>   # Namespace where the PVC is
+  fill_percentage: 50           # Target percentage to fill up the cluster, value must be higher than current percentage, valid values are between 0 and 99
+  duration: 60                  # Duration in seconds for the fault
 ```
 
 ##### Steps
+ - Get the `pod_name` where the PVC is mounted
+ - Get the `mount_path` where the PVC is mounted in the pod
+ - Get the PVC capacity and current used capacity
+ - Calculate file size to fill the PVC to the target `fill_percentage`
  - Connect to the `pod_name`
- - Change directory to the `mount_path`
- - Create a temp file `kraken.tmp` with random on the `mount_path`:
-    - `dd bs=1024 count=$file_size </dev/urandom >kraken.tmp`
+ - Create a temp file `kraken.tmp` with random data on the `mount_path`:
+    - `dd bs=1024 count=$file_size </dev/urandom > /mount_path/kraken.tmp`
  - Wait for the `duration` time
  - Remove the temp file created:
     - `rm kraken.tmp`
--- a/docs/pvc_scenario.md
+++ b/docs/pvc_scenario.md
@@ -5,19 +5,22 @@ Scenario to fill up a given PersistenVolumeClaim by creating a temp file on the 
 ```
 pvc_scenario:
   pvc_name: <pvc_name>          # Name of the target PVC
+  pod_name: <pod_name>          # Name of the pod where the PVC is mounted, it will be ignored if the pvc_name is defined
   namespace: <namespace_name>   # Namespace where the PVC is
   fill_percentage: 50           # Target percentage to fill up the cluster, value must be higher than current percentage, valid values are between 0 and 99
   duration: 60                  # Duration in seconds for the fault
 ```
 
 ##### Steps
- - Get the `pod_name` where the PVC is mounted
- - Get the `mount_path` where the PVC is mounted in the pod
+ - Get the pod name where the PVC is mounted
+ - Get the volume name mounted in the container pod
+ - Get the container name where the PVC is mounted
+ - Get the mount path where the PVC is mounted in the pod
  - Get the PVC capacity and current used capacity
- - Calculate file size to fill the PVC to the target `fill_percentage`
- - Connect to the `pod_name`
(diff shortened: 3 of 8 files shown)


config/config.yaml
```diff
@@ -39,6 +39,8 @@ kraken:
             - scenarios/zone_outage.yaml
         -   application_outages:
             - scenarios/app_outage.yaml
+        -   pvc_scenarios:
```
paigerube14, line 42: Can you also add this to the config_performance.yaml file

docs/pvc_scenario.md
```diff
@@ -0,0 +1,21 @@
+### PVC scenario
+Scenario to fill up a given PersistenVolumeClaim by creating a temp file of a given size from a pod that has the volume mounted in the given path. The purpose of this scenario is to fill up a volume to understand faults cause by the application using this volume. 
+
+##### Sample scenario config
+```
+pvc_scenario:                        # Scenario to stop all the nodes for specified duration and restart the nodes
```
paigerube14, line 6: Comment on this line does not match the scenario

chaitanyaenr:
Hi @agullon, we might want to run the pre-commit using the instructions: https://github.com/cloud-bulldozer/kraken/blob/master/docs/contribute.md#fix-formatting to fix the CI failures - mostly cosmetic. We can run $ pre-commit run --all-files and git add, commit to auto fix them.

kraken/pvc/pvc_scenario.py
```diff
@@ -0,0 +1,57 @@
...
+# Reads the scenario config and creates a temp file to fill up the PVC
+def run(scenarios_list, config):
+    failed_post_scenarios = ""
+    for app_config in scenarios_list:
+        if len(app_config) > 1:
+            with open(app_config, "r") as f:
+                config_yaml = yaml.full_load(f)
+                scenario_config = config_yaml["pvc_scenario"]
+                pod_name = scenario_config.get("pod_name", "")
+                namespace = scenario_config.get("namespace", "")
```
chaitanyaenr, line 17: Maybe exit if namespace/podname is empty? Thoughts?

paigerube14:
This scenario requires a pod to already be created with maps to a PVC, correct? It might be nice to be able to nice to include the pod creation as part of this setup/scenario because the user might not have one set up. I think we should do all of the set up of each scenario within kraken and not require any set up before the scenario runs

Thoughts @chaitanyaenr @agullon

docs/pvc_scenario.md
```diff
@@ -0,0 +1,21 @@
...
+  pod_name: pod_name                 # Name of the pod with the PVC linked to
```
chaitanyaenr, line 7: Maybe leverage pod labels to grab the pod-name to ease the CI/automated use cases where we might not know the pod name especially when it's appended with a random id each time it gets created. Something along the lines of user providing the pod-label to query and the number of pods to pick for injecting the chaos. Thoughts?

chaitanyaenr:
> This scenario requires a pod to already be created with maps to a PVC, correct? It might be nice to be able to nice to include the pod creation as part of this setup/scenario because the user might not have one set up. I think we should do all of the set up of each scenario within kraken and not require any set up before the scenario runs
> 
> Thoughts @chaitanyaenr @agullon

@paigerube14 There might be cases where applications running will have PV's allocated in public clouds in which case users might have a use case to target that particular PV to understand how it impacts that particular application. For example, kafka running on AWS will have a PV allocated and @agullon is looking at filling up that particular PV to understand the behavior of Kafka I guess. 

Maybe we can add a note about only targeting the pods with existing PV's for this scenario? Thoughts?

agullon:
@paigerube14 @chaitanyaenr I agree with the idea that we shouldn't do any prior set up before running the scenario. My first idea was to pass the PVC name as a parameter, but I pass the name of the namespace and pod instead. The reason I did it this way is because the implementation is easier. I'll investigate if we can get the pod, namespace and mount path from the PVC name (maybe we need to create a temp pod to do that).

What do you think about this approach?

config/config.yaml
agullon: done

docs/pvc_scenario.md
agullon: done

kraken/pvc/pvc_scenario.py
agullon: good idea, it's done

chaitanyaenr:
> @paigerube14 @chaitanyaenr I agree with the idea that we shouldn't do any prior set up before running the scenario. My first idea was to pass the PVC name as a parameter, but I pass the name of the namespace and pod instead. The reason I did it this way is because the implementation is easier. I'll investigate if we can get the pod, namespace and mount path from the PVC name (maybe we need to create a temp pod to do that).
> 
> What do you think about this approach?



> @paigerube14 @chaitanyaenr I agree with the idea that we shouldn't do any prior set up before running the scenario. My first idea was to pass the PVC name as a parameter, but I pass the name of the namespace and pod instead. The reason I did it this way is because the implementation is easier. I'll investigate if we can get the pod, namespace and mount path from the PVC name (maybe we need to create a temp pod to do that).
> 
> What do you think about this approach?

Maybe we can get the pod using the label and namespace specified in the config and then rsh/exec to the pod like the current implementation to create the file under /tmp to use the space if the mount_path is not defined or under the mount_path if specified with an assumption that the pod label/namespace specified by the user has a PV attached to it. Thoughts?

agullon:
I changed the approach. Now, the `pod_name` and `mount_path` are retrieved from the `pvc_name` and the `namespace`. Also added a new input parameter `fill_percentage` to specify how much the PVC will be filled up during the test.

kraken/pvc/pvc_scenario.py
```diff
@@ -0,0 +1,139 @@
...
+                # Check input params
+                if len(pvc_name) == 0:
+                    logging.error("You must specify the pvc_name")
+                    sys.exit(1)
+                if len(namespace) == 0:
+                    logging.error("You must specify the namespace where the PVC is")
+                    sys.exit(1)
+
+                # Get pod name
+                command = "oc describe pvc %s -n %s | grep \"Used By:\" | grep -Eo '[^: ]*$'" % (
```
chaitanyaenr, line 35: Maybe s/oc/kubectl to be able to run the scenario on kubernetes environments as well? Thoughts?

chaitanyaenr:
@agullon hitting the following error when running the scenario:

2021-11-22 20:07:19,825 [INFO] Executing scenarios for iteration 0
2021-11-22 20:07:19,825 [INFO] connection set up
127.0.0.1 - - [22/Nov/2021 20:07:19] "GET / HTTP/1.1" 200 -
2021-11-22 20:07:19,827 [INFO] response RUN
2021-11-22 20:07:19,827 [INFO] Running PVC scenario
  File "run_kraken.py", line 288, in <module>
    main(options.cfg)
  File "run_kraken.py", line 218, in main
    pvc_scenario.run(scenarios_list, config)
  File "/root/pvc/kraken/kraken/pvc/pvc_scenario.py", line 55, in run
    volume_mounts_list_json = json.loads(volume_mounts_list)
  File "/usr/lib64/python3.6/json/__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python3.6/json/decoder.py", line 342, in decode
    raise JSONDecodeError("Extra data", s, end)
json.decoder.JSONDecodeError: Extra data: line 72 column 1 (char 1842)

kraken/pvc/pvc_scenario.py
agullon: Done. I change the `oc` commands to `kubectl` commands.

agullon:
> @agullon hitting the following error when running the scenario:
> 
> ```
> 2021-11-22 20:07:19,825 [INFO] Executing scenarios for iteration 0
> 2021-11-22 20:07:19,825 [INFO] connection set up
> 127.0.0.1 - - [22/Nov/2021 20:07:19] "GET / HTTP/1.1" 200 -
> 2021-11-22 20:07:19,827 [INFO] response RUN
> 2021-11-22 20:07:19,827 [INFO] Running PVC scenario
> Traceback (most recent call last):
>   File "run_kraken.py", line 288, in <module>
>     main(options.cfg)
>   File "run_kraken.py", line 218, in main
>     pvc_scenario.run(scenarios_list, config)
>   File "/root/pvc/kraken/kraken/pvc/pvc_scenario.py", line 55, in run
>     volume_mounts_list_json = json.loads(volume_mounts_list)
>   File "/usr/lib64/python3.6/json/__init__.py", line 354, in loads
>     return _default_decoder.decode(s)
>   File "/usr/lib64/python3.6/json/decoder.py", line 342, in decode
>     raise JSONDecodeError("Extra data", s, end)
> json.decoder.JSONDecodeError: Extra data: line 72 column 1 (char 1842)
> ```
> 
> Here's the scenario file:
> 
> ```
> pvc_scenario:
>   pvc_name: prometheus-k8s-db-prometheus-k8s-0
>   namespace: openshift-monitoring   # Namespace where the PVC is
>   fill_percentage: 50           # Target percentage to fill up the cluster, value must be higher than current percentage, valid values are between 0 and 99
>   duration: 60                  # Duration in seconds for the fault
> ```

Thanks @chaitanyaenr , I fix it. Now I look for the volume name to get the mount path. I also get the container name to secure the execution of the commands in the container's pods.

paigerube14:
Can we add in a sample yaml to create a pod and pvc for this scenario? will be useful to have when adding a CI test for this scenario

agullon:
> Can we add in a sample yaml to create a pod and pvc for this scenario? will be useful to have when adding a CI test for this scenario

Sure, I'll work on this.

agullon:
Hi @paigerube14 I added a yaml file to create a temp pv, pvc and pod to run the pvc scenario for CI, please, take a look at the commit https://github.com/cloud-bulldozer/kraken/pull/182/commits/db2fdfeb814bf8a46adb5d64336fa988a7cec080

paigerube14:
@agullon thanks so much! Will use that in my testing today. Sorry for the delay

agullon:
@chaitanyaenr @paigerube14 Added support to find the PVC from a given pod on commit https://github.com/cloud-bulldozer/kraken/pull/182/commits/533369cc9b0049a0606536dd63296116fabaf5e4

chaitanyaenr (approved):
LGTM. Great addition to the supported chaos scenarios @agullon!

kraken/pvc/pvc_scenario.py
```diff
@@ -0,0 +1,175 @@
...
+                command = 'kubectl get pods %s -n %s -o json | jq -r ".spec.volumes"' % (str(pod_name), str(namespace),)
+                logging.debug("Get mount path command:\n %s" % command)
+                volumes_list = runcommand.invoke(command, 60).rstrip()
+                volumes_list_json = json.loads(volumes_list)
+                for entry in volumes_list_json:
+                    if len(entry["persistentVolumeClaim"]["claimName"]) > 0:
+                        volume_name = entry["name"]
+                        pvc_name = entry["persistentVolumeClaim"]["claimName"] if (len(pvc_name) == 0) else pvc_name
+                        break
+                logging.info("Volumen name: %s" % volume_name)
```
paigerube14, line 74: small edit, take out the "n" at the end of volume

agullon: done

chaitanyaenr:
@agullon Thoughts on adding support for this scenario in Kraken-hub post merge - https://github.com/cloud-bulldozer/kraken-hub/blob/main/docs/test_your_changes.md#adding-a-new-scenario?

paigerube14 (approved):
Used the volume_sceanrio.yaml to create a pv, pvc, and pod to use in the scenario and it worked great! LGTM

CI/scenarios/volume_scenario.yaml
```diff
@@ -0,0 +1,44 @@
...
+spec:
+  volumes:
+    - name: kraken-test-pv
+      persistentVolumeClaim:
+        claimName: kraken-test-pvc
+  containers:
+    - name: kraken-test-container
+      image: 'image-registry.openshift-image-registry.svc:5000/openshift/httpd:latest'
+      volumeMounts:
+        - mountPath: "/home/krake-dir/"
```
paigerube14, line 41: sorry, one more thing. the directory is missing the end "n" in kraken

chaitanyaenr: Merged the PR when this comment was in flight, my bad.

paigerube14: No problem!

agullon:
> @agullon Thoughts on adding support for this scenario in Kraken-hub post merge - https://github.com/cloud-bulldozer/kraken-hub/blob/main/docs/test_your_changes.md#adding-a-new-scenario?

Yes, I will work on it during the next days.

PR #183, opened 2021-11-18
Time fix

paigerube14:
### Description
Selecting container if not specified in config for time scenarios

### Fixes
https://github.com/cloud-bulldozer/kraken/issues/177

### DIFF
--- a/kraken/time_actions/common_time_functions.py
+++ b/kraken/time_actions/common_time_functions.py
@@ -7,12 +7,13 @@ import re
 import sys
 import kraken.cerberus.setup as cerberus
 import yaml
+import random
 
 
-def pod_exec(pod_name, command, namespace):
+def pod_exec(pod_name, command, namespace, container_name):
     i = 0
     for i in range(5):
-        response = kubecli.exec_cmd_in_pod(command, pod_name, namespace)
+        response = kubecli.exec_cmd_in_pod(command, pod_name, namespace, container_name)
         if not response:
             time.sleep(2)
             continue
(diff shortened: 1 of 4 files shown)


kraken/time_actions/common_time_functions.py
```diff
@@ -29,6 +30,20 @@ def node_debug(node_name, command):
     return response
 
 
+def get_container_name(pod_name, namespace, container_name=""):
+    container_names = runcommand.invoke(
+        'oc get pods %s -n %s -o jsonpath="{.spec.containers[*].name}"' % (pod_name, namespace)
```
chaitanyaenr, line 35: Maybe use read_namespaced_pod function from the kubernetes python client library to get the list of containers in pod? Thoughts?

paigerube14: Should be updated now, PTAL

kraken/time_actions/common_time_functions.py
```diff
@@ -79,23 +94,45 @@ def skew_time(scenario):
...
+                        "Couldn't reset time on container %s in pod %s in namespace %s"
+                        % (selected_container_name, pod[0], pod[1])
+                    )
+                    sys.exit(1)
+                pod_names[pod_counter].append(selected_container_name)
             else:
-                pod_exec(pod, skew_command, scenario["namespace"])
+                selected_container_name = get_container_name(pod, scenario["namespace"], container_name)
+                pod_exec(pod, skew_command, scenario["namespace"], selected_container_name)
+                if pod_exec_response is False:
```
chaitanyaenr, line 112: Think we missed capturing the pod_exec_response from the pod_exec command.

paigerube14: Should be fixed now, PTAL

chaitanyaenr:
@paigerube14 should we add the container_name field to the sample scenario config - https://github.com/cloud-bulldozer/kraken/blob/master/scenarios/time_scenarios_example.yml? Thoughts?

chaitanyaenr (approved):
LGTM

PR #197, opened 2021-12-09
Time issue 2

paigerube14:
### Description
Reworked the re.search (now re.match) for the time scenario. This will get rid of all extra spaces within the string and if there are any new line characters at the end of the string 

The main issue was that for days that are a single digit (like today Dec 9) it would never match the search string

### Fixes
https://github.com/cloud-bulldozer/kraken/issues/193


chaitanyaenr (approved):
LGTM

PR #199, opened 2021-12-20
Adding check to be sure label or namespace is not none first

paigerube14:
### Description
Adding in check to be sure label and namespace are not _None_ before trying to strip()


paigerube14:
rerun

chaitanyaenr (approved):
LGTM

PR #207, opened 2022-03-02
Bump jinja2 version to use latest MarkupSafe

yogananth-subramanian:
### Description
By default a jinja2 version less then 3.0.1 is installed, which raises error "cannot import name 'soft_unicode' from 'markupsafe'".
Using jinja2 3.0.1 or greater avoids this issue.


chaitanyaenr (approved):
LGTM

agullon (approved):
lgtm

PR #209, opened 2022-03-14

Sau1506mya:

### DIFF
--- a/run_kraken.py
+++ b/run_kraken.py
@@ -44,16 +44,19 @@ def main(cfg):
 	path=o.path
 	res=os.path.split(path)
 	cfg=res[1]'''
-    f = requests.get(cfg)
-    cfg=f.text
+    '''f = requests.get(cfg)
+    cfg=f.text'''
 
     # Parse and read the config
-    if os.path.isfile(cfg):
-	#if urlparse(cfg):
+    #if os.path.isfile(cfg):
+    if urlparse(cfg):
 	    #o=urlparse(cfg)
 	    #path=o.path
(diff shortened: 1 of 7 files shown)


paigerube14:
Looks like you need to rebase your PR with updates that have gone in during your work


```
git fetch upstream
git rebase upstream/master
<fix any issues> 
git push origin <branch_name> --force
```

Also might be good to create a branch off of your master branch to work on and squash your commits, see directions [here](https://github.com/cloud-bulldozer/kraken/blob/master/docs/contribute.md#pull-request)

run_kraken.py
```diff
@@ -37,9 +39,19 @@ def main(cfg):
...
     # Parse and read the config
+    flag = 0
     if os.path.isfile(cfg):
+        flag = 1
         with open(cfg, "r") as f:
             config = yaml.full_load(f)
+   
+    if urlparse(cfg):
+        flag = 1
+        f = requests.get(cfg)
```
chaitanyaenr, line 50: we might want to name the variable to reflect what it represents. Thoughts?

run_kraken.py
```diff
@@ -37,9 +39,19 @@ def main(cfg):
     logging.info("Starting kraken")
 
     # Parse and read the config
+    flag = 0
     if os.path.isfile(cfg):
+        flag = 1
         with open(cfg, "r") as f:
             config = yaml.full_load(f)
+   
+    if urlparse(cfg):
```
chaitanyaenr, line 48: Maybe add comments on a high level as to what the logic does? Thoughts?

PR #211, opened 2022-03-15

Sau1506mya:
cc: @chaitanyaenr @paigerube14 @yogananth-subramanian

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -10,40 +10,63 @@ kraken:
     litmus_uninstall_before_run: True                      # If you want to uninstall litmus before a new run starts
     chaos_scenarios:                                       # List of policies/chaos scenarios to load
         -   container_scenarios:                                 # List of chaos pod scenarios to load
-            - -    scenarios/container_etcd.yml
+          # - -  scenarios/container_etcd.yml
+            - -    https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/container_etcd.yml
         -   pod_scenarios:
-            - -    scenarios/etcd.yml
-            - -    scenarios/regex_openshift_pod_kill.yml
-              -    scenarios/post_action_regex.py
+          # - -    scenarios/etcd.yml
+            - -    https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/etcd.yml
+          # - -    scenarios/regex_openshift_pod_kill.yml
+            - -    https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/regex_openshift_pod_kill.yml
+          #   -    scenarios/post_action_regex.py
+              -    https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/post_action_regex.py
         -   node_scenarios:                                # List of chaos node scenarios to load
-            -   scenarios/node_scenarios_example.yml
+          # -   scenarios/node_scenarios_example.yml
+            -   https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/node_scenarios_example.yml
         -   pod_scenarios:
-            - -    scenarios/openshift-apiserver.yml
-            - -    scenarios/openshift-kube-apiserver.yml
+          # - -    scenarios/openshift-apiserver.yml
+            - -    https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/openshift-apiserver.yml
+          # - -    scenarios/openshift-kube-apiserver.yml
+            - -    https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/openshift-kube-apiserver.yml
         -   time_scenarios:                                # List of chaos time scenarios to load
-            - scenarios/time_scenarios_example.yml
+          # - scenarios/time_scenarios_example.yml
+            - https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/time_scenarios_example.yml
         -   litmus_scenarios:                              # List of litmus scenarios to load
-            - - scenarios/templates/litmus-rbac.yaml
-              - scenarios/node_cpu_hog_engine.yaml
-            - - scenarios/templates/litmus-rbac.yaml
-              - scenarios/node_mem_engine.yaml
-            - - scenarios/templates/litmus-rbac.yaml
-              - scenarios/node_io_engine.yaml
+          # - - scenarios/templates/litmus-rbac.yaml
+            - - https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/templates/litmus-rbac.yaml
+          #   - scenarios/node_cpu_hog_engine.yaml
+              - https://raw.githubusercontent.com/cloud-bulldozer/kraken/blob/master/scenarios/node_cpu_hog_engine.yaml
+          # - - scenarios/templates/litmus-rbac.yaml
+            - - https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/templates/litmus-rbac.yaml
+          #   - scenarios/node_mem_engine.yaml
+              - https://raw.githubusercontent.com/cloud-bulldozer/kraken/blob/master/scenarios/node_mem_engine.yaml
+          # - - scenarios/templates/litmus-rbac.yaml
+            - - https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/templates/litmus-rbac.yaml
+          #   - scenarios/node_io_engine.yaml
+              - https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/node_io_engine.yaml
         -   cluster_shut_down_scenarios:
-            - - scenarios/cluster_shut_down_scenario.yml
-              - scenarios/post_action_shut_down.py
+          # - - scenarios/cluster_shut_down_scenario.yml
+            - - https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/cluster_shut_down_scenario.yml
+          #   - scenarios/post_action_shut_down.py
+              - https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/post_action_shut_down.py
         -   namespace_scenarios:
-             - - scenarios/regex_namespace.yaml
-             - - scenarios/ingress_namespace.yaml
-               - scenarios/post_action_namespace.py
+          #  - - scenarios/regex_namespace.yaml
+             - - https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/regex_namespace.yaml
+          #  - - scenarios/ingress_namespace.yaml
+             - - https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/ingress_namespace.yaml
+          #    - scenarios/post_action_namespace.py
+               - https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/post_action_namespace.py
         -   zone_outages:
-            - scenarios/zone_outage.yaml
+          # - scenarios/zone_outage.yaml
+            - https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/zone_outage.yaml
         -   application_outages:
-            - scenarios/app_outage.yaml
+          # - scenarios/app_outage.yaml
+            - https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/app_outage.yaml
         -   pvc_scenarios:
-            - scenarios/pvc_scenario.yaml
+          # - scenarios/pvc_scenario.yaml
+            - https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/pvc_scenario.yaml
         -   network_chaos:
-            - scenarios/network_chaos.yaml
+          # - scenarios/network_chaos.yaml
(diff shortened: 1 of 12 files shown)


chaitanyaenr:
rerun

run_kraken.py
```diff
@@ -37,9 +39,19 @@ def main(cfg):
...
     # Parse and read the config
+    flag = 0
     if os.path.isfile(cfg):
+        flag = 1
         with open(cfg, "r") as f:
             config = yaml.full_load(f)
+
+    if urlparse(cfg):
+        flag = 1
+        f = requests.get(cfg)
```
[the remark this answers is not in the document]
paigerube14: +1 on renaming variables to be more representative

run_kraken.py
```diff
@@ -37,9 +39,26 @@ def main(cfg):
...
         with open(cfg, "r") as f:
             config = yaml.full_load(f)
+
+    # For config as remote file
+    if urlparse(cfg):
+        # Set flag as 1 since config is remote file
+        flag = 1
+        f = requests.get(cfg)
+        texts = f.text
+        config = yaml.safe_load(texts)
```
paigerube14, line 58: Let's try to be consistent with the yaml load function we use here and on line 50. 

Probably also good to put a try except around both this line and line 50 in case the file or url they are passing isn't a yaml type

paigerube14:
@Sau1506mya can you please take a look at this error, I tried running a local config file and got the below error. We want to be sure that running passing a url of the config file and still using a local copy of the config file works properly 

python run_kraken.py
```

2022-03-31 12:10:53,936 [INFO] Starting kraken
  File "run_kraken.py", line 313, in <module>
    main(options.cfg)
  File "run_kraken.py", line 56, in main
    f = requests.get(cfg)
  File "/Users/prubenda/PycharmProjects/pplRepos/sauyma/kraken-1/venv3/lib/python3.8/site-packages/requests/api.py", line 75, in get
    return request('get', url, params=params, **kwargs)
  File "/Users/prubenda/PycharmProjects/pplRepos/sauyma/kraken-1/venv3/lib/python3.8/site-packages/requests/api.py", line 61, in request
    return session.request(method=method, url=url, **kwargs)
  File "/Users/prubenda/PycharmProjects/pplRepos/sauyma/kraken-1/venv3/lib/python3.8/site-packages/requests/sessions.py", line 515, in request
    prep = self.prepare_request(req)
  File "/Users/prubenda/PycharmProjects/pplRepos/sauyma/kraken-1/venv3/lib/python3.8/site-packages/requests/sessions.py", line 443, in prepare_request
    p.prepare(
  File "/Users/prubenda/PycharmProjects/pplRepos/sauyma/kraken-1/venv3/lib/python3.8/site-packages/requests/models.py", line 318, in prepare
    self.prepare_url(url, params)
  File "/Users/prubenda/PycharmProjects/pplRepos/sauyma/kraken-1/venv3/lib/python3.8/site-packages/requests/models.py", line 392, in prepare_url
    raise MissingSchema(error)
requests.exceptions.MissingSchema: Invalid URL 'config/config.yaml': No scheme supplied. Perhaps you meant http://config/config.yaml?
```

run_kraken.py
```diff
@@ -37,9 +39,28 @@ def main(cfg):
...
+    try:
+        # For config as remote file
+        if urlparse(cfg):
+            # Set flag as 1 since config is remote file
+            flag = 1
+            content = requests.get(cfg)
+            texts = content.text
+            config = yaml.safe_load(texts)
+
+    except:
```
yogananth-subramanian, line 54: As indicated by pre-commit check, bare 'except' should not be used.

paigerube14: also good to print off the exception you got, like this example [here](https://github.com/cloud-bulldozer/kraken/blob/adb465cab02f7b1ecc289e99f3c8d4ed773ca08a/kraken/node_actions/aws_node_scenarios.py#L26)

run_kraken.py
```diff
@@ -37,9 +39,28 @@ def main(cfg):
...
+        if urlparse(cfg):
+            # Set flag as 1 since config is remote file
+            flag = 1
+            content = requests.get(cfg)
+            texts = content.text
+            config = yaml.safe_load(texts)
+
+    except Exception:
+        # For config as local file
+        if os.path.isfile(cfg):
```
paigerube14, line 56: this if statement/code block should also be under the try, want to be sure that the config's we are passing (whether by url or a local config) is a yaml file layout

Will want to exit out of kraken if an exception is thrown here

run_kraken.py
```diff
@@ -37,9 +39,39 @@ def main(cfg):
...
-            config = yaml.full_load(f)
+    # Initializing flag variable as 0 to test config is a file or not
+    flag = 0
+
+    try:
+        # Checking config is yaml file or not
+        check_yaml_file = urlparse(cfg)
+        path = check_yaml_file.path
+
+        if path[-5:] == ".yaml" or path[-4:] == ".yml":
```
paigerube14, line 50: @yogananth-subramanian @chaitanyaenr what are your thoughts on this if statement? I asked Sauyma to add in a way to throw an exception if the config file could not be loaded as a yaml.

yogananth-subramanian: Agreed, we can just raise an exception and exit.

run_kraken.py
```diff
@@ -37,9 +39,39 @@ def main(cfg):
     logging.info("Starting kraken")
 
     # Parse and read the config
-    if os.path.isfile(cfg):
-        with open(cfg, "r") as f:
-            config = yaml.full_load(f)
+    # Initializing flag variable as 0 to test config is a file or not
+    flag = 0
```
yogananth-subramanian, line 43: flag might not be required, we can just exit after raising an exception.

run_kraken.py
```diff
@@ -37,9 +39,39 @@ def main(cfg):
...
+            # For config as remote file
+            else:
+                # Set flag as 1 since config is remote file
+                flag = 1
+                content = requests.get(cfg)
+                texts = content.text
+                config = yaml.full_load(texts)
+
+        else:
+            raise Exception
```
yogananth-subramanian, line 68: A message with reason for the exception would be informative.

run_kraken.py
```diff
@@ -37,9 +39,39 @@ def main(cfg):
...
+            if os.path.isfile(cfg):
+                # Set flag as 1 since config is local file
+                flag = 1
+                with open(cfg, "r") as f:
+                    config = yaml.full_load(f)
+
+            # For config as remote file
+            else:
+                # Set flag as 1 since config is remote file
+                flag = 1
```
yogananth-subramanian, line 62: if you want to use flag, it should be set after the yaml is loaded.

run_kraken.py
chaitanyaenr: +1 for raising an exception and exiting.

run_kraken.py
Sau1506mya: Actually flag was set to test config is a file or not. If it is, all yaml file is loaded in the config variable and it goes on scenarios part otherwise it goes into the else part and exit from there.

run_kraken.py
Sau1506mya: Reason will be thrown once the exception was raised.

run_kraken.py
Sau1506mya: I agree, but when it is already checked that the config file is of yaml type, then either the file is local or remote. Once flag is set as 1, after that yaml will be loaded.

run_kraken.py
```diff
@@ -37,9 +39,39 @@ def main(cfg):
...
+        # Checking config is yaml file or not
+        check_yaml_file = urlparse(cfg)
+        path = check_yaml_file.path
+
+        if path[-5:] == ".yaml" or path[-4:] == ".yml":
+
+            # For config as local file
+            if os.path.isfile(cfg):
+                # Set flag as 1 since config is local file
+                flag = 1
```
chaitanyaenr, line 55: Thoughts on using boolean ( True/False ) for evaluations and naming the var to something more representative like local_file = True?

run_kraken.py
paigerube14: do not think we need flag this if the code is exiting on an exception

run_kraken.py
```diff
@@ -37,9 +39,39 @@ def main(cfg):
...
+                with open(cfg, "r") as f:
+                    config = yaml.full_load(f)
+
+            # For config as remote file
+            else:
+                # Set flag as 1 since config is remote file
+                flag = 1
+                content = requests.get(cfg)
```
paigerube14, line 63: Want to validate that the get request is getting a valid response, should check that the status_code is 200 here

paigerube14:
@Sau1506mya it looks like you pushed your chaos virtualenv folder, please remove that. Be sure you're only doing "git add" of files you know you edited

run_kraken.py
```diff
@@ -37,9 +38,32 @@ def main(cfg):
...
+
+        # For config as remote file
+        else:
+            content = requests.get(cfg)
+            if content.status_code == 200:
+                texts = content.text
+                config = yaml.full_load(texts)
+                # Set found_config as True since config is remote file
+                found_config = True
+
```
paigerube14, line 60: think if you add an else statement here that throws an exception, all cases will be covered that you don't need to have the found_config variable

paigerube14:
@Sau1506mya when you get a chance can you rebase your PR with upstream/master, some of the code looks a little funky that I can't tell what the issue is. 

Reviewing the doc of issues, seems like there are a couple of scenarios where you need to edit the scenarios file to enter in some parameters to actually run the scenario on your cluster 

ex.) Network scenario need to add in [_<line_selector>_](https://github.com/chaos-kubox/krkn/blob/eeac4ff9d3532e9d2542d1fe7eecd8a410be2470/scenarios/network_chaos.yaml#L4) and _<interface_name>_

I think you also need to edit the code under [post_actions/actions.py](https://github.com/chaos-kubox/krkn/blob/master/kraken/post_actions/actions.py) to be able to properly run python scripts.  I think in a couple places (here and pod_scenarios passing to powerfulseal) you will need to copy down the file to a temporary location to be able to run them 

Ex.) So instead of this:  
`
python3 https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/post_action_shut_down.py
`
You would have something like this
`
python3 tmp/scenarios/post_action_shut_down.py
`

@chaitanyaenr Any other ideas on how to get around not having to copy down the files to be able to run python scripts?

chaitanyaenr:
> 

Maybe we can add a note linking an issue mentioning that post-actions are not supported in case of using remote configs and work on the issue outside this PR? Thoughts?

chaitanyaenr:
> @Sau1506mya when you get a chance can you rebase your PR with upstream/master, some of the code looks a little funky that I can't tell what the issue is.
> 
> Reviewing the doc of issues, seems like there are a couple of scenarios where you need to edit the scenarios file to enter in some parameters to actually run the scenario on your cluster
> 
> ex.) Network scenario need to add in [_<line_selector>_](https://github.com/chaos-kubox/krkn/blob/eeac4ff9d3532e9d2542d1fe7eecd8a410be2470/scenarios/network_chaos.yaml#L4) and _<interface_name>_
> 
> I think you also need to edit the code under [post_actions/actions.py](https://github.com/chaos-kubox/krkn/blob/master/kraken/post_actions/actions.py) to be able to properly run python scripts. I think in a couple places (here and pod_scenarios passing to powerfulseal) you will need to copy down the file to a temporary location to be able to run them
> 
> Ex.) So instead of this: `python3 https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/post_action_shut_down.py` You would have something like this `python3 tmp/scenarios/post_action_shut_down.py`
> 
> @chaitanyaenr Any other ideas on how to get around not having to copy down the files to be able to run python scripts?

Maybe we can add a note linking an issue mentioning that post-actions are not supported in case of using remote configs and work on the issue outside this PR? Thoughts?

paigerube14:
@chaitanyaenr think we need to also specify pod_scenarios can only be local because powerfulseal is expecting a local file path. Think we need to maybe make a table of what scenarios/parts work with url's from this PR and just make a plan on what to add in later

config/config.yaml
```diff
@@ -9,38 +9,38 @@ kraken:
     litmus_uninstall: False                                # If you want to uninstall litmus if failure
     chaos_scenarios:                                       # List of policies/chaos scenarios to load
         -   container_scenarios:                                 # List of chaos pod scenarios to load
-            - -    scenarios/container_etcd.yml
+            - -  scenarios/container_etcd.yml
         -   pod_scenarios:
-            - -    scenarios/etcd.yml
-            - -    scenarios/regex_openshift_pod_kill.yml
-              -    scenarios/post_action_regex.py
+            - -    https://raw.githubusercontent.com/cloud-bulldozer/kraken/master/scenarios/etcd.yml
```
paigerube14, line 14: Lets update these links to have the new organization and branch 

Ex.) https://raw.githubusercontent.com/chaos-kubox/krkn/main/scenarios/etcd.yml

paigerube14:
I had tested the following scenarios with this PR, is anyone able to run the rest of the sceanrio types so we can get this PR merged. Note, some of the scenario types I listed might not be possible with remote scenario options. 

- [x] Containers
- [ ] Pod scenario (NA)
- [x] Node scenario
- [x] Time scenario
- [ ] Litmus scenario (NA)
- [ ] Cluster shut down
- [x] Namespace scenario
- [x] Zone outage
- [ ] Application outage
- [ ] Pvc scenario
- [ ] Network chaos

chaitanyaenr (approved):
LGTM

chaitanyaenr:
Closing this PR in favor of #269. Thanks.

PR #212, opened 2022-03-18
Update Litmus version in config_performance.yaml

StevenBarre:
### Description
Update Litmus version to match `config.yaml`

### Fixes
v1.10 uses a different case for `Experimentstatus` than v1.13 and thus will always fail.


chaitanyaenr (approved):
LGTM. Thanks @StevenBarre!

PR #214, opened 2022-03-22
Set itsdangerous package to 2.0.1

yogananth-subramanian:
### Description
Since import of JSON is deprecated from itsdangerous version 2.1.0 [1], this patch fixes the version to 2.0.1.
By fixing to version, we can avoid the below import error.
Error "from itsdangerous import json as _json ImportError: cannot import name 'json' from 'itsdangerous'"

[1] https://itsdangerous.palletsprojects.com/en/2.1.x/changes/#version-2-1-0

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,3 +19,4 @@ python-ipmi
 podman-compose
 docker-compose
 jinja2>=3.0.1
+itsdangerous==2.0.1


paigerube14:
@yogananth-subramanian I'm a little confused about what you're talking about here. Where are we using _itsdangerous_? Did you find which requirement installs it

yogananth-subramanian:
> @yogananth-subramanian I'm a little confused about what you're talking about here. Where are we using _itsdangerous_? Did you find which requirement installs it. Think this needs to be upstream and not in our repostiory specifically
@paigerube14 Flask package seems to install itsdangerous package and Flask seem to be installed due to powerfulseal.
But when Powerful seal is installed using setup.py it is able to install a version of itsdangerous<2.0,>=0.24.
But pip with current kraken requirements installs a version of itsdangerous>2.0 that is 2.1.1, which has deprecated import of json.
Since pip is not able to find the appropriate version of  itsdangerous to install, hence have added this patch to do so.

paigerube14 (approved):
LGTM

PR #215, opened 2022-03-30
Alibaba cloud type and updating pre commit versioning  

paigerube14:
### Description
Adding alibaba specific node scenarios 

I also ran into issues with pre-commit versioning that were fixed with the below updates

### Fixes
https://github.com/cloud-bulldozer/kraken/issues/202

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,11 @@
 ---
 repos:
-  - repo: git://github.com/Lucas-C/pre-commit-hooks
+  - repo: https://github.com/Lucas-C/pre-commit-hooks
     rev: v1.1.1
     hooks:
       - id: remove-tabs
 
-  - repo: git://github.com/pre-commit/pre-commit-hooks
+  - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v2.0.0
     hooks:
       - id: trailing-whitespace
@@ -16,7 +16,7 @@ repos:
       - id: check-ast
 
   - repo: https://github.com/psf/black
-    rev: 19.10b0
+    rev: 22.3.0
     hooks:
       - id: black
         args: ['--line-length', '120']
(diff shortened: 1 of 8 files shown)


yogananth-subramanian (approved):
LGTM!
It errored out appropriately when I tried executing with fake alibaba credentials.

mffiedler (approved):
Tested on an alicloud cluster.   Cloud-specific scenarios all work well.

chaitanyaenr (approved):
Code looks good to me! I didn't test it due to lack of access to Alibaba cloud.

kraken/node_actions/abstract_node_scenarios.py
```diff
@@ -7,17 +7,22 @@
...
-    def node_stop_scenario(self, instance_kill_count, node, timeout):
+    def node_stop_scenario(self, instance_kill_count, node, timeout, force_action=False):
         pass
 
     # Node scenario to stop and then start the node
-    def node_stop_start_scenario(self, instance_kill_count, node, timeout):
+    def node_stop_start_scenario(self, instance_kill_count, node, timeout, force_action=False):
         logging.info("Starting node_stop_start_scenario injection")
-        self.node_stop_scenario(instance_kill_count, node, timeout)
+        self.node_stop_scenario(instance_kill_count, node, timeout, force_action)
```
paigerube14, line 25: @chaitanyaenr @mffiedler any ideas on how to rework this so I don't have to go through and add "force_action" as a parameter in all the stop scenario function definitons in each cloud? Want to add in the option for alibaba specifically and ignore for all other clouds

Works for alibaba but get the below error on AWS. 
`TypeError: node_stop_scenario() takes 4 positional arguments but 5 were given`

chaitanyaenr: I would go with not defining the force_action as parameter to keep things generic for all cloud providers and instead set force_action=true by default for Alibaba node actions as that might be more representative of chaos. Thoughts?

paigerube14: Set default force value to True

PR #217, opened 2022-04-05
Catching Unknown status of node when cluster becomes disconnected 

paigerube14:
### Description
This fix adds in exception catches when performing node reboot and the cluster gets disconnected. Also added in a timeout to the get_node_status function to help with this

### Fixes
https://github.com/cloud-bulldozer/kraken/issues/216

### DIFF
--- a/kraken/kubernetes/client.py
+++ b/kraken/kubernetes/client.py
@@ -282,16 +282,16 @@ def get_job_status(name, namespace="default"):
 
 
 # Obtain node status
-def get_node_status(node):
+def get_node_status(node, timeout=60):
     try:
-        node_info = cli.read_node_status(node, pretty=True)
+        node_info = cli.read_node_status(node, pretty=True, _request_timeout=timeout)
     except ApiException as e:
         logging.error(
             "Exception when calling \
                        CoreV1Api->read_node_status: %s\n"
             % e
         )
-        raise e
+        return None
     for condition in node_info.status.conditions:
         if condition.type == "Ready":
             return condition.status
--- a/kraken/node_actions/common_node_functions.py
+++ b/kraken/node_actions/common_node_functions.py
@@ -43,11 +43,17 @@ def wait_for_ready_status(node, timeout):
 # Wait till node status becomes NotReady
 def wait_for_unknown_status(node, timeout):
     for _ in range(timeout):
-        if kubecli.get_node_status(node) == "Unknown":
-            break
+        try:
+            node_status = kubecli.get_node_status(node, timeout)
+            if node_status is None or node_status == "Unknown":
+                break
+        except Exception:
+            logging.error("Encountered error while getting node status, waiting 3 seconds and retrying")
         time.sleep(3)
-    if kubecli.get_node_status(node) != "Unknown":
-        raise Exception("Node condition status isn't Unknown")
+    node_status = kubecli.get_node_status(node, timeout)
+    logging.info("node status " + str(node_status))
+    if node_status is not None and node_status != "Unknown":
+        raise Exception("Node condition status isn't Unknown after %s seconds" % str(timeout))
 
 
 # Get the ip of the cluster node


kraken/kubernetes/client.py
```diff
@@ -274,15 +274,16 @@ def get_job_status(name, namespace="default"):
 
 
 # Obtain node status
-def get_node_status(node):
+def get_node_status(node, timeout=60):
     try:
-        node_info = cli.read_node_status(node, pretty=True)
+        node_info = cli.read_node_status(node, pretty=True, _request_timeout=timeout)
```
chaitanyaenr, line 279: The parameter must be named as _request_timeout in Kubernetes python client. Just want to make sure that's the case. Thoughts?

paigerube14: Yes I think it is properly named 


```
 def read_node_status(self, name, **kwargs):  # noqa: E501
        """read_node_status  # noqa: E501

        read status of the specified Node  # noqa: E501
        This method makes a synchronous HTTP request by default. To make an
        asynchronous HTTP request, please pass async_req=True
        >>> thread = api.read_node_status(name, async_req=True)
        >>> result = thread.get()

        :param async_req bool: execute request asynchronously
        :param str name: name of the Node (required)
        :param str pretty: If 'true', then the output is pretty printed.
        :param _preload_content: if False, the urllib3.HTTPResponse object will
                                 be returned without reading/decoding response
                                 data. Default is True.
        :param _request_timeout: timeout setting for this request. If one
                                 number provided, it will be total request
                                 timeout. It can also be a pair (tuple) of
                                 (connection, read) timeouts.
        :return: V1Node
                 If the method is called asynchronously,
                 returns the request thread.
        """
```

chaitanyaenr (approved):
LGTM

sandrobonazzola:
"This fix adds in exception catches when performing node reboot and the cluster gets disconnected. Also added in a timeout to the get_node_status function to help with this"

The "Also added..." part suggests this could have been split in two separate PRs, making review easier.

sandrobonazzola (changes requested):
Comments inline

paigerube14:
> 

The also part I just had to add in to be able to properly get a status back if the cluster becomes disconnected. It was part of the update and I think should still be in this PR

kraken/node_actions/common_node_functions.py
```diff
@@ -32,17 +32,26 @@ def get_node(node_name, label_selector, instance_kill_count):
 
 # Wait till node status becomes Ready
 def wait_for_ready_status(node, timeout):
-    runcommand.invoke("kubectl wait --for=condition=Ready " "node/" + node + " --timeout=" + str(timeout) + "s")
+    runcommand.invoke("kubectl wait --for=condition=Ready node/" + node + " --timeout=" + str(timeout) + "s")
```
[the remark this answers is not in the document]
paigerube14: Had already created an issue to update this in the near future: https://github.com/cloud-bulldozer/kraken/issues/218

paigerube14:
> "This fix adds in exception catches when performing node reboot and the cluster gets disconnected. Also added in a timeout to the get_node_status function to help with this"
> 
> The "Also added..." part suggests this could have been split in two separate PRs, making review easier.

This is also part of the fix I implemented to be able to properly get a status back if the cluster becomes disconnected. It was part of the update and I think should still be in this PR

kraken/node_actions/common_node_functions.py
```diff
@@ -32,17 +32,26 @@ def get_node(node_name, label_selector, instance_kill_count):
...
+        except Exception:
+            logging.error("Encountered error while getting node status, waiting 3 seconds and retrying")
         time.sleep(3)
-    if kubecli.get_node_status(node) != "Unknown":
-        raise Exception("Node condition status isn't Unknown")
+    try:
+        node_status = kubecli.get_node_status(node, timeout)
+        logging.info("node status " + str(node_status))
+        if node_status is not None and node_status != "Unknown":
+            logging.error("Node condition status isn't Unknown after %s seconds" % str(timeout))
```
yogananth-subramanian, line 52: Should it be an Exception instead of just logging it as "isn't unknown" like before.
`raise Exception("Node condition status isn't Unknown")`

kraken/node_actions/common_node_functions.py
```diff
@@ -32,17 +32,26 @@ def get_node(node_name, label_selector, instance_kill_count):
...
+    except Exception:
+        logging.error("Encountered error while getting node status, for more than %s seconds" % str(timeout))
```
yogananth-subramanian, line 54: `ApiException` is already handled in get_node_status and any other exception that was raised by `wait_for_unknown_status` in the old was handled by the calling function, so is there a need to handle the exception here?

kraken/node_actions/common_node_functions.py
```diff
@@ -32,17 +32,23 @@ def get_node(node_name, label_selector, instance_kill_count):
...
+        try:
+            node_status = kubecli.get_node_status(node, timeout)
+            if node_status is None or node_status == "Unknown":
+                break
+        except Exception:
+            logging.error("Encountered error while getting node status, waiting 3 seconds and retrying")
+    time.sleep(3)
+    node_status = kubecli.get_node_status(node, timeout)
+    logging.info("node status " + str(node_status))
+    if node_status is not None and node_status != "Unknown":
```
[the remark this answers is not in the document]
yogananth-subramanian: /lgtm
@paigerube14  If you could just rebase the patch, it would be ready to be merged.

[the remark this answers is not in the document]
paigerube14: @yogananth-subramanian done!

PR #226, opened 2022-04-12
Add build container step to PRs, change README wording

sanjacodes:
### Description

Changes some wording/typos in the README and introduces Actions changes.

### Fixes
 
* Now runs container builds on PRs to verify basic steps
* Excludes certain Actions steps handling Quay uploads for PRs

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,26 +1,31 @@
 name: Docker Image CI
 on:
   push:
-    branches: ['master']
+    branches:
+      - master
+  pull_request:
 
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
     - name: Login in quay
+      if: github.ref == 'refs/heads/master' && github.event_name == 'push'
       run: docker login quay.io -u ${QUAY_USER} -p ${QUAY_TOKEN}
       env:
         QUAY_USER: ${{ secrets.QUAY_USER_1 }}
         QUAY_TOKEN: ${{ secrets.QUAY_TOKEN_1 }}
     - name: Check out code
-      uses: actions/checkout@master
+      uses: actions/checkout@v3
     - name: Build the Docker images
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-image.yml
```diff
@@ -1,26 +1,31 @@
...
+      - master
+  pull_request:
 
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
     - name: Login in quay
+      if: github.ref == 'refs/heads/master' && github.event_name == 'push'
       run: docker login quay.io -u ${QUAY_USER} -p ${QUAY_TOKEN}
```
paigerube14, line 14: I'm not 100% sure on this but if there's a PR from a fork I do not think this will work with the user and token coming from secrets. We hit a lot of restrictions on what we could and couldn't do in github actions because our process includes forking

paigerube14: Might need to create a PR from a fork as well to verify this works both ways. Again not sure this would work and pick up the updates made from this PR. Might have to merge and then retest on fork

sanjacodes: Yes, you're right - in this case and related steps, I've added an if condition that stops this step from running from a fork - it will only run with push events into the main branch, such as when a PR is merged.

This only enables us to get the other basic checks we need for PRs so that we don't have to run these steps manually.

.github/workflows/docker-image.yml
```diff
@@ -1,26 +1,31 @@
 name: Docker Image CI
 on:
   push:
-    branches: ['master']
+    branches:
+      - master
```
paigerube14, line 5: do we want to take off the branches here so that we can run this on any updates that are in PR's? or is that handled by the pull_request section?

sanjacodes: Yes, that's handled by the pull_request section. 👍🏻

chaitanyaenr (approved):
LGTM

PR #236, opened 2022-04-28
Remove contributor-unfriendly linting rules

sanjacodes:
We can introduce a workflow that removes trailing whitespaces and missing newlines rather than going through additional reviews and fixes with contributions


chaitanyaenr (approved):
LGTM

PR #240, opened 2022-05-02
Rework node ready wait status

obrienrobert:


sanjacodes:
Thanks for your PR!

We're right now trying to rework our current code base to allow for more scalability and security. This is one such instance, where we would need to use [watches](https://github.com/kubernetes-client/python/blob/master/examples/watch/pod_namespace_watch.py) instead of polling every few seconds. Watches provide a more efficient detection, as explained in [Kubernetes API Concepts](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes).

Could you implement your change to use watches instead of polling?

sanjacodes (changes requested):
See [comment](https://github.com/chaos-kubox/krkn/pull/240#issuecomment-1116393063).

sanjacodes (approved):
We'll add an issue to rework instances where we use polling instead of watches - let's merge this PR since it solves what we need right now.

Thanks for the PR, @obrienrobert!

paigerube14:
Issue for rework with watches instead: https://github.com/chaos-kubox/krkn/issues/243

chaitanyaenr:
Nice addition @obrienrobert! Thanks.

PR #244, opened 2022-05-04
Fix reusable workflows top-level reference issue

sanjacodes:
Fixes _reusable workflows should be referenced at the top-level `jobs.*.uses' key, not within steps_ error with Actions - using composite workflow instead of reusable workflow now.


chaitanyaenr (approved):
LGTM

PR #248, opened 2022-05-06
Throwing exceptions in Kubernetes on failure

janosdebugs:
This PR fixes several cases where a Kubernetes API failure would lead to the error being logged, but not handled. This would result in undefined behavior.


paigerube14 (approved):
/lgtm

yogananth-subramanian (approved):
/lgtm

PR #249, opened 2022-05-06
Fixes incorrect command nesting in kube_burner

janosdebugs:
This PR fixes incorrect nesting in kube_burner, where the code would only work if the `prometheus_url` was *not* defined.


chaitanyaenr (approved):
LGTM

PR #253, opened 2022-05-09
Rework node status to use watches

obrienrobert:
A very early draft to see if I am going in the right direction with this. Contains three functions (ready/not ready/unknown) that use a watch to determine the status of the given node. Not sure if unknow/not ready can be rolled together into the same function. Related to [krkn/issues/243](https://github.com/chaos-kubox/krkn/issues/243).

It would be possible to create a single function (`get_node_status`) that accepts the requested status as a parameter and then watches the node for the given status. It just might make the function calls harder to read, as the only thing differentiating them in the code is the status parameter.

Haven't tested this with a Krkn test, yet, but I have tested it using a simple python script to ensure that the functions return the correct status.


kraken/node_actions/common_node_functions.py
```diff
@@ -83,10 +137,21 @@ def check_service_status(node, service, ssh_private_key, timeout):
...
+            logging.error(
+                "Service %s is in %s state" % (service_name, service_status.strip())
+            )
     ssh.close()
+
+
+# Get the resource version for the specified node
+def get_node_resource_version(node):
+    v1 = client.CoreV1Api()
+    return v1.read_node(name=node).metadata.resource_version
```
paigerube14, line 157: Let's add this call to https://github.com/chaos-kubox/krkn/blob/main/kraken/kubernetes/client.py (the corev1api is already set up there) and make a common function just to read node and then get the needed resource version so we could reuse the read_node to do a variety of calls

kraken/node_actions/common_node_functions.py
```diff
@@ -30,36 +33,83 @@ def get_node(node_name, label_selector, instance_kill_count):
...
-                break
-        except Exception:
-            logging.error("Encountered error while getting node status, waiting 3 seconds and retrying")
-        time.sleep(3)
-    node_status = kubecli.get_node_status(node, timeout)
-    logging.info("node status " + str(node_status))
-    if node_status is not None and node_status != "Unknown":
-        raise Exception("Node condition status isn't Unknown after %s seconds" % str(timeout))
+    w = watch.Watch()
+    v1 = client.CoreV1Api()
```
paigerube14, line 63: again, let's move this to https://github.com/chaos-kubox/krkn/blob/main/kraken/kubernetes/client.py so all functionality/setting of CoreV1Api is in that file

kraken/node_actions/common_node_functions.py
```diff
@@ -30,30 +30,58 @@ def get_node(node_name, label_selector, instance_kill_count):
...
 
 
-# Wait till node status becomes Ready
+# Wait until the node status becomes Ready
 def wait_for_ready_status(node, timeout):
-    for _ in range(timeout):
-        if kubecli.get_node_status(node) == "Ready":
+    w = watch.Watch()
+    v1 = client.CoreV1Api()
+    for event in w.stream(
```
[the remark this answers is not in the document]
obrienrobert: I've made the change to capture the resource version.

kraken/kubernetes/client.py
```diff
@@ -281,20 +285,18 @@ def get_job_status(name, namespace="default"):
...
-            "Exception when calling \
-                       CoreV1Api->read_node_status: %s\n"
-            % e
-        )
-        return None
-    for condition in node_info.status.conditions:
-        if condition.type == "Ready":
-            return condition.status
+# Obtain the node status
+def get_node_status(node, status, timeout):
```
paigerube14, line 289: Are we using this function now? Seems to be doing  almost the same thing as watch_node_status.

kraken/kubernetes/client.py
```diff
@@ -388,3 +390,24 @@ def find_kraken_node():
...
+
+# Watch for a specific node status
+def watch_node_status(node, status, timeout, resource_version):
+    for event in watch_resource.stream(
+        cli.list_node,
+        field_selector=f"metadata.name={node}",
+        timeout_seconds=timeout,
+        resource_version=f"{resource_version}",
+    ):
+        conditions = [status for status in event["object"].status.conditions if status.type == "Ready"]
```
paigerube14, line 403: Need to add in a path for covering when the cluster becomes disconnected like what was added in [PR ](https://github.com/chaos-kubox/krkn/pull/217)

paigerube14:
I was successfully able to run the reboot scenario now on a 3 worker azure cluster, a couple of comments though

kraken/kubernetes/client.py
obrienrobert: I've removed this function in favour of `watch_node_status`.

kraken/kubernetes/client.py
```diff
@@ -388,3 +390,24 @@ def find_kraken_node():
         except Exception as e:
             logging.info("%s" % (e))
             sys.exit(1)
+
+
+# Watch for a specific node status
+def watch_node_status(node, status, timeout, resource_version):
+    for event in watch_resource.stream(
```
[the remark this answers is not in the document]
obrienrobert: I've added a decrementing counter variable into the loop, as it is done in the example: [pod_namespace_watch.py](https://github.com/kubernetes-client/python/blob/master/examples/watch/pod_namespace_watch.py).

sandrobonazzola:
@obrienrobert can you please rebase the patch?

Issue #185, opened 2021-11-23
Replace openshift client references with kube client


chaitanyaenr:
There are currently two areas that can be improved:
- While we are using the kubernetes python client in most of the code base, we still have a couple of places where we use the kubectl and oc clients which we need to replace wherever possible to avoid the need for having them installed on the host in case of using standlone way of running kraken ( python ) vs containers where this shouldn't be an issue given that we package the clients into the image.
- There are a couple of references of oc cli calls which can be replaced with kubectl. In case where an operation is specific to OpenShift, we need to make sure to use the distribution mentioned in the config to avoid running them in case of running kraken on kubernetes -  https://github.com/cloud-bulldozer/kraken/blob/master/config/config.yaml#L2 to avoid bombing out.

sanjacodes:
Let's try and replace all instances of oc and kubectl by using the [Kubernetes Python Client](https://github.com/kubernetes-client/python) wherever possible.

chaitanyaenr:
Fixed in the recent releases. Thanks.

PR #256, opened 2022-05-18
Starts fixing small parts of issue #185

sanjacodes:
Replaces 2 lines using kubectl in `run_kraken.py` with their Kubernetes Python client equivalent.


run_kraken.py
```diff
@@ -97,11 +97,26 @@ def main(cfg):
 
         # Cluster info
         logging.info("Fetching cluster info")
-        cluster_version = runcommand.invoke("kubectl get clusterversion", 60)
-        cluster_info = runcommand.invoke(
-            "kubectl cluster-info | awk 'NR==1' | sed -r " "'s/\x1B\[([0-9]{1,3}(;[0-9]{1,2})?)?[mGK]//g'", 60
-        )  # noqa
-        logging.info("\n%s%s" % (cluster_version, cluster_info))
+
+        kubernetes.config.load_kube_config()
```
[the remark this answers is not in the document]
sanjacodes: Yup, agree on not duplicating. Changed it.

kraken/kubernetes/client.py
```diff
@@ -23,6 +23,32 @@ def initialize_clients(kubeconfig_path):
         sys.exit(1)
 
 
+def get_host() -> str:
+    """Returns the Kubernetes server URL"""
+    return client.configuration.Configuration.get_default_copy().host
+
+
+def get_clusterversion_string() -> str:
+    """Returns clusterversion status text on OpenShift, empty string on other distributions"""
```
paigerube14, line 32: Small nit, some of the functions if they have a description/comment about what they do are above the function definition. Looks good to me other than that

sanjacodes: I tried to follow the [docstring conventions](https://peps.python.org/pep-0257/) and think we should continue with that as we refactor some of the code since it will help us with documentation long-term.

paigerube14 (approved):
/lgtm

kraken/kubernetes/client.py
paigerube14: Makes sense! Thanks

PR #257, opened 2022-05-18
Remove pre-commit contributor annoyance

sanjacodes:
It's really fine without being too opinionated - we can accommodate single and double quotes as well as variable line lengths. If anything, we could add PEP 8 as a requirement once we've refactored the code.


chaitanyaenr (approved):
LGTM

PR #258, opened 2022-05-20
Add support for setting performance-dashboards on Kubernetes

chaitanyaenr:


kraken/performance_dashboards/setup.py
```diff
@@ -5,7 +5,7 @@
 
 # Installs a mutable grafana on the Kubernetes/OpenShift cluster and loads the performance dashboards
 def setup(repo):
-    command = "cd /tmp/performance-dashboards/dittybopper && ./deploy.sh"
+    command = "cd /tmp/performance-dashboards/dittybopper && ./k8s-deploy.sh"
```
paigerube14, line 8: We do have a distribution setting in the config file, should we use that to depict which command to use? Will that script work both on kubernetes and openshift?

chaitanyaenr: @paigerube14 tweaked it use different install scripts based on the distribution like you suggested given that there might be changes in the future that will be added to the respective scripts depending on the distribution - OpenShift or Kubernetes. PTAL. Thanks.

paigerube14 (approved):
/lgtm

PR #259, opened 2022-05-26
Adding namespace deletion using kuberenetes python client

paigerube14:
Changing the namespace deletion scenario to use the kubernetes python client instead of "oc delete.."

I also took out the references to the "action" under the namespace scenario since the only action we are doing now is delete


ghost (approved):

chaitanyaenr (approved):
LGTM

PR #266, opened 2022-06-03
Fixes #265: Replace powerfulseal

### LINKED ISSUE #265: Replace powerfulseal
Currently, we are using [powerfulseal](https://github.com/powerfulseal/powerfulseal), but it is blocking upgrading the Kubernetes client from v18. (see powerfulseal/powerfulseal#356 and chaos-kubox/krkn#262)

We need to remove powerfulseal and create a new pod scenario functionality. The configuration file should look as follows:

@chaitanyaenr @paigerube14 please provide at least 3 sample configurations below for the functionality.

janosdebugs:
This PR replaces powerfulseal by a built-in pod scenario runner.


scenarios/kube/composite.schema.json
```diff
@@ -0,0 +1,89 @@
...
+    "examples": [{
+        "iterations": 1,
+        "steps": [{
+            "pod": {
+                "name_pattern": "test-.*",
+                "namespace_pattern": "default"
+            }
+        }]
+    }]
+}
```
sandrobonazzola, line 89: Please add newline at end of file

run_kraken.py
```diff
@@ -22,6 +24,9 @@
 import kraken.pvc.pvc_scenario as pvc_scenario
 import kraken.network_chaos.actions as network_chaos
 import server as server
+from kraken.scenarios.base import Scenario
+from kraken.scenarios.pod import PodScenario
+from kraken.scenarios.runner import ScenarioRunner
```
sandrobonazzola, line 30: Any reason for breaking import style consistency here? previous imports were including the whole module rather than specific classes within the module. Also, this way you're losing namespace for the imported classes.

kraken/scenarios/test_pod.py
```diff
@@ -0,0 +1,43 @@
...
+            self.fail("Getting the pod after a pod scenario run should result in a NotFoundException.")
+        except NotFoundException:
+            return
+
+    @staticmethod
+    def _remove_test_pod(c: kube.Client, pod_name: str, pod_namespace: str):
+        try:
+            c.remove_pod(pod_name, pod_namespace)
+        except NotFoundException:
+            pass
```
sandrobonazzola, line 39: No other action if trying to remove a non existent pod? Not even logging?

kraken/scenarios/pod.py
```diff
@@ -0,0 +1,96 @@
...
+        self.logger.info("Killing %d pods...", pods_to_kill)
+        for i in range(pods_to_kill):
+            pod = pod_candidates[i]
+            self.logger.info("Killing pod %s...", pod.name)
+            removed_pod.append(pod)
+            kube.remove_pod(pod.name, pod.namespace)
+
+        self.logger.info("Waiting for pods to be removed...")
+        for i in range(60):
+            time.sleep(1)
```
sandrobonazzola, line 83: I would consider setting constants instead of hardcoding numbers here. It would make it easier later on tuning the numbers rather than having to find them in the middle of the code. You may discover later than 1 second is too short for the loop or that 60 loops are not enough.

kraken/scenarios/base.py
```diff
@@ -0,0 +1,61 @@
...
+        thrown.
+        """
+        pass
+
+
+class TimeoutException(Exception):
+    """
+    TimeoutException is an exception thrown when a scenario has a timeout waiting for a condition to happen.
+    """
+    pass
```
sandrobonazzola, line 61: `pass` here shouldn't really be needed for the class definition

sandrobonazzola (changes requested):
See inline comments. @janosdebugs can you try to make smaller PRs next time?  :-)

This one seems a bit more than `Replace powerfulseal`. Looks like
- Replace powerfulseal
- Add unit testing
- Setup GitHub Actions for running unit testing
- Perform docker image build only if unit tests passes

and possibly another couple of PRs :-)

run_kraken.py
janosdebugs: Yes, ALT+Enter, import. :) I use an IDE and don't really look or even see the imports. Is this something we should pay attention to? If yes, is there an autofixer?

kraken/scenarios/test_pod.py
janosdebugs: No, the intent is for the pod to be gone. It's gone. We can log, but on the other hand, the entire client doesn't log.

kraken/scenarios/pod.py
janosdebugs: Good point.

scenarios/kube/composite.schema.json
janosdebugs: Why is that necessary in 2022?

run_kraken.py
sandrobonazzola: No hard feeling about this, if the IDE doesn't allow to configure this feature, I won't be the one telling you to not use it :-)

scenarios/kube/composite.schema.json
sandrobonazzola: new line character is line termination. Without it the last line of the file is not terminated.
It can cause issues with some tooling including git blame.
There are several discussions about it , for reference: 
* https://gist.github.com/camh-/1bebfcff1b0f814e9b191edc60d5206b
* https://en.wikipedia.org/wiki/Newline

kraken/scenarios/test_pod.py
sandrobonazzola: If I get it right, if NotFoundException is raised here it means that the test pod either didn't got created or crashed while running the scenario. Shouldn't it fail test here in such case?

scenarios/kube/composite.schema.json
janosdebugs: I just checked, none of Vim, Emacs, IntelliJ+friends, Atom, VS Code, TextMate, and Sublime require an exlicit setting to be enabled for automatically adding a newline, which is just as bad as reformatting a file in an unrelated PR. This newline thing is, as far as I can tell, outdated and no longer true.

run_kraken.py
janosdebugs: image

@sandrobonazzola this is what PyCharm supports. Which of these do you think would be sensible?

kraken/scenarios/test_pod.py
janosdebugs: No, `_remove_test_pod` gets called in the cleanup phase, which is there to ensure we don't leave junk in the k8s cluster. See line 19.

run_kraken.py
sandrobonazzola: Above settings looks good to me.
In oVirt we used [isort](https://pycqa.github.io/isort/) to establish common standard for imports.

Issue #267, opened 2022-06-14
Bug in network interface error message


startxfr:
When running the network scenarios with a bad interface, the action.py raise and error because of the message format:

```yaml
network_chaos:
  duration: 60
  node_name: 
  label_selector: 'node-role.kubernetes.io/infra'
  instance_count: 1
  interfaces:
  # - "tun0"
  # - "lo"
  - "ens5"
  execution: serial
  egress:
    latency: 50ms  
    loss: 0.05
    bandwidth: 100mbit
```
Result :

```
Traceback (most recent call last):
  File "/root/kraken/run_kraken.py", line 318, in <module>
    main(options.cfg)
  File "/root/kraken/run_kraken.py", line 239, in main
    network_chaos.run(scenarios_list, config, wait_duration)
  File "/root/kraken/kraken/network_chaos/actions.py", line 39, in run
    test_interface = verify_interface(test_interface, nodelst, pod_template)
  File "/root/kraken/kraken/network_chaos/actions.py", line 111, in verify_interface
    "Interface %s not found in node %s interface list %s" % (interface, nodelst[pod_index]),
TypeError: not enough arguments for format string
```

startxfr:
closed

PR #268, opened 2022-06-14
Solve issue #267

startxfr:
Fix for the #267 issue

### DIFF
--- a/kraken/network_chaos/actions.py
+++ b/kraken/network_chaos/actions.py
@@ -107,10 +107,7 @@ def verify_interface(test_interface, nodelst, template):
             interface_lst = output[:-1].split(",")
             for interface in test_interface:
                 if interface not in interface_lst:
-                    logging.error(
-                        "Interface %s not found in node %s interface list %s" % (interface, nodelst[pod_index]),
-                        interface_lst,
-                    )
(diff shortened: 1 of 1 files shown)


chaitanyaenr (approved):
Thanks for the PR @startxfr!

startxfr:
you're welcome. I discover this when building a chaos pipeline using tekton. Need to be optimized but if you're interested, i can write a description with sample files.

startxfr:
> > you're welcome. I discover this when building a chaos pipeline using tekton. Need to be optimized but if you're interested, i can write a description with sample files.
> 
> That would be super helpful @startxfr. Can we add it as part of the chaos testing guide as a new section after https://github.com/chaos-kubox/krkn/blob/main/docs/index.md#chaos-testing-in-practice-within-the-OpenShift-Organization to get published in https://chaos-kubox.github.io/krkn/ please?

OK, if you can tell me were to put this content into that page, i'll follow your advice. 

Furthermore, theses task and pipelines are deployed using argocd with a helm chart that can also deploy cerberus and kraken into their own namespaces. Chart create namespace, networkpolicy to allow kraken to talk to cerberus, SA, SCC and roleBinding for privilegied access, optionnal route for both as well as pipeline, tasks and configMap loaded with scenarios.
Are you also interested into getting examples and explanation on this part too ?

PR #269, opened 2022-06-15

Sau1506mya:


config/config.yaml
```diff
@@ -10,7 +10,8 @@ kraken:
     litmus_uninstall_before_run: True                      # If you want to uninstall litmus before a new run starts
     chaos_scenarios:                                       # List of policies/chaos scenarios to load
         -   container_scenarios:                                 # List of chaos pod scenarios to load
-            - -    scenarios/openshift/container_etcd.yml
+            #- -    scenarios/openshift/container_etcd.yml
```
paigerube14, line 13: be sure to either uncomment this line and take one of the lines out

chaitanyaenr:
Closing the PR due to inactivity. We can reopen as we see fit. Thanks.

PR #277, opened 2022-07-14
WIP: Continue fixing small parts of issue #185 

shreyasar2202:
- Adds a Pod and Container dataclass in the kuberenetes client python file. This is used to store information related to a pod and its containers
- Replaces few calls to runcommand.invoke to call kubectl get pods with the kuberenetes python client equivalent
- apply_yaml is used to create resources from yaml file (equivalent to kubectl apply -f commands). This will be used in the common_litmus.py file which has calls to kubectl apply -f .


chaitanyaenr:
Nice addition @shreyasar2202! Adding https://github.com/chaos-kubox/krkn/blob/main/kraken/application_outage/actions.py#L42 to the list of calls that can be replaced with https://github.com/shreyasar2202/krkn/blob/7a163026418e73c497aca6f69bd12742005833a2/kraken/kubernetes/client.py#L413.

chaitanyaenr (approved):

chaitanyaenr:
@shreyasar2202 We might want to switch this pull request to ready for review.

shreyasar2202:
I made the changes that @janosdebugs requested. Not sure why the build failed though as they were only documentation changes

kraken/kubernetes/client.py
```diff
@@ -14,10 +17,17 @@
...
     global batch_cli
+    global api_client
+    global dyn_client
+    global custom_object_client
     try:
         config.load_kube_config(kubeconfig_path)
         cli = client.CoreV1Api()
         batch_cli = client.BatchV1Api()
+        api_client = client.ApiClient()
+        custom_object_client = client.CustomObjectsApi()
```
paigerube14, line 28: Since we are now setting a global custom object api can you use that object in the function _get_clusterversion_string_. It is double initializing it

shreyasar2202:
Thank you @paigerube14. I have incorporated the changes.

PR #279, opened 2022-07-20
Removing cryptography pinning

janosdebugs:
Remove `cryptography=36.0.2` pinning because paramiko/paramiko#2038 is now fixed.


chaitanyaenr (approved):
LGTM

PR #285, opened 2022-08-09
Added VMware Node Scenarios

shreyasar2202:
The following VMware node scenarios have been added :  node-stop, node-start, node-reboot, node-terminate. In order to run the scenarios, the environment variables ```VSPHERE_IP, VSPHERE_USERNAME, VSPHERE_PASSWORD``` have to be set. Also, ```scenarios/openshift/vmware_node_scenarios.yml``` file should be changed accordingly. ```verify_session``` was set to ```False``` and ```skip_openshift_checks``` was set to ```False``` during testing. 

From  ```watch_node_status``` function, ```resource_version``` was removed because it used to change when the VM was powered off/on which was causing runtime errors.

### DIFF
--- a/docs/cloud_setup.md
+++ b/docs/cloud_setup.md
@@ -5,6 +5,7 @@ Supported Cloud Providers:
 * [Openstack](#openstack)
 * [Azure](#azure)
 * [Alibaba](#alibaba)
+* [VMware](#vmware)
 
 ## AWS
 
@@ -53,3 +54,17 @@ See the [Installation guide](https://www.alibabacloud.com/help/en/alibaba-cloud-
 Refer to [region and zone page](https://www.alibabacloud.com/help/en/elastic-compute-service/latest/regions-and-zones#concept-2459516) to get the region id for the region you are running on.
 
 Set cloud_type to either alibaba or alicloud in your node scenario yaml file.
+
+## VMware
+
+Set the following environment variables
+
+1. ```export VSPHERE_IP=<vSphere_client_IP_address>```
+
+2. ```export VSPHERE_USERNAME=<vSphere_client_username>```
+
+3. ```export VSPHERE_PASSWORD=<vSphere_client_password>```
+
+These are the credentials that you would normally used to access the vSphere client.
+
+
--- a/scenarios/openshift/vmware_node_scenarios.yml
+++ b/scenarios/openshift/vmware_node_scenarios.yml
@@ -1,0 +1,10 @@
+# yaml-language-server: $schema=../plugin.schema.json
+- id: <node-stop/node-start/node-reboot/node-terminate>
+  config:
+    name: <node_name>                   # Node on which scenario has to be injected; can set multiple names separated by comma                           
+    label_selector: <label_selector>    # When node_name is not specified, a node with matching label_selector is selected for node chaos scenario injection 
+    runs: 1                             # Number of times to inject each scenario under actions (will perform on same node each time)                                                           
+    instance_count: 1                   # Number of nodes to perform action/select that match the label selector                                             
+    timeout: 300                        # Duration to wait for completion of node scenario injection
+    verify_session: True                # Set to True if you want to verify the vSphere client session using certificates; else False
+    skip_openshift_checks: False        # Set to True if you don't want to wait for the status of the nodes to change on openshift before passing the scenario 
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -14,6 +14,7 @@ kraken:
         -   plugin_scenarios:
             - scenarios/openshift/etcd.yml
             - scenarios/openshift/regex_openshift_pod_kill.yml
+            - scenarios/openshift/vmware_node_scenarios.yml
         -   node_scenarios:                                # List of chaos node scenarios to load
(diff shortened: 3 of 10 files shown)


docs/cloud_setup.md
```diff
@@ -53,3 +54,17 @@ See the [Installation guide](https://www.alibabacloud.com/help/en/alibaba-cloud-
...
+
+Set the following environment variables
+
+1. ```export VSPHERE_IP=<vSphere_client_IP_address>```
+
+2. ```export VSPHERE_USERNAME=<vSphere_client_username>```
+
+3. ```export VSPHERE_PASSWORD=<vSphere_client_password>```
+
+These are the credentials that you would normally used to access the vSphere client.
```
sanjacodes, line 68: used -> use

kraken/plugins/vmware/kubernetes_functions.py
```diff
@@ -0,0 +1,179 @@
+from kubernetes import config, client
+from kubernetes.client.rest import ApiException
+import logging
+import random
+from enum import Enum
+
+
+class Actions(Enum):
+    """
+    This enumeration is used to indicate the kind of node operation being done
```
sanjacodes, line 10: You could write "This enumeration indicates different kinds of node operations"

kraken/plugins/vmware/kubernetes_functions.py
```diff
@@ -0,0 +1,179 @@
...
+    for node in ret.items:
+        for cond in node.status.conditions:
+            if str(cond.type) == "Ready" and str(cond.status) != "True":
+                nodes.append(node.metadata.name)
+    return nodes
+
+
+def get_node_list(cfg, action, core_v1):
+    """
+    Return a list of nodes to be used in the node scenarios. The list returned is constructed as follows:
```
sanjacodes, line 84: Return -> Returns

scenarios/openshift/vmware_node_scenarios.yml
```diff
@@ -0,0 +1,10 @@
+# yaml-language-server: $schema=../plugin.schema.json
+- id: <node-stop/node-start/node-reboot/node-terminate>
+  config:
+    name: <node_name>                   # Node on which scenario has to be injected; can set multiple names separated by comma                           
+    label_selector: <label_selector>    # When node_name is not specified, a node with matching label_selector is selected for node chaos scenario injection 
+    runs: 1                             # Number of times to inject each scenario under actions (will perform on same node each time)                                                           
+    instance_count: 1                   # Number of nodes to perform action/select that match the label selector                                             
+    timeout: 300                        # Duration to wait for completion of node scenario injection
+    verify_session: True                # Set to True if you want to verify the vSphere client session using certificates; else False
+    skip_openshift_checks: False        # Set to True if you don't want to wait for the status of the nodes to change on openshift before passing the scenario
```
sanjacodes, line 10: openshift -> OpenShift

sanjacodes (approved):
Excellent work, thanks a bunch for this great PR!

kraken/plugins/vmware/kubernetes_functions.py
```diff
@@ -0,0 +1,179 @@
+from kubernetes import config, client
```
paigerube14, line 1: I think this file needs to be moved up to the plugins base folder or rename the vmware folder to be renamed to nodes. This code is general that should be able to be used by almost every cloud provider

paigerube14: will each cloud type for the node scenarios be considered a separate plugin each needing this same set of functions duplicated? I think I'm not super clear on the architecture here.

scenarios/openshift/vmware_node_scenarios.yml
shreyasar2202: Thanks for spotting these! Fixed them now.

paigerube14:
Got the tests to properly run and things look good to me!

kraken/plugins/vmware/kubernetes_functions.py
paigerube14: Think this is something we can address as we keep adding and have the addition of the engine and grasping more of the plugin ideology.

chaitanyaenr (approved):
LGTM. Tested few scenarios on a live vSphere environment. Nice work @shreyasar2202!

kraken/plugins/vmware/kubernetes_functions.py
chaitanyaenr: While I see and agree with how the plugins should be self contained, there's going to be lot of code duplication for the scenarios in Kraken like @paigerube14 mentioned. Maybe we can carry the kubernetes or any other helper functions for the plugins hosted in Kraken repository i.e chaos scenarios and eventually orchestrate the chaos plugins using the engine? 

I need to go through the arcaflow workflow to understand how engine orchestrates the plugins - if it's containers then we should be good as long as we package the helper functions in the image. If it's going to be a centralized repository for all the plugins irrespective of chaos, then it might not work like @janosdebugs suggested. Ideally, it shouldn't matter for the engine to orchestrate things I guess.

We can merge this PR and iterate on addressing depending on the design. 

Thoughts?

paigerube14 (approved):
LGTM

PR #288, opened 2022-08-16
Create prometheus token to use for OCP versions >=4.11

chaitanyaenr:
This commit adopts the code from https://github.com/redhat-chaos/cerberus/pull/176
to support using the existing/creating new token needed to query prometheus depending
on the OpenShift version in use.

Co-authored-by: Paige Rubendall <prubenda@redhat.com>


paigerube14 (approved):
LGTM

PR #290, opened 2022-08-17
Fixing parts of issue #185 for PVC scenario

shreyasar2202:
- Created a new file called ```kraken/kubernetes/resources.py`` which contains dataclasses for several Kubernetes resources and moved the dataclasses from ```kraken/kubernetes/client.py``` to the new file 
- Modified ```kraken/pvc/pvc_scenario.py``` to replace ```runcommand.invoke``` commands with their Kubernetes python client equivalent.
-  Several pods can be bound to the same PVC. So, in case a PVC name is supplied which has 2 or more pods bound to it, a pod is chosen randomly


kraken/pvc/pvc_scenario.py
```diff
@@ -46,64 +46,50 @@ def run(scenarios_list, config):
...
-                volumes_list = runcommand.invoke(command, 60).rstrip()
-                volumes_list_json = json.loads(volumes_list)
-                for entry in volumes_list_json:
-                    if len(entry["persistentVolumeClaim"]["claimName"]) > 0:
-                        volume_name = entry["name"]
-                        pvc_name = entry["persistentVolumeClaim"]["claimName"]
+                pod = kubecli.get_pod_info(name=pod_name, namespace=namespace)
+                
+                if pod is None:
+                    sys.exit(1)
```
paigerube14, line 65: Let's give a reason here as to why kraken exited

kraken/pvc/pvc_scenario.py
```diff
@@ -46,64 +46,50 @@ def run(scenarios_list, config):
...
-                command = "kubectl describe pvc %s -n %s | grep \"Capacity:\" | grep -Eo '[^: ]*$'" % (
-                    str(pvc_name),
-                    str(namespace),
-                )
-                pvc_capacity = runcommand.invoke(
-                    command,
-                    60,
-                ).rstrip()
-                logging.debug("Get PVC capacity command:\n %s" % command)
+                pvc_capacity = pvc.capacity
```
paigerube14, line 92: This capacity is a little off from the actual capacity in the PVC 

Can do something similar to what we do to find the Used bytes on the system 
`command = "df %s -B 1024 | sed 1d | awk -F' ' '{print $4}'" % (str(mount_path))`


Output I saw: 
`2022-08-24 21:28:47,466 [INFO] Pod name: kraken-test-pod
2022-08-24 21:28:50,026 [INFO] Volume name: kraken-test-pv
2022-08-24 21:28:50,026 [INFO] PVC name: kraken-test-pvc
2022-08-24 21:28:50,027 [INFO] Container path: kraken-test-container
2022-08-24 21:28:50,027 [INFO] Mount path: /home/krake-dir/
2022-08-24 21:28:50,027 [INFO] PVC capacity: 2097152 KB
2022-08-24 21:28:50,251 [INFO] PVC used: 10307228 KB
2022-08-24 21:28:50,251 [ERROR] 
                        Target fill percentage (50.00%) is lower than current fill percentage (491.49%)
                         or higher than 99%

% oc exec kraken-test-pod -- df /home/krake-dir/ 
Filesystem     1K-blocks     Used Available Use% Mounted on
/dev/nvme0n1p4 523752428 10307356 513445072   2% /home/krake-dir
`

shreyasar2202: Modified this in the latest commit

kraken/pvc/pvc_scenario.py
shreyasar2202: Added this in the latest commit

kraken/pvc/pvc_scenario.py
```diff
@@ -46,65 +46,52 @@ def run(scenarios_list, config):
...
-                    str(pvc_name),
-                    str(namespace),
-                )
-                pvc_capacity = runcommand.invoke(
-                    command,
-                    60,
-                ).rstrip()
-                logging.debug("Get PVC capacity command:\n %s" % command)
-                pvc_capacity_kb = toKbytes(pvc_capacity)
+                command = "df %s -B 1024 | sed 1d | awk -F' ' '{print $4}'" % (str(mount_path))
```
paigerube14, line 93: thoughts on combining this line with line 98 to only have to do the exec call once, then do the parsing of the output in the script to get the capacity and used bytes

shreyasar2202: That makes sense, I've modified it now

paigerube14 (approved):
/lgtm

PR #292, opened 2022-08-23
Add option to skip litmus installation

chaitanyaenr:
This commit adds an option for the user to pick whether to install
litmus or not depending on their use case. One use case is disconnected
environments where litmus is pre-installed insted of reaching out to the
internet.


paigerube14 (approved):
/lgtm

PR #298, opened 2022-09-01
Document node scenario actions for VMware

chaitanyaenr:


shreyasar2202 (approved):
We specify that ```node_stop_start_scenario``` is not supported on VMware. Should we do the same for all the other non supported scenarios in the list? Otherwise, LGTM

chaitanyaenr:
> We specify that `node_stop_start_scenario` is not supported on VMware. Should we do the same for all the other non supported scenarios in the list? Otherwise, LGTM

Hi @shreyasar2202, other scenarios in the list rely on Kubernetes API to run, so they should work on all the cloud providers.

docs/node_scenarios.md
```diff
@@ -2,21 +2,22 @@
 
 The following node chaos scenarios are supported:
 
-1. **node_start_scenario**: Scenario to stop the node instance.
-2. **node_stop_scenario**: Scenario to stop the node instance.
-3. **node_stop_start_scenario**: Scenario to stop and then start the node instance.
-4. **node_termination_scenario**: Scenario to terminate the node instance.
-5. **node_reboot_scenario**: Scenario to reboot the node instance.
+1. **node_start_scenario**: Scenario to stop the node instance. node_start is the action/id when running on VMware.
```
paigerube14, line 5: Think it would be better if we just updated the vmware_plugin.py and scenario file to be consistent with the wording we used currently in the node scenarios.

chaitanyaenr: Agreed. Updated the id's in vmware scenarios to match rest of the node scenarios. PTAL. Thanks.

paigerube14: @chaitanyaenr should you update this documentation now?

paigerube14 (approved):
/lgtm

PR #330, opened 2022-09-13
Really use CI test results

sandrobonazzola:
Dropped unsupported tests, fail on failing test.

### DIFF
--- a/CI/tests/my_tests
+++ b/CI/tests/my_tests
@@ -1,4 +1,3 @@
-test_app_outages
 test_container
 test_io_hog
 test_mem_hog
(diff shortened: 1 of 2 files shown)


CI/tests/my_tests
```diff
@@ -1,12 +1 @@
...
-test_nodes
-test_time
-test_app_outages
-test_container
-test_zone
-test_io_hog
-test_mem_hog
-test_cpu_hog
-test_shut_down
 test_net_chaos
```
chaitanyaenr, line 1: Are we running the network chaos test ( https://github.com/redhat-chaos/krkn/blob/main/CI/scenarios/network_chaos.yaml ) on Kind?

sandrobonazzola: Yes, it passes there.

paigerube14 (approved):
/lgtm

PR #345, opened 2022-10-18
refactor: use arcaflow plugin

sandrobonazzola:
Use plugin from arcalot repo while waiting for it to be available on PyPi

Signed-off-by: Sandro Bonazzola <sbonazzo@redhat.com>


requirements.txt
```diff
@@ -29,3 +29,4 @@ aliyun-python-sdk-ecs==4.24.25
 arcaflow-plugin-sdk
 wheel
 git+https://github.com/vmware/vsphere-automation-sdk-python.git@v8.0.0.0
+git+https://github.com/arcalot/arcaflow-plugin-kill-pod.git@main
```
chaitanyaenr, line 32: Hitting "arcaflow-plugin-kill-pod 0.0.0 requires arcaflow-plugin-sdk>=0.8.0, but you'll have arcaflow-plugin-sdk 0.3.0 which is incompatible". We might want to add in a release tag for arcaflow-plugin-sdk. Thoughts?

paigerube14: do we want to just add to the installation that we should always do 'pip install -U -r requirements' so that we continue to get the latest release of _arcaflow-plugin-sdk_?

paigerube14, line 32: Is this going to stay as a separate repo or is it going to be merged into https://github.com/arcalot/arcaflow-plugin-sdk-python in the future?

sandrobonazzola: correct, it will be one plugin per repo

chaitanyaenr (approved):
LGTM

PR #346, opened 2022-10-18
Reorganize the content from https://github.com/startx-lab

chaitanyaenr:
Moving the content around installing kraken using helm to the chaos in practice section of the guide to showcase how startx-lab is deploying and leveraging Kraken.


docs/index.md
```diff
@@ -258,14 +261,47 @@ to reflect your cluster configuration. Refer to the [kraken configuration](https
...
 ```
 
-#### Start as a pipelinerun
+3) Start as a pipelinerun
 
 ```yaml
 oc apply -f https://github.com/startxfr/tekton-catalog/raw/stable/task/kraken-scenario/0.1/samples/pipelinerun.yaml
 ```
+
+4) Deploying kraken using a helm-chart
```
paigerube14, line 276: I'm not super clear about the tekton pipeline vs this installation type. Think this does the same thing as the above 3 steps? Not sure

chaitanyaenr: Right, it's just an alternate way of deploying and running Kraken.

paigerube14: Having these as numbers confused me to thinking that they were steps but they are independent from eachother, correct? Maybe use a bullet or sub title

docs/installation.md
```diff
@@ -59,39 +58,6 @@ If you want to build your own kraken image see [here](https://github.com/redhat-
 Refer [Instructions](https://github.com/redhat-chaos/krkn/blob/main/containers/README.md) on how to deploy and run Kraken as a Kubernetes/OpenShift deployment.
```
paigerube14, line 60: Maybe let's add a reference to the helm chart and tekton way of running things in this document but say it was created by a specific user so any issues with it need to be reported to the upstream repo

docs/index.md
chaitanyaenr: Right, they are independent from each other, will add a bullet.

chaitanyaenr: Done. Thanks.

docs/index.md
```diff
@@ -230,7 +230,10 @@ Within the OpenShift organization we use kraken to perform chaos testing through
 
     3. We are starting to add in test cases that perform chaos testing during an upgrade (not many iterations of this have been completed).
 
-### Using kraken as part of a tekton pipeline
+
+##### startx-lab
```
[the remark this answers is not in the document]
chaitanyaenr: Tweaked it and explicitly mentioned that the content is how Startx is leveraging Kraken for their use case and that the issues need to be reported at the links mentioned given that it's natively supported in Kraken. Thoughts?

docs/installation.md
chaitanyaenr: Think it might be better to just have pointers in the guide to rule out any confusions given that the tekton way of running things is not natively maintained in Kraken. Thoughts?

chaitanyaenr:
@paigerube14 updated the commit to incorporate your suggestions. PTAL when you get time. Thanks.

docs/index.md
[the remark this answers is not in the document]
paigerube14: Hey Ravi, what are your thoughts of moving the specifics of how to run in tekton and helm to another file all together. I feel like this is a lot of very specific data for this more generalized file. Thoughts? Maybe we can create a usage folder or something

docs/installation.md
paigerube14: Maybe lets keep a reference to using this type of installation type but give the same notice as the other spot in the index doc. Again think a separate file all together might be good for this info though

paigerube14 (approved):
/lgtm

PR #351, opened 2022-11-04
Allow users to set the listening address

chaitanyaenr:
This commit provides an option for the user to set the listening address for the signal. This also fixes a security vulnerability.

Fixes https://github.com/redhat-chaos/krkn/issues/307

### DIFF
--- a/run_kraken.py
+++ b/run_kraken.py
@@ -48,7 +48,8 @@ def main(cfg):
         publish_running_status = config["kraken"].get(
             "publish_kraken_status", False
         )
-        port = config["kraken"].get("port", "8081")
+        port = config["kraken"].get("port")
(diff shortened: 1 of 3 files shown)


run_kraken.py
```diff
@@ -113,7 +114,7 @@ def main(cfg):
                 "Using port 8081 as %s isn't a valid port number" % (port)
             )
             port = 8081
```
paigerube14, line 116: think you need to not set this here and have a get of the port above from the config

chaitanyaenr: @paigerube14 tweaked it to exit(1) if the port passed by the user in the config is invalid. Also removed the defaults to force the user to know what address and port is being used. Thoughts?

paigerube14 (approved):
/lgtm

PR #355, opened 2022-11-08
Use autoescape=True to mitigate XSS vulnerabilities

chaitanyaenr:
Fixes https://github.com/redhat-chaos/krkn/issues/354


paigerube14 (approved):
/lgtm

PR #358, opened 2022-11-08
Use run dir instead of tmp

chaitanyaenr:
This commit also logs a message to handle the exception during the node checks.

Fixes https://github.com/redhat-chaos/krkn/issues/356, https://github.com/redhat-chaos/krkn/issues/357


paigerube14 (approved):
/lgtm

PR #382, opened 2023-02-03
Github Action + functional tests (no *hog tests)

tsebastiani:
This PR introduces the new *functional tests* GitHub Action on a disposable OpenShift instance 4.x leveraging the [crc-cloud](https://github.com/tsebastiani/crc-cloud) project created on purpose.
The GitHub action is triggered by commenting a pull request with the keyword ```/functional``` (the keyword can be changed) and to make it work **must** be first merged in the main branch.
The action will instantiate and dispose a fully functional OpenShift instance on AWS. The action will checkout the last commit in the PR branch from which the comment is added and the OpenShift instance will be disposed even if an error is encountered during the test run.
The following tests are actually executed;
|Name|
|---|
|test_app_outages_gh|
|test_container|
|test_namespace|
|test_net_chaos|
|test_time|

 The following secrets must be created in the repository :

|Secret|Description|
|---|---|
|AWS_ACCESS_KEY_ID| AWS access key ID|
|AWS_SECRET_ACCESS_KEY| AWS secret access key|
|KUBEADMIN_PWD| password for the kubeadmin account in the newly created OpenShift Instance|
|DEVELOPER_PWD| password for the developer account in the newly created OpenShift Instance|
| REDHAT_PWD | password for the redhat account in the newly created OpenShift Instance|
|PULL_SECRET| a base64 encoded pull secret that can be downloaded from [here](https://console.redhat.com/openshift/create/local) after registrating or logging in with a RedHat account|

The Action will produce a test report containing the Passed/Failed tests and the code coverage in the action summary. From my experiments the full run takes approximately 23 minutes.

### DIFF
--- a/.github/workflows/functional_tests.yaml
+++ b/.github/workflows/functional_tests.yaml
@@ -1,0 +1,93 @@
+on: issue_comment
+
+jobs:
+  pr_commented:
+    # This job only runs for pull request comments containing /functional
+    name: Functional Tests
+    if: contains(github.event.comment.body, '/functional') && contains(github.event.comment.html_url, '/pull/')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out Kraken
+        uses: actions/checkout@v3
+      - name: Checkout Pull Request
+        run: hub pr checkout ${{ github.event.issue.number }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install OC CLI
+        uses: redhat-actions/oc-installer@v1
+        with:
+          oc_version: latest
+      - name: Install python 3.9
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.9' 
+      - name: Setup kraken dependencies
+        run:  pip install -r requirements.txt
+      - name: Create Workdir & export the path
+        run: |
+         mkdir workdir
+         echo "WORKDIR_PATH=`pwd`/workdir" >> $GITHUB_ENV
+      - name: Teardown CRC (Post Action)
+        uses: webiny/action-post-run@3.0.0
+        id: post-run-command
+        with:
+          # currently using image coming from tsebastiani quay.io repo
+          # waiting that a fix is merged in the upstream one
+          # post action run cannot (apparently) be properly indented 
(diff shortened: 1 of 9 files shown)


tsebastiani:
> Have you added the secrets to this main github page already? We might want to merge the workflow and then the code changes separately to be able to run the workflow against the test changes. Thoughts?

Paige regarding the AWS credentials and the pull secret we should ask to the management to provide us them, regarding the password instead we can simply generate them and they will work, anyway I'm afraid that I don't have the grants to set them in the krkn repo :-\

chaitanyaenr:
@tsebastiani sent you an invite, do let me know if it allows you to add secrets to the project. Thanks.

tsebastiani:
> @tsebastiani sent you an invite, do let me know if it allows you to add secrets to the project. Thanks.

I can access some settings, but not the actions and secret section apparently.

CI/scenarios/ingress_namespace.yaml
```diff
@@ -1,7 +1,7 @@
 scenarios:
-- action: delete
```
chaitanyaenr, line 2: We might want to change the name of the file to something along the lines of pod scenario. Thoughts?

CI/tests/test_cpu_hog_gh.sh
```diff
@@ -0,0 +1,20 @@
+set -xeEo pipefail
```
chaitanyaenr, line 1: Do you think we can remove the hog scenario scripts given that they are disabled for the runs? Thoughts?

tsebastiani: do you want me to remove all the hog scenarios or only the _gh that I added? If that's the case should I remove all the tests that are not ran? I was keeping them just in case we figure out how to make them work.

CI/scenarios/ingress_namespace.yaml
tsebastiani: Sorry I don't understand, this scenario actually deletes the namespace, can you give me an example?

CI/tests/test_cpu_hog_gh.sh
chaitanyaenr: Let's add for them for now @tsebastiani. Thanks.

.github/workflows/functional_tests.yaml
```diff
@@ -0,0 +1,89 @@
+on: issue_comment
+
+jobs:
+  pr_commented:
+    # This job only runs for pull request comments containing /functional
+    name: Functional Tests
+    if: contains(github.event.comment.body, '/test') && contains(github.event.comment.html_url, '/pull/')
```
chaitanyaenr, line 7: @tsebastiani the job gets triggered only if one of the project members in github comments /test right?

paigerube14 (approved):
/lgtm

chaitanyaenr (approved):
LGTM

chaitanyaenr:
we can merge it in once rebased. Thanks.

PR #384, opened 2023-02-10
arcaflow integration POC

tsebastiani:
This Draft PR is an integration proposal between Arcaflow and Kraken using the newly created python Pypi wheel.


scenarios/arcaflow/sysbench/input.yaml
```diff
@@ -0,0 +1,5 @@
+kubeconfig: {}
+sysbench_cpumaxprime: 12000
```
chaitanyaenr, line 2: We need to expose two more options to the user: 1) Node to target on OpenShift/Kubernetes env 2) Percent CPU to hog given that we might not know exact number of cores available. Thoughts?

chaitanyaenr: Options exposed on the arca plugin side

scenarios/arcaflow/sysbench/input.yaml
```diff
@@ -0,0 +1,5 @@
+kubeconfig: {}
```
chaitanyaenr, line 1: Can we read the kubeconfig from the location specified in the Kraken config vs having to define in this config? Thoughts?

scenarios/arcaflow/kill-pod/workflow.yaml
```diff
@@ -0,0 +1,42 @@
...
+            type_id: string
+        kill:
+          display:
+            description: the number of pods to kill
+            name: kill
+          type:
+            type_id: integer
+steps:
+  kill_pod:
+    plugin: quay.io/tsebastiani/arcaflow/kill-pod:latest
```
chaitanyaenr, line 34: We might want to move this image to redhat-chaos quay org - https://quay.io/organization/redhat-chaos.

chaitanyaenr:
Ran the sysbench workflow on an OpenShift cluster and it worked!! We can merge the integration in post testing the pod-kill plugin. TO DO in later pull requests: Replace litmus scenarios after tweaking the PR to add the node-selector and CPU percentage options to let users pick a node to hog the % CPU specified in the config.

scenarios/arcaflow/kill-pod/workflow.yaml
```diff
@@ -0,0 +1,42 @@
+input:
```
chaitanyaenr, line 1: We might want to tweak the docs to walk through the workflow and how to tweak it as per user needs: https://github.com/redhat-chaos/krkn/blob/main/docs/pod_scenarios.md.

chaitanyaenr (approved):
LGTM. Just one comment: maybe we can add docs on the integration as to how users can trigger workflows or we can do that as part of the pod-kill workflow addition in the follow up PR? Thoughts?

paigerube14:
Agree with Ravi we should add some documentation to the workflows that are in arcaflow and give more links to configuration details

PR #389, opened 2023-02-23
removed useless step that was failing

tsebastiani:
removed useless step from github action that was making the pipeline fail


paigerube14 (approved):
/lgtm

PR #397, opened 2023-03-27
Exit on critical alert failures

chaitanyaenr:
This commit captures and exits on non-zero return code i.e when critical alerts are fired

Fixes https://github.com/redhat-chaos/krkn/issues/396


paigerube14 (approved):
/lgtm

PR #399, opened 2023-04-06
Adding privileged namespace

paigerube14:
To be able to properly run the pvc pod we need to create and add the pod to a privileged namespace


chaitanyaenr (approved):
LGTM

PR #411, opened 2023-04-27
Add support to check for critical alerts

chaitanyaenr:
This commit enables users to opt in to check for critical alerts firing in the cluster after certain amount of soak time for the cluster to settle down. Chaos scenario is considered as failed if the cluster is unhealthy in which case user can start debugging to fix and harden respective areas.

Fixes https://github.com/redhat-chaos/krkn/issues/410


config/config.yaml
```diff
@@ -65,7 +65,8 @@ performance_monitoring:
     uuid:                                                 # uuid for the run is generated by default if not set
     enable_alerts: False                                  # Runs the queries specified in the alert profile and displays the info or exits 1 when severity=error
     alert_profile: config/alerts                          # Path to alert profile with the prometheus queries
-
+    check_critical_alerts: False                          # When enabled will check prometheus for critical alerts firing post chaos after soak time for the cluster to settle down
```
chaitanyaenr, line 68: Add docs.

config/config_kubernetes.yaml
```diff
@@ -32,7 +32,8 @@ performance_monitoring:
     uuid:                                                 # uuid for the run is generated by default if not set
     enable_alerts: False                                  # Runs the queries specified in the alert profile and displays the info or exits 1 when severity=error
     alert_profile: config/alerts                          # Path to alert profile with the prometheus queries
-
+    check_critical_alerts: False                          # When enabled will check prometheus for critical alerts firing post chaos after soak time for the cluster to settle down
+    critical_alerts_check_soak_time: 120                  # Soak time for the cluster to settle down before checking for critical alerts. Default: 2 mins
```
yogananth-subramanian, line 36: Maybe we can reuse wait_duration, since critical alerts need to be checked at the end of each chaos run and before the next chaos scenario, we could wait on wait_duration and verify for critical alerts if enabled.

chaitanyaenr: Fixed it. Also updated it to run the critical alerts check at the end of each scenario in case multiple scenarios are defined in the config like we discussed. Thanks.

paigerube14:
Think we should note that for kuberenetes configurations a user will have to set the prometheus url and token for these checks as well. 

I got the below error when I was running on a 4.14 cluster. Think if [reorder the commands](https://github.com/redhat-chaos/krkn/blob/900ca74d8026e8d8f770150d22457542b6efcf2f/kraken/prometheus/client.py#L13-L15) to the below no error will fire
```
2023-05-02 15:37:15,466 [INFO] Checking for critical alerts firing post choas
Command "get-token" is deprecated, and will be removed in the future version. Use oc create token instead.
error: could not find a service account token for service account "prometheus-k8s"
2023-05-02 15:37:16,578 [INFO] No critical alerts are firing!!
```

```
            "oc create token -n openshift-monitoring prometheus-k8s  --duration=12h "
            "|| oc -n openshift-monitoring sa get-token prometheus-k8s "
            "|| oc sa new-token -n openshift-monitoring prometheus-k8s"
```

chaitanyaenr:
> "oc create token -n openshift-monitoring prometheus-k8s  --duration=12h "
>             "|| oc -n openshift-monitoring sa get-token prometheus-k8s "
>             "|| oc sa new-token -n openshift-monitoring prometheus-k8s"

Done! Thanks.

paigerube14 (approved):

PR #415, opened 2023-05-03
Add missing parameters supported by container scenarios

chaitanyaenr:
Also renames retry_wait to expected_recovery_time to make it clear that the Kraken will exit 1 if the container doesn't recover within the expected time.
Fixes https://github.com/redhat-chaos/krkn/issues/414


paigerube14 (approved):
/lgtm

PR #419, opened 2023-05-15
Extracted scenario container namespace as an input.yaml variable

tsebastiani:
with this change will be possible to set the scenario deployment namespace in the input.yaml instead of having it hardcoded in the workflow.


chaitanyaenr (approved):
lgtm

PR #429, opened 2023-05-25
Allow kraken to run with environment variables instead of kubeconfig file

josecastillolema:
This PR allows krkn to be run inside a k8s cluster using environment variables instead of a kubeconfig file.


chaitanyaenr (approved):
LGTM

PR #432, opened 2023-05-31
Change the severity of the etcd leader election check to warning

chaitanyaenr:
This is the first step towards the goal to only have metrics tracking the overall health and performance of the component/cluster. For instance, for etcd disruption scenarios, leader elections are expected, we should instead track etcd leader availability and fsync latency under critical catergory vs leader elections.


paigerube14 (approved):
/lgtm

PR #433, opened 2023-06-05
temporarly removed io-hog scenario

tsebastiani:
temporarly removed io-hog waiting the RBAC and namespace labeling implementation.


chaitanyaenr (approved):
LGTM

PR #435, opened 2023-06-09
Krkn telemetry integration

tsebastiani:
Adding telemetry data collection to krkn.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -69,3 +69,8 @@ telemetry:
     api_url: https://ulnmf9xv7j.execute-api.us-west-2.amazonaws.com/production
     username: username
     password: password
+    prometheus_backup: True
+    full_prometheus_backup: False
+    backup_threads: 5
+    archive_path: /tmp
+    max_retries: 0
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -65,14 +65,19 @@ tunings:
     iterations: 1                                          # Number of times to execute the scenarios
     daemon_mode: False                                     # Iterations are set to infinity which means that the kraken will cause chaos forever
 telemetry:
-    enabled: True
-    api_url: https://ulnmf9xv7j.execute-api.us-west-2.amazonaws.com/production
-    username: username
-    password: password
-    prometheus_backup: True
-    full_prometheus_backup: False
-    backup_threads: 5
-    archive_path: /tmp
-    max_retries: 0
-    run_tag: chaos
-    archive_size: 10000 #in KiloBytes
+    enabled: False                                           # enable/disables the telemetry collection feature
+    api_url: https://ulnmf9xv7j.execute-api.us-west-2.amazonaws.com/production #telemetry service endpoint
+    username: username                                      # telemetry service username
+    password: password                                      # telemetry service password
+    prometheus_backup: True                                 # enables/disables prometheus data collection
+    full_prometheus_backup: False                           # if is set to False only the /prometheus/wal folder will be downloaded.
+    backup_threads: 5                                       # number of telemetry download/upload threads
+    archive_path: /tmp                                      # local path where the archive files will be temporarly stored
+    max_retries: 0                                          # maximum number of upload retries (if 0 will retry forever)
+    run_tag: ''                                             # if set, this will be appended to the run folder in the bucket (useful to group the runs)
+    archive_size: 10000                                     # the size of the prometheus data archive size in KB. The lower the size of archive is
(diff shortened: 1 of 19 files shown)


config/config.yaml
```diff
@@ -64,3 +64,8 @@ tunings:
     wait_duration: 60                                      # Duration to wait between each chaos scenario
     iterations: 1                                          # Number of times to execute the scenarios
     daemon_mode: False                                     # Iterations are set to infinity which means that the kraken will cause chaos forever
+telemetry:
```
chaitanyaenr, line 67: @tsebastiani let's add documents on the feature as well as the data being collected when enabled.

config/config.yaml
```diff
@@ -64,3 +64,8 @@ tunings:
...
+    enabled: True
```
chaitanyaenr, line 68: We might want to disable it by default given that it's authenticated and users might not have the credentials.

run_kraken.py
```diff
@@ -203,9 +205,10 @@ def main(cfg):
                             )
                             sys.exit(1)
                         elif scenario_type == "arcaflow_scenarios":
```
chaitanyaenr, line 207: We will need to support telemetry for all the scenarios i.e telemetry as the positional arguments. Thoughts?

run_kraken.py
```diff
@@ -352,7 +355,8 @@ def main(cfg):
 
             iteration += 1
             logging.info("")
-
+        # send telemetry
+        telemetry.send_telemetry(config["telemetry"], chaos_telemetry)
```
chaitanyaenr, line 359: Output data being sent to the telemetry in stdout like we discussed. Maybe we can also add an option to output it to a file to save as artifacts for the run in case of CI. Thoughts?

run_kraken.py
```diff
@@ -352,7 +355,8 @@ def main(cfg):
 
             iteration += 1
             logging.info("")
-
+        # send telemetry
```
chaitanyaenr, line 358: Also document how to access the doc in S3 if user is authenticated.

chaitanyaenr:
Testing update: successfully ran hog scenarios with telemetry enabled. Here is the output stored in S3:
```
{
    "scenarios": [
        {
            "startTimeStamp": 1689037675,
            "endTimeStamp": 1689037762.1072097,
            "scenario": "scenarios/arcaflow/cpu-hog/input.yaml",
            "exitStatus": 0,
            "parametersBase64": "",
            "parameters": {
                "input_list": [
                    {
                        "cpu_count": 1,
                        "cpu_load_percentage": 80,
                        "cpu_method": "all",
                        "duration": "30s",
                        "kubeconfig": "anonymized",
                        "namespace": "default",
                        "node_selector": {}
                    }
                ]
            }
        },
        {
            "startTimeStamp": 1689037762,
            "endTimeStamp": 1689037797.4187903,
            "scenario": "scenarios/arcaflow/memory-hog/input.yaml",
            "exitStatus": 0,
            "parametersBase64": "",
            "parameters": {
                "input_list": [
                    {
                        "duration": "30s",
                        "kubeconfig": "anonymized",
                        "namespace": "default",
                        "node_selector": {},
                        "vm_bytes": "10%",
                        "vm_workers": 2
                    }
                ]
            }
        }
    ]
}
```

Nice addition. Thinking about a couple of enhancements: 
- Use uuid generated during the run as name of the document as an identifier
- Alerts triggered pre and post chaos ( separate entries )
- Thinking if OCP metadata can be part of the same document to make it easy to tie things together. Thoughts?

run_kraken.py
paigerube14: It should automatically go to cerberus.report too I believe

chaitanyaenr:
Testing update: 

Warnings around capturing builds
```
2023-07-27 19:27:00,903 [WARNING] CustomObjectsApi -> (404)
Reason: Not Found
{},"status":"Failure","message":"builds.config.openshift.io \"status\" not found","reason":"NotFound","details":{"name":"status","group":"config.openshift.io","kind":"builds"},"code":404}
```

run_kraken.py
```diff
@@ -352,7 +364,11 @@ def main(cfg):
...
+        request_id=str(uuid.uuid1())
```
chaitanyaenr, line 368: Thoughts on using the Krkn generated uuid: https://github.com/redhat-chaos/krkn/blob/main/run_kraken.py#L150?

chaitanyaenr:
@tsebastiani let's rebase this PR.

chaitanyaenr (approved):
LGTM!

PR #437, opened 2023-06-14
Expand SLOs profile to cover monitoring for more alerts

chaitanyaenr:
This commit:
- Also sets appropriate severity to avoid false failures for the test cases especially given that theses are monitored during the chaos vs post chaos. Critical alerts are all monitored post chaos with few monitored during the chaos that represent overall health and performance of the service.
- Renames Alerts to SLOs validation 
 
Metrics reference: https://github.com/cloud-bulldozer/kube-burner/blob/f09a492b1343079d5988290761145a0fff3fd23e/cmd/kube-burner/ocp-config/alerts.yml


paigerube14 (approved):
/lgtm

PR #443, opened 2023-06-17
Install oc and kubectl clients from stable releases

chaitanyaenr:
This makes sure latest clients are installed and used:
- This will avoid compatability issues with the server
- Fixes security vulnerabilities and CVEs


paigerube14 (approved):
/lgtm

PR #444, opened 2023-06-21
Switch from centos to redhat ubi base image

chaitanyaenr:
This replaces the base image for Kraken container images to use redhat ubi image to be more secure and stable.


paigerube14 (approved):
/lgtm

chaitanyaenr:
cc: @tsebastiani

PR #448, opened 2023-06-29
separating build and test workflows

paigerube14:
Want to separate out the workflows to run build separately when pushes happen to main 

Q: Do we need to run test coverage after it is merged?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,8 +1,5 @@
 name: Build Krkn
 on:
-  push:
-    branches:
-      - main
   pull_request:
 
 jobs:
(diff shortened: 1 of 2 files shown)


chaitanyaenr:
@paigerube14 I don't think we need test coverage post merge given that we run them pre-merge.

paigerube14:
@chaitanyaenr I updated the build.yml to only run on a pull-request. Think we might want with push on non main branches? 
I think what we have what you described in docker-image file. Or should I create a separate one for checking out main code only?

paigerube14:
Ravi, does this pr look ok? See the 2 checks in the pr that were ran separately

chaitanyaenr (approved):
LGTM

PR #450, opened 2023-07-20
Bake in yq dependency in Kraken container images

chaitanyaenr:
This commit also updates ppc64le image to have the latest bits.

### DIFF
--- a/containers/Dockerfile
+++ b/containers/Dockerfile
@@ -1,7 +1,5 @@
 # Dockerfile for kraken
 
-FROM quay.io/openshift/origin-tests:latest as origintests
-
 FROM mcr.microsoft.com/azure-cli:latest as azure-cli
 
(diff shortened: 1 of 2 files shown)


containers/Dockerfile
```diff
@@ -1,7 +1,5 @@
 # Dockerfile for kraken
 
-FROM quay.io/openshift/origin-tests:latest as origintests
```
paigerube14, line 3: I dont think you can delete this

chaitanyaenr: We used to use this image to copy the kubectl and oc clients, we are now installing them from a different source. Thoughts?

paigerube14: Ahh I see now, didnt see them at the bottom

paigerube14:
lets let the checks run and i think we are good to merge

paigerube14 (approved):

PR #455, opened 2023-07-24
Bake in virtualenv in krkn images

chaitanyaenr:
This is needed to tie the python version being used in case multiple versions are installed.


paigerube14 (approved):
/lgtm

PR #466, opened 2023-08-17
Use images tagged with a release for hog scenarios

chaitanyaenr:
This commit switches from using latest images to a specific release to review changes and update configs before using the latest bits.


paigerube14 (approved):
/lgtm

PR #472, opened 2023-08-23
update to plugin release stressng:0.3.0

dustinblack:
The stressng plugin was updated this week to 0.3.0, bringing an updated stress-ng release and minor changes to the input schema. Update the hog scenario workflows to use the new plugin release.


chaitanyaenr (approved):
LGTM!

PR #479, opened 2023-08-25
Prints the telemetry json after sending it to the webservice

tsebastiani:


run_kraken.py
```diff
@@ -387,11 +387,15 @@ def main(cfg):
...
             try:
-                telemetry.send_telemetry(config["telemetry"], telemetry_request_id, chaos_telemetry)
-                safe_logger.info("archives download started:")
-                prometheus_archive_files = telemetry.get_ocp_prometheus_data(config["telemetry"], telemetry_request_id)
-                safe_logger.info("archives upload started:")
-                telemetry.put_ocp_prometheus_data(config["telemetry"], prometheus_archive_files, telemetry_request_id)
+                telemetry_json = telemetry.send_telemetry(config["telemetry"], telemetry_request_id, chaos_telemetry)
+                if telemetry_json:
+                    telemetry_obj = ChaosRunTelemetry(json.loads(telemetry_json))
+                    logging.info(f"Telemetry data:\n{telemetry_obj.to_json()}")
```
chaitanyaenr, line 393: It might be useful if we capture and print the json even when telemetry is disabled i.e 
if telemetry is enabled we ship it to the centralized service location and print it as part of stdout
if telemetry is disabled we just print the json as part of stdout

Thoughts?

tsebastiani: :+1:

paigerube14 (approved):
/lgtm

Issue #350, opened 2022-11-02
Fix container scenarios to accept just the signal number for the action


chaitanyaenr:
The current container scenario allows users to pass an action with a command that gets run inside a container, this allows users to potentially run any command vs just passing the kill signal which is the intent. The config should look something like:
 
```
scenarios:
- name: "kill etcd container"
  namespace: "openshift-etcd"
  label_selector: "k8s-app=etcd"
  container_name: "etcd"
  action: "1" ( instead of kill 1 )
  count: 1
  retry_wait: 60
```

chaitanyaenr:
cc: @tsebastiani

PR #485, opened 2023-09-04
Fix container scenario to accept only signal number (#350)

jtydlack:


kraken/pod_scenarios/setup.py
```diff
@@ -127,14 +129,20 @@ def container_run(kubeconfig_path,
...
-
 def container_killing_in_pod(cont_scenario, kubecli: KrknKubernetes):
     scenario_name = cont_scenario.get("name", "")
     namespace = cont_scenario.get("namespace", "*")
     label_selector = cont_scenario.get("label_selector", None)
     pod_names = cont_scenario.get("pod_names", [])
     container_name = cont_scenario.get("container_name", "")
-    kill_action = cont_scenario.get("action", "kill 1")
+    kill_action = cont_scenario.get("action", 1)
+    if not isinstance(kill_action, int):
```
paigerube14, line 139: I dont think this if statement will ever be hit since we are setting a default if kill_action is not set

chaitanyaenr: Think this checks if kill_action is an int if set, if not set in the config we go with 1. Thoughts?

kraken/pod_scenarios/setup.py
```diff
@@ -127,14 +129,20 @@ def container_run(kubeconfig_path,
...
+        logging.error("Please make sure you inserted kill action number")
```
paigerube14, line 140: can you maybe clarify in the comment here to end with "as a number" or "as an int"

chaitanyaenr, line 140: Maybe change this to "Please make sure the action parameter in the config is an integer". Thoughts?

scenarios/openshift/container_etcd.yml
```diff
@@ -3,6 +3,6 @@ scenarios:
   namespace: "openshift-etcd"
   label_selector: "k8s-app=etcd"
   container_name: "etcd"
-  action: "kill 1"
+  action: 1
```
chaitanyaenr, line 6: We need to update https://github.com/redhat-chaos/krkn/blob/main/docs/container_scenarios.md#example-config and https://github.com/redhat-chaos/krkn-hub/blob/main/container-scenarios/env.sh#L9 post this PR merge.

docs/container_scenarios.md
```diff
@@ -15,7 +15,7 @@ scenarios:
   pod_names:  # This is optional, can take out and will select all pods with given namespace and label
   - <pod_name>
   count: <number of containers to disrupt, default=1>
-  action: <Action to run. For example kill 1 ( hang up ) or kill 9. Default is set to kill 1>
+  action: <number of kill action to run. For example 1 ( hang up ) or 9. Default is set to 1>
```
chaitanyaenr, line 18: s/number/signal

chaitanyaenr (approved):
LGTM!

PR #488, opened 2023-09-12
io-hog working configuration

tsebastiani:
It works instantiating a privileged container mounting the hostPath (seems to be the only solution to r/w on the node filesystem). It works on Vanilla OCP (4.14.0-0.nightly-2023-09-07-211548) needs to be tested on prow.

### DIFF
--- a/scenarios/arcaflow/io-hog/sub-workflow.yaml
+++ b/scenarios/arcaflow/io-hog/sub-workflow.yaml
@@ -1,0 +1,138 @@
+input:
+  root: RootObject
+  objects:
+    RootObject:
+      id: RootObject
+      properties:
+        kubeconfig:
+          display:
+            description: The complete kubeconfig file as a string
+            name: Kubeconfig file contents
+          type:
+            type_id: string
+          required: true
+        namespace:
+          display:
+            description: The namespace where the container will be deployed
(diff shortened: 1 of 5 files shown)


scenarios/arcaflow/io-hog/sub-workflow.yaml
```diff
@@ -0,0 +1,138 @@
...
+
+steps:
+  kubeconfig:
+    plugin: quay.io/arcalot/arcaflow-plugin-kubeconfig:0.2.0
+    input:
+      kubeconfig: !expr $.input.kubeconfig
+  stressng:
+    plugin: quay.io/arcalot/arcaflow-plugin-stressng:0.3.1
+    step: workload
+    input:
```
chaitanyaenr, line 104: We need to update with the recent changes similar to https://github.com/redhat-chaos/krkn/blob/main/scenarios/arcaflow/cpu-hog/sub-workflow.yaml#L69.

tsebastiani: Hey Ravi, seems to me that the structure is already exactly the same as the other two arcaflow scenarios (including the recent changes):

image

This is the diff between io-hog and cpu-hog.

PR #489, opened 2023-09-13
Fixed io-hog scenario

tsebastiani:
It works instantiating a privileged container mounting the hostPath (seems to be the only solution to r/w on the node filesystem). It works on Vanilla OCP (4.14.0-0.nightly-2023-09-07-211548) needs to be tested on prow.


chaitanyaenr (approved):
LGTM!

PR #499, opened 2023-09-25
Bump krkn-lib version

chaitanyaenr:


paigerube14 (commented):
I see there's a 1.2.0 krkn-lib version. should we set to that version?

chaitanyaenr:
> I see there's a 1.2.0 krkn-lib version. should we set to that version?

Looks like 1.2.0 tag exists but the release is missing.

chaitanyaenr:
> > I see there's a 1.2.0 krkn-lib version. should we set to that version?
> 
> Looks like 1.2.0 tag exists but the release is missing.

Maybe we should cut a new release in krkn-lib? Krkn should be able to pick up the latest given that we have set the minimum version in the requirements to v1.1.1. Thoughts? cc: @tsebastiani

tsebastiani:
folks if you can quickly review https://github.com/redhat-chaos/krkn-lib/pull/46 I can cut a new release and we can point the requirements to it.

tsebastiani:
> > > I see there's a 1.2.0 krkn-lib version. should we set to that version?
> > 
> > 
> > Looks like 1.2.0 tag exists but the release is missing.
> 
> Maybe we should cut a new release in krkn-lib? Krkn should be able to pick up the latest given that we have set the minimum version in the requirements to v1.1.1. Thoughts? cc: @tsebastiani
1.2.0 is on pypi as latest version https://pypi.org/project/krkn-lib/ .
If you  could give a look to  https://github.com/redhat-chaos/krkn-lib/pull/46 we can cut a new release containing the log collection piece and we can point the requirements to it.

tsebastiani (approved):

PR #514, opened 2023-10-16
Use function get_yaml_item_value

jtydlack:
Enables using default even though the value was loaded as None.

### DIFF
--- a/run_kraken.py
+++ b/run_kraken.py
@@ -30,6 +30,7 @@ from krkn_lib.k8s import KrknKubernetes
 from krkn_lib.telemetry import KrknTelemetry
 from krkn_lib.models.telemetry import ChaosRunTelemetry
 from krkn_lib.utils import SafeLogger
+from krkn_lib.utils.functions import get_yaml_item_value
 
 KUBE_BURNER_URL = (
     "https://github.com/cloud-bulldozer/kube-burner/"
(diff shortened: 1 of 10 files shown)


tsebastiani (approved):
LGTM

run_kraken.py
```diff
@@ -49,50 +50,77 @@ def main(cfg):
...
         kraken_config = cfg
-        chaos_scenarios = config["kraken"].get("chaos_scenarios", [])
-        publish_running_status = config["kraken"].get("publish_kraken_status", False)
+        chaos_scenarios = get_yaml_item_value(
+            config["kraken"], "chaos_scenarios", []
+        )
+        publish_running_status = get_yaml_item_value(
+            config["kraken"], "publish_kraken_status", False
+        )
         port = config["kraken"].get("port")
```
paigerube14, line 66: getting this error in the CI tests, lets go ahead and also set some sort of default value, maybe -1? @chaitanyaenr thoughts on default value? 

```
  File "/home/runner/work/krkn/krkn/run_kraken.py", line 517, in <module>
    main(options.cfg)
  File "/home/runner/work/krkn/krkn/run_kraken.py", line 165, in main
    if not 0 <= int(port) <= 65535:
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

chaitanyaenr: Yes, we can set it to 8081 -  default in the krkn config.

run_kraken.py
```diff
@@ -49,50 +50,77 @@ def main(cfg):
...
-        publish_running_status = config["kraken"].get("publish_kraken_status", False)
-        port = config["kraken"].get("port")
+        chaos_scenarios = get_yaml_item_value(
+            config["kraken"], "chaos_scenarios", []
+        )
+        publish_running_status = get_yaml_item_value(
+            config["kraken"], "publish_kraken_status", False
+        )
+        port = get_yaml_item_value(config["kraken"], "port", 8081)
         signal_address = config["kraken"].get("signal_address")
```
chaitanyaenr, line 67: let's set the default to 0.0.0.0

chaitanyaenr (approved):
LGTM!

PR #516, opened 2023-10-23
Chaos recommender refactoring

tsebastiani:
Refactoring to better integrate the recommender in the krkn codebase


kraken/chaos_recommender/prometheus.py
```diff
@@ -15,14 +14,6 @@ def convert_data_to_dataframe(data, label):
     return df
 
 
-def get_value_from_data(data, service):
```
tsebastiani, line 19: @mudverma removed this method that was unused and was containing a typo

kraken/chaos_recommender/prometheus.py
```diff
@@ -79,7 +70,7 @@ def fetch_utilization_from_prometheus(prometheus_endpoint, auth_token, namespace
     cpu_limits_result = prometheus.custom_query(cpu_limits_query)
 
 
-    mem_query = 'sum by (pod) (avg_over_time(container_memory_usage_bytes{image!="", namespace="%s"}[%s]))' % (namespace, duration)
+    mem_query = 'sum by (pod) (avg_over_time(container_memory_usage_bytes{image!="", namespace="%s"}[%s]))' % (namespace, scrape_duration)
```
tsebastiani, line 73: @mudverma this method was taking the scrape duration as parameter but was using a constant defined on top of the flle, I've restored the parameter passed by the CLI/config that otherwise would be unused.

kraken/chaos_recommender/prometheus.py
mudverma: Agreed

utils/chaos_recommender/chaos_recommender.py
```diff
@@ -0,0 +1,135 @@
...
+from kubernetes import config as kube_config
+
+
+
+def parse_arguments(parser):
+
+    # command line options
+    parser.add_argument("-c", "--config-file", action="store", help="Config file path")
+    parser.add_argument("-o", "--options", action="store_true", help="Evaluate command line options")
+    parser.add_argument("-a", "--application", action="store", default="", help="Kubernetes application name")
```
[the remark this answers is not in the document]
mudverma: Yes, agreed. These were "for future" options incase, we wanted to support these, including libraries other than Krkn, but should be ok to remove for now.

mudverma (approved):

config/recommender_config.yaml
```diff
@@ -0,0 +1,29 @@
+application: openshift-etcd
```
chaitanyaenr, line 1: Thoughts on keeping the config as part of utils to make it easy for the user to configure and run it?

requirements.txt
```diff
@@ -37,4 +37,5 @@ service_identity
 setuptools==65.5.1
 werkzeug==2.2.3
 wheel
-zope.interface==5.4.0
\ No newline at end of file
+zope.interface==5.4.0
+pandas<2.0.0
```
chaitanyaenr, line 41: Maybe we should keep the dependencies separate in the utils to be able to build a lean container image in Krkn-hub for running the chaos-recommender? Thoughts?

chaitanyaenr (approved):
LGTM!

PR #518, opened 2023-10-25
Deprecate kubernetes method of deploying Krkn

chaitanyaenr:
This will ensure users will use the recommended methods ( standlone or containerized ) of installing and running Krkn.


paigerube14 (approved):
/lgtm

PR #522, opened 2023-10-30
Bump telemetry archive default size to 500MB

chaitanyaenr:
This commit also removes litmus configs as they are not maintained.


tsebastiani (approved):
/LGTM

PR #536, opened 2023-12-15
Rename Krkn to KrknChaos

chaitanyaenr:
This change will help reflect the use case of the tool more evidently.


paigerube14:
Will we need to rename the whole project/org as well? Should we do that first?

paigerube14 (approved):
/lgtm

PR #537, opened 2024-01-02
Deployer -> Deployers

paigerube14:
Arcaflow scenarios of deployer to deployers

Still need to test 

Using documentation updates from [here](https://github.com/arcalot/arcaflow-engine/releases/tag/v0.9.0)


jaredoconnell:
`deployer` isn't just replaced with `deployers`. You also need to add the deployer type. In the case of Kubernetes, that's image. So it should be something like:
```
deployers:
  image:
    connection: {}
    type: kubernetes
```

jaredoconnell:
I missed one other thing. We made type less ambiguous by renaming it to `deployer_name`. This was a one-time change. I don't expect major renames in the near future.

jaredoconnell (approved):
Untested, but all changes look good.

chaitanyaenr (approved):
LGTM!

PR #541, opened 2024-01-04
Focus on Kubernetes in the chaos testing guide

chaitanyaenr:


docs/index.md
```diff
@@ -211,7 +211,7 @@ Let us take a look at few recommendations on how and where to run the chaos test
 #### Chaos testing in Practice
 
 ##### OpenShift organization
```
paigerube14, line 213: Should we leave this openshift refernce since its focusing on openshift or move the whole section? 

Is tekton a kubernetes ci, or specific to openshift?

paigerube14:
Theres a missed openshift reference in the workflow section

docs/index.md
chaitanyaenr: I think we should leave it as it's more of a story around how the chaos testing is being practiced in the OpenShift organization. Thoughts?

Tekton can run both on Kubernetes and OpenShift.

paigerube14: agree, think on line 214 we should keep the OpenShift reference then

chaitanyaenr:
> Theres a missed openshift reference in the workflow section

Nice catch! Fixed it, thanks.

docs/index.md
chaitanyaenr: Updated, thanks.

paigerube14 (approved):
/lgtm

PR #544, opened 2024-01-10
Fixes prometheus url check on Kubernetes

tsebastiani:
If the prometheus url is not set in kubernetes the prometheus client fails the initialization and the scenario execution fails.


chaitanyaenr (approved):
LGTM!

PR #550, opened 2024-01-12
Adding push to both red hat and krkn chaos quay

paigerube14:
Want redhat-chaos to stay up to date with current krkn until we have all steps in place


chaitanyaenr (approved):
LGTM!

Issue #511, opened 2023-10-16
[Chaos-recommender] Output recommendations in json


chaitanyaenr:
Provide an option to output recommendations in json format to make it easy for the user/tools to consume.

PR #567, opened 2024-02-08
Add json output file option for recommender (#511)

jtydlack:


chaitanyaenr:
Fixes #567

utils/chaos_recommender/chaos_recommender.py
```diff
@@ -54,16 +60,47 @@ def read_configuration(config_file_path):
...
+                "cpu_limit_query": queries[1],
+                "memory_query": queries[2],
+                "memory_limit_query": queries[3]
+            }
+        ],
+        "profiling": analysis_data[0],
+        "heatmap_analysis": analysis_data[1],
+        "recommendations": analysis_data[2]
+    }
+    file = f"utils/chaos_recommender/recommender_output/{file}"
```
chaitanyaenr, line 99: @jtydlack We might want to append a timestamp to the file name to allow storing data from multiple runs. Thoughts?

utils/chaos_recommender/chaos_recommender.py
```diff
@@ -110,8 +149,15 @@ def main():
...
-    file_path = prometheus.fetch_utilization_from_prometheus(prometheus_endpoint, auth_token, namespace, scrape_duration)
-    analysis(file_path, chaos_tests)
+    file_path, queries = prometheus.fetch_utilization_from_prometheus(prometheus_endpoint, auth_token, namespace, scrape_duration)
+    analysis_data = analysis(file_path, chaos_tests, json_output)
+
+    if json_output is True:
+        time_str = time.strftime("%d%m%y_%H%M%S", time.localtime())
+        recommendation_json = f"recommender_{namespace}_{time_str}.json"
+        make_json_output(recommendation_json, namespace, kubeconfig, prometheus_endpoint, chaos_tests, scrape_duration, queries, analysis_data)
+        logging.info(f"Recommendation output is in file {recommendation_json} in chaos recommender output folder.")
```
chaitanyaenr, line 159: We might want to also print the json to stdout for user to refer. Thoughts?

utils/chaos_recommender/chaos_recommender.py
jtydlack: There is already a timestamp but it is created on different spot so I will move it to this function so it is all at one place

utils/chaos_recommender/chaos_recommender.py
jtydlack: All the information is printed in the terminal when the recommender runs, so I am either not sure if it's unnecessary or if I just do not understand what you mean

chaitanyaenr: Maybe we should consider printing just the json vs text when json_output flag is set. Thoughts?

mudverma (approved):

config/recommender_config.yaml
```diff
@@ -1,4 +1,4 @@
-application: openshift-etcd
+application: opeshift-etcd
```
chaitanyaenr, line 1: s/opeshift/openshift

chaitanyaenr (approved):
LGTM!

PR #577, opened 2024-02-20
Critical alerts collection and upload

tsebastiani:


run_kraken.py
```diff
@@ -382,14 +379,16 @@ def main(cfg):
             telemetry_k8s.collect_cluster_metadata(chaos_telemetry)
 
         decoded_chaos_run_telemetry = ChaosRunTelemetry(json.loads(chaos_telemetry.to_json()))
-        logging.info(f"Telemetry data:\n{decoded_chaos_run_telemetry.to_json()}")
+        chaos_output.telemetry = decoded_chaos_run_telemetry
+        logging.info(f"Chaos data:\n{chaos_output.to_json()}")
         telemetry_elastic.upload_data_to_elasticsearch(decoded_chaos_run_telemetry.to_json(), elastic_index)
```
tsebastiani, line 384: @paigerube14 `upload_data_to_elasticsearch`, from its signature, gets a dictionary but the `to_json()` method returns a string, please give it a look!

run_kraken.py
```diff
@@ -348,21 +350,16 @@ def main(cfg):
...
-                            query = r"""ALERTS{severity="critical"}"""
-                            end_time = datetime.datetime.now()
-                            critical_alerts = prometheus.process_query(
-                                query
-                            )
-                            critical_alerts_count = len(critical_alerts)
-                            if critical_alerts_count > 0:
-                                logging.error("Critical alerts are firing: %s", critical_alerts)
-                                logging.error("Please check, exiting")
+                            prometheus_plugin.critical_alerts(prometheus,
```
tsebastiani, line 353: moved the critical alert collection to the prometheus plugin to not pollute too much the `run_kraken.py`

kraken/prometheus/client.py
```diff
@@ -27,4 +30,53 @@ def alerts(prom_cli: KrknPrometheus, start_time, end_time, alert_profile):
...
+        end_time=end_time
+
+    )
+
+    for alert in during_critical_alerts:
+        alert = ChaosRunAlert(alert["metric"]["alertname"], alert["metric"]["alertstate"], alert["metric"]["namespace"], alert["metric"]["severity"])
+        summary.chaos_alerts.append(alert)
+
+
+    post_critical_alerts = prom_cli.process_query(
```
tsebastiani, line 59: @chaitanyaenr I've the gut feeling that, as we discussed yesterday, in this way, the `during_critical_alerts` will be a subset of the `post_critical_alerts`, I'm pretty convinced that here we should use a query with range using the `end_time` of the chaos run as the start_time of the query range.

chaitanyaenr: We can take down etcd pod and compare the during vs post alerts I guess. Thoughts?

run_kraken.py
```diff
@@ -382,14 +381,16 @@ def main(cfg):
...
+        chaos_output.telemetry = decoded_chaos_run_telemetry
+        logging.info(f"Chaos data:\n{chaos_output.to_json()}")
         telemetry_elastic.upload_data_to_elasticsearch(decoded_chaos_run_telemetry.to_json(), elastic_index)
         if config["telemetry"]["enabled"]:
             logging.info(f"telemetry data will be stored on s3 bucket folder: {telemetry_api_url}/download/{telemetry_request_id}")
             logging.info(f"telemetry upload log: {safe_logger.log_file_name}")
             try:
                 telemetry_k8s.send_telemetry(config["telemetry"], telemetry_request_id, chaos_telemetry)
                 telemetry_k8s.put_cluster_events(telemetry_request_id, config["telemetry"], start_time, end_time)
+                telemetry_k8s.put_critical_alerts(telemetry_request_id, config["telemetry"], summary)
```
paigerube14, line 393: I'm not sure if the alert is in the right place in the telemetry. Also should **scenario** and **run_id** be outside of the _critical_alerts_ section?

```
{
....
....
"critical_alerts": {
        "chaos_alerts": [
            {
                "alertname": "NorthboundStale",
                "alertstate": "pending",
                "namespace": "openshift-ovn-kubernetes",
                "severity": "critical"
            }
        ],
        "post_chaos_alerts": [],
        "scenario": "plugin_scenarios",
        "run_id": "6ca05da4-81ac-49fd-8730-0c5005f4515c"
    }
}
```

run_kraken.py
```diff
@@ -348,21 +350,18 @@ def main(cfg):
...
-                                logging.error("Please check, exiting")
+                            prometheus_plugin.critical_alerts(prometheus,
+                                                              summary,
+                                                              run_uuid,
+                                                              scenario_type,
+                                                              start_time,
+                                                              datetime.datetime.now())
+
+                            chaos_output.critical_alerts = summary
+                            if len(summary.post_chaos_alerts) > 0:
```
paigerube14, line 361: below we do a check (~line 443) to exit 1 if there are alerts at the end of the run. Need the variable to be updated based on these changes. It is exiting with success currently even though I had a critical error firing 

```
2024-02-20 19:02:25,975 [INFO] telemetry collection disabled, skipping.
2024-02-20 19:02:25,975 [INFO] Successfully finished running Kraken. UUID for the run: 6ca05da4-81ac-49fd-8730-0c5005f4515c. Report generated at kraken.report. Exiting
```

tsebastiani: @chaitanyaenr told me to exit only if there were `post` scenario critical alerts so I made this change yesterday, if we wanna exit for any alert I can add the check on both of them.

run_kraken.py
tsebastiani: This thing is related to the fact that this output is printed together with the telemetry in the krkn output, but in the telemetry webservice the two files are distinct so I put the run id in the chaos alert to easily reconcile them on a post processing job, so if it doesn't bother too much, I'd keep the run id within the structure.

run_kraken.py
paigerube14: I got that, but I think we still need to check at the end of the run if the post chaos alerts were firing when we break on line 363 to tell krkn the run had the post scenario failures. Just need to change the variable on the check below to either summary.post_chaos_alerts or chaos_output.critical_alerts.post_chaos_alerts

https://github.com/tsebastiani/krkn/blob/bf3a652f895c4ab1c62ed7d024719624c3734d6d/run_kraken.py#L443-L445

run_kraken.py
paigerube14: gotcha, so if you run multiple scenarios you would be able to see it. Makes sense, thanks

run_kraken.py
paigerube14: actually think we need to just set critical_alerts_count in the if statement that this comment is on so that it'll be set if alerts are turned off but set if they are on

kraken/prometheus/client.py
chaitanyaenr: Alert manager is taking some time to reconcile causing the during and post chaos alerts to overlap. It should be fixed if the wait_duration post chaos is long enough.

run_kraken.py
```diff
@@ -439,7 +442,7 @@ def main(cfg):
                 logging.error("Alert profile is not defined")
                 sys.exit(1)
 
-        if critical_alerts_count > 0:
+        if post_critical_alerts > 0:
```
tsebastiani, line 445: @paigerube14

run_kraken.py
```diff
@@ -347,22 +349,21 @@ def main(cfg):
...
-                                logging.error("Please check, exiting")
+                            prometheus_plugin.critical_alerts(prometheus,
+                                                              summary,
+                                                              run_uuid,
+                                                              scenario_type,
+                                                              start_time,
+                                                              datetime.datetime.now())
+
+                            chaos_output.critical_alerts = summary
+                            post_critical_alerts = len(summary.post_chaos_alerts)
```
tsebastiani, line 362: @paigerube14

chaitanyaenr (approved):
Tested and worked as expected:
```
    "critical_alerts": {
        "chaos_alerts": [
            {
                "alertname": "etcdMembersDown",
                "alertstate": "pending",
                "namespace": "openshift-etcd",
                "severity": "critical"
            }
        ],
        "post_chaos_alerts": [
            {
                "alertname": "etcdMembersDown",
                "alertstate": "pending",
                "namespace": "openshift-etcd",
                "severity": "critical"
            }
        ],
        "scenario": "container_scenarios",
        "run_id": "cc33b8f3-cb5a-4f38-b3f0-b7d8175eb94d"
    }
}
2024-02-28 12:33:06,826 [ERROR] Critical alerts are firing, please check; exiting
# echo $?
1
```

run_kraken.py
paigerube14: looks good! thanks for the chnages

paigerube14 (approved):
/lgtm

PR #583, opened 2024-03-01
Checking alert properties presence

tsebastiani:


kraken/prometheus/client.py
```diff
@@ -74,5 +79,5 @@ def critical_alerts(prom_cli: KrknPrometheus,
     if post_critical_alerts_count > 0:
         firing_alerts = True
 
-    if not firing_alerts:
+    if not firing_alerts:-
```
paigerube14, line 82: typo?

tsebastiani: yeah....

tsebastiani: apparently it was already there

tsebastiani: ah , no, my fault!

PR #584, opened 2024-03-01
Checking post run alerts properties presence

tsebastiani:


kraken/prometheus/client.py
```diff
@@ -66,7 +66,11 @@ def critical_alerts(prom_cli: KrknPrometheus,
     )
 
     for alert in post_critical_alerts:
-        alert = ChaosRunAlert(alert["metric"]["alertname"], alert["metric"]["alertstate"], alert["metric"]["namespace"], alert["metric"]["severity"])
+        alertname = alert["metric"]["alertname"] if "alertname" in alert["metric"] else "none"
```
paigerube14, line 69: do we need to add the `if "metric" in alert` again as well?

tsebastiani: You're right, I don't know what's wrong with me today....

paigerube14 (approved):
/lgtm

PR #586, opened 2024-03-04
Add /usr/local/bin to the path for krkn images

chaitanyaenr:
This is needed to ensure oc and kubectl binaries under /usr/local/bin are accessible.


paigerube14 (approved):
/lgtm

PR #590, opened 2024-03-07
Added exit code 2 on post run critical alerts and failed post scenari…

tsebastiani:
…os to avoid funtests to fail


chaitanyaenr (approved):
LGTM

Issue #513, opened 2023-10-16
[chaos-recommender] Consider node level resource usage for app with no limits


chaitanyaenr:
For identifying CPU/Memory intensive applications, we might want to compare % usage with respect to availability on nodes in cases where there are no limits set.

chaitanyaenr:
cc: @mudverma @sandeephans

chaitanyaenr:
The feature has been added, thanks to @mudverma!

PR #593, opened 2024-03-11
taking in node resources in account

mudverma:
solving #513


chaitanyaenr (commented):
@mudverma the changes are looking good. Can we rebase the PR and also add documentation on how we compare the threshold against node capacity if limits are not set please?

chaitanyaenr (approved):
LGTM

chaitanyaenr:
@mudverma looks like there's a conflict, we might want to rebase with the latest changes. Thoughts?

PR #595, opened 2024-03-12
fix document's nit

LiangquanLi930:


chaitanyaenr (approved):
Thanks for the fix @LiangquanLi930! LGTM!

PR #602, opened 2024-03-21
Copy oc and kubectl clients to additional paths

chaitanyaenr:
This will make sure oc and kubectl clients are accessible for users with both /usr/bin and /usr/local/bin paths set on the host.


tsebastiani (approved):
/lgtm

paigerube14 (approved):

PR #603, opened 2024-03-22
Fix: Reslove ingress network chaos plugin issue

yogananth-subramanian:
Added network_chaos to plugin step and job wait time to be based on the test duration and set the default wait_time to 30s


chaitanyaenr (approved):
LGTM! Tested and it works as expected!

PR #615, opened 2024-04-29
Have a config in the chaos_recommender dir

chaitanyaenr:
This will make it easy for the users to find, configure and run it.


tsebastiani (approved):
LGTM

PR #616, opened 2024-04-30
Add instructions to copy the python package file to docker dir

chaitanyaenr:
Docker image build fails without this.


tsebastiani (approved):
LGTM

PR #636, opened 2024-05-29
support for node limits

mudverma:
added support for node limits when pod limits are not defined

### DIFF
--- a/kraken/chaos_recommender/prometheus.py
+++ b/kraken/chaos_recommender/prometheus.py
@@ -17,16 +17,41 @@ def convert_data_to_dataframe(data, label):
 
 
 def convert_data(data, service):
-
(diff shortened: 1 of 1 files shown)


kraken/chaos_recommender/prometheus.py
```diff
@@ -17,16 +17,41 @@ def convert_data_to_dataframe(data, label):
...
+    node_name = result[0]['metric']['node']
+
+    for item in node_data:
+        if item['metric']['node'] == node_name:
+            return item['value'][1]
+
+    return '1000000000'
+
+
+def save_utilization_to_file(utilisation, filename, prometheus):
```
chaitanyaenr, line 53: s/utilisation/utilization

PR #643, opened 2024-06-06
Use tmp instead of root dir

chaitanyaenr:
This will help with the goal towards avoiding having to run privileged containers where it's not needed.


chaitanyaenr:
Working on making the changes on the krkn-hub side - we need to hold on merging this PR until krkn-hub changes are ready to avoid any issues. Thanks.

tsebastiani (commented):
Isn't better to add a new user and use his /home folder as workdir?

PR #651, opened 2024-06-20
Add pre chaos alert profile

jtydlack:


jtydlack:
Because these queries needs to be managed differently than standard alert queries and will be modified during the run there was created new alert profile. Those queries are not universal yet. Until I figure out how to load and process queries I was not focusing on comparison of pre and post chaos results

run_kraken.py
```diff
@@ -94,6 +94,9 @@ def main(cfg):
             config["performance_monitoring"], "enable_alerts", False
         )
         alert_profile = config["performance_monitoring"].get("alert_profile")
+        pre_chaos_alerts = config["performance_monitoring"].get(
```
chaitanyaenr, line 97: We need to evaluate the rate of increase between start and end time of the chaos scenario instead of pre chaos. Adding the alerts to https://github.com/krkn-chaos/krkn/blob/main/config/alerts_openshift.yaml will automatically get evaluated between start and end time of chaos.

config/pre_chaos_alerts.yaml
```diff
@@ -0,0 +1,35 @@
+# etcd
+
+- expr: sum (rate (container_cpu_usage_seconds_total{namespace='kube-system'}[10m])) by (pod) *1000
```
chaitanyaenr, line 3: We need to compare the output/evaluate against a threshold for example 10%

chaitanyaenr (commented):
Nice start @juliemathew. Added comments around the improvements. Thanks.

PR #652, opened 2024-06-20
Chaos AI updates

sandeephans:
- Check if cluster is accessible before starting chaos
- Allow multiple namespaces
- Automatically find all pods-labels, if not given by user


utils/chaos_ai/docker/requirements.txt
```diff
@@ -4,3 +4,4 @@ requests
 Flask==2.2.5
 Werkzeug==3.0.3
 flasgger==0.9.5
+kubernetes
```
chaitanyaenr, line 7: We might want to set the working version of kubernetes to make sure it doesn't break because of any changes in the upcoming versions. Thoughts?

utils/chaos_ai/docker/swagger_api.py
```diff
@@ -54,6 +55,8 @@ def startchaos(self, kubeconfigfile, file_id, params):
                 params['iterations'] = config_params['iterations']
                 params['maxfaults'] = config_params['maxfaults']
         # faults = [f + ':' + p for f in params['faults'].split(',') for p in params['podlabels'].split(',')]
+        if params['podlabels'] is None or params['podlabels'] == '':
```
chaitanyaenr, line 58: Hi @sandeephans pod labels are not being automatically picked up when not defined. Thoughts?
```
  File "/usr/local/lib/python3.9/threading.py", line 980, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.9/threading.py", line 917, in run
    self._target(*self._args, **self._kwargs)
  File "/app/swagger_api.py", line 58, in startchaos
    if params['podlabels'] is None or params['podlabels'] == '':
KeyError: 'podlabels'
```

PR #656, opened 2024-07-01
Kill pod arca plugin update adaptation

tsebastiani:


requirements.txt
```diff
@@ -15,27 +15,28 @@ google-api-python-client==2.116.0
...
+krkn-lib==2.1.4
 lxml==5.1.0
-kubernetes==26.1.0
+kubernetes==28.1.0
 oauth2client==4.1.3
 pandas==2.2.0
 openshift-client==1.0.21
 paramiko==3.4.0
 podman-compose==1.0.6
+podman-compose==1.0.6
```
paigerube14, line 26: duplicate line here

requirements.txt
```diff
@@ -15,27 +15,28 @@ google-api-python-client==2.116.0
...
 service_identity==24.1.0
-PyYAML==6.0
+PyYAML==6.0.1
 setuptools==65.5.1
 werkzeug==3.0.3
 wheel==0.42.0
 zope.interface==5.4.0
 
-git+https://github.com/krkn-chaos/arcaflow-plugin-kill-pod.git
+git+https://github.com/krkn-chaos/arcaflow-plugin-kill-pod.git@v0.1.0
```
paigerube14, line 40: I see there is a pip package out there for this, is it not set up to properly update versions? 
https://pypi.org/project/arcaflow-plugin-kill-pod/

chaitanyaenr (approved):

chaitanyaenr:
Can be merged post changes suggested by @paigerube14. Thanks.

requirements.txt
tsebastiani: Not sure if reflects the current build. I'm afraid that comes from when the plugin was in charge of the arca team. AFAIK the action is broken due to some arca related steps that are broken that's why we're fetching it from GitHub. I could fix the action eventually, but for the moment I'd stick to this workaround, thoughts?

paigerube14: ok! that sounds good to me for now! dont want to block based on this

chaitanyaenr (approved):
LGTM

PR #660, opened 2024-07-03
moves the krkn-hub build from push on main to tag

tsebastiani:
- solves the Dockerfile manual update on releases
- some enhancements to the Dockerfile to slim down the image a bit


chaitanyaenr (approved):
LGTM

PR #681, opened 2024-08-01
Check cluster access

yogananth-subramanian:
Checks if the user has cluster level access, if not skip collecting cluster metadata and krkn node details.


yogananth-subramanian:
Will refactor and create a  new PR

PR #684, opened 2024-08-07
Adds the startup option to produce prow junit XML output for sippy integration

tsebastiani:
Adds three new options to the krkn startup:

```
  --junit-testcase=JUNIT_TESTCASE
                        junit test case description
  --junit-testcase-path=JUNIT_TESTCASE_PATH
                        junit test case path
  --junit-testcase-version=JUNIT_TESTCASE_VERSION
                        junit test case version
```

if the first two are valorized a file containing the junit tescase will be placed in the specified path ( `${ARTIFACT_DIR}` in prow)


run_kraken.py
```diff
@@ -491,19 +502,102 @@ def main(cfg):
...
     if options.cfg is None:
         logging.error("Please check if you have passed the config")
-        sys.exit(1)
+        option_error = True
+
+    if option_error:
+        retval = 1
     else:
-        main(options.cfg)
+        retval = main(options.cfg)
```
tsebastiani, line 584: We should improve the error handling at all levels in order to centralize the return values in this place. Mainly we should remove all the `sys.exit(x)` in the codebase and replace them with exception trying to catch them in a single place and returning the correct exit status from in the main function, this might become a new issue.

requirements.txt
```diff
@@ -15,7 +15,7 @@ google-api-python-client==2.116.0
 ibm_cloud_sdk_core==3.18.0
 ibm_vpc==0.20.0
 jinja2==3.1.4
-krkn-lib==2.1.7
+krkn-lib@git+https://github.com/krkn-chaos/krkn-lib.git@sippy_integration
```
paigerube14, line 18: should we go ahead and create a krkn-lib tag for this branch since it got merged along with the version changes

chaitanyaenr: +1

tsebastiani: already done 2.1.8

chaitanyaenr (approved):
LGTM!

run_kraken.py
chaitanyaenr: Let's create a issue to track it @tsebastiani

paigerube14 (approved):
lgtm

PR #685, opened 2024-08-09
Deprecate running krkn as kubernetes app

chaitanyaenr:
This commit removes the instructions on running krkn as kubernetes deployment as it is not supported/maintained and also not recommended.


paigerube14 (approved):
/lgtm

PR #686, opened 2024-08-11
Add duration parameter for node scenarios

chaitanyaenr:
This option is enabled only for node_stop_start scenario where user will want to 
stop the node for certain duration to understand the impact before starting the 
node back on. 

This commit also bumps the timeout for the scenario to 360 seconds from 120 seconds 
to make sure there's enough time for the node to get to Ready state from the Kubernetes 
side after the node is started on the infra side.

### DIFF
--- a/kraken/node_actions/abstract_node_scenarios.py
+++ b/kraken/node_actions/abstract_node_scenarios.py
@@ -1,5 +1,6 @@
 import sys
 import logging
+import time
 import kraken.invoke.command as runcommand
 import kraken.node_actions.common_node_functions as nodeaction
 from krkn_lib.k8s import KrknKubernetes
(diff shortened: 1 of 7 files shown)


kraken/node_actions/abstract_node_scenarios.py
```diff
@@ -18,9 +18,10 @@ def node_stop_scenario(self, instance_kill_count, node, timeout):
         pass
 
     # Node scenario to stop and then start the node
-    def node_stop_start_scenario(self, instance_kill_count, node, timeout):
+    def node_stop_start_scenario(self, instance_kill_count, node, timeout, duration):
         logging.info("Starting node_stop_start_scenario injection")
         self.node_stop_scenario(instance_kill_count, node, timeout)
+        logging.info("Waiting for %s seconds before starting the node" % (duration))
```
paigerube14, line 24: I think you want some sort of time.sleep(duration) here right? You are just logging the duration but not actually sleeping

chaitanyaenr: Right! Done.

PR #699, opened 2024-09-28
OCP Chaos Arcaflow Workflow

mfleader:
Adds an Arcaflow workflow to target OpenShift clusters for chaotic experiments!

- [x] subworkflow for cpu hog scenario
- [x] subworkflow for kubeburner load scenario
- [x] subworkflow for kill pod scenario
- [x] main workflow to orchestrate scenarios


scenarios/arcaflow/ocp-chaos/README.md
```diff
@@ -0,0 +1,304 @@
+# OpenShift Shenanigans
```
chaitanyaenr, line 1: @mfleader https://github.com/krkn-chaos/krkn/tree/main/utils might be a good place to host the parallelization effort until it's integrated in krkn after which we need to add a pointer ro https://github.com/krkn-chaos/krkn/blob/main/docs/getting_started.md. Thoughts?

chaitanyaenr (approved):
Ran few tests with different variations to test the toggle feature and it works as expected! LGTM other than moving the workflow to utils!

PR #707, opened 2024-10-04
Do not fail GCP auth when GOOGLE_APPLICATION_CREDENTIALS is not found

pablomh:
Auth could have been set up using the Google Cloud CLI:

- `gcloud init`
- `gcloud auth application-default login`

### DIFF
--- a/krkn/scenario_plugins/node_actions/gcp_node_scenarios.py
+++ b/krkn/scenario_plugins/node_actions/gcp_node_scenarios.py
@@ -14,13 +14,15 @@ from krkn_lib.k8s import KrknKubernetes
 
 class GCP:
     def __init__(self):
-        try:
+        if "GOOGLE_APPLICATION_CREDENTIALS" in os.environ:
             gapp_creds = os.getenv("GOOGLE_APPLICATION_CREDENTIALS")
             with open(gapp_creds, "r") as f:
                 f_str = f.read()
                 self.project = json.loads(f_str)["project_id"]
             # self.project = runcommand.invoke("gcloud config get-value project").split("/n")[0].strip()
             logging.info("project " + str(self.project) + "!")
+
+        try:
             credentials = GoogleCredentials.get_application_default()
             self.client = discovery.build(
                 "compute", "v1", credentials=credentials, cache_discovery=False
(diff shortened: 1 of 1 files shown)


krkn/scenario_plugins/node_actions/gcp_node_scenarios.py
```diff
@@ -14,13 +14,15 @@
 
 class GCP:
     def __init__(self):
-        try:
+        if "GOOGLE_APPLICATION_CREDENTIALS" in os.environ:
             gapp_creds = os.getenv("GOOGLE_APPLICATION_CREDENTIALS")
             with open(gapp_creds, "r") as f:
                 f_str = f.read()
                 self.project = json.loads(f_str)["project_id"]
             # self.project = runcommand.invoke("gcloud config get-value project").split("/n")[0].strip()
```
tsebastiani, line 22: Why do we keep this commented line?

pablomh: Feel free to cherry-pick only the feature commit if desired.

paigerube14:
Got this error when not setting the google application credentials. 


Also would be good to add this to [the docs](https://github.com/krkn-chaos/krkn/blob/main/docs/cloud_setup.md#gcp) as an alternative way to configure the cloud for this scenario

```
2024-10-07 13:54:51,605 [INFO] Starting node_stop_start_scenario injection
2024-10-07 13:54:51,605 [INFO] stop scenario
2024-10-07 13:54:51,605 [INFO] Starting node_stop_scenario injection
2024-10-07 13:54:51,606 [ERROR] Failed to stop node instance. Encountered following exception: 'GCP' object has no attribute 'project'. Test Failed
```

chaitanyaenr:
> Got this error when not setting the google application credentials.

cc: @pablomh

pablomh:
I've changed the status to DRAFT while I figure out how to avoid @paigerube14 problem. The reason I didn't spot it is because I've been testing this code using the `google.auth` Python API which sets `project` using:

```
credentials, project = google.auth.default()
```

pablomh:
Will close in favor of https://github.com/krkn-chaos/krkn/pull/723.

PR #716, opened 2024-10-09
Added fallback on dd if fallocate is not in the $PATH

tsebastiani:
Test deployment based on an image that does not contain fallocate

```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: pvc-test
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
---
apiVersion: v1
kind: Pod
metadata:
  name: pod-with-pvc
spec:
  containers:
    - name: null-container
      image: quay.io/rh_ee_tsebasti/null-container:minimal
      volumeMounts:
        - mountPath: /data
          name: storage-volume
  volumes:
    - name: storage-volume
      persistentVolumeClaim:
        claimName: pvc-test
```


chaitanyaenr (approved):
LGTM! Tested and works as expected!

PR #719, opened 2024-10-16
Pvc scenario block size parameter

tsebastiani:
Added block size as a parameter for dd failover to increase performances


krkn/scenario_plugins/pvc/pvc_scenario_plugin.py
```diff
@@ -241,45 +246,6 @@ def run(
...
-
-            start_time = int(time.time())
-            # Create temp file in the PVC
-            full_path = "%s/%s" % (str(mount_path), str(file_name))
-            command = "fallocate -l $((%s*1024)) %s" % (
-                str(file_size_kb),
-                str(full_path),
-            )
-            logging.debug("Create temp file in the PVC command:\n %s" % command)
-            lib_telemetry.get_lib_kubernetes().exec_cmd_in_pod(
```
paigerube14, line 271: I think we still need this part to actually create the file of the specified size

chaitanyaenr (approved):
LGTM

PR #746, opened 2025-01-13
Convert thresholds to float

chaitanyaenr:
This is needed to avoid issues due to comparing two different data types: TypeError: Invalid comparison between dtype=float64 and str. This commit also avoids setting defaults for the thresholds to make it mandatory for the users to define them as it plays a key role in determining the outliers.


paigerube14 (approved):
/lgtm

PR #747, opened 2025-01-14
Node timing

paigerube14:
This is adding in the ability to add AffectedNode timing to a list that we can track in the telemetry and other output. 

This tracks the amount of time that is taken for the cloud provider to stop/start the node, and the amount of time after the node from the cloud side is stopped/started what is the time the node is in not ready/ready state. 

Needs to go in after https://github.com/krkn-chaos/krkn-lib/pull/143

Any suggestions on how to make this not touch as many files?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,7 +16,7 @@ google-cloud-compute==1.22.0
 ibm_cloud_sdk_core==3.18.0
 ibm_vpc==0.20.0
 jinja2==3.1.5
-krkn-lib==4.0.4
+#krkn-lib==4.0.4
 lxml==5.1.0
 kubernetes==28.1.0
 numpy==1.26.4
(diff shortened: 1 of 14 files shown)


chaitanyaenr (commented):
Tested different combinations and the reporting is working as expected. Working node scenario configs: 
 - one action - node_stop_start_scenario targeting one instance count
 - one action - node_stop_start_scenario targeting multiple instance count

Scenario with multiple actions is failing to report the metrics due to a bug in the node-scenarios code base outside this PR: https://github.com/krkn-chaos/krkn/issues/749

requirements.txt
```diff
@@ -36,6 +36,7 @@ werkzeug==3.0.6
 wheel==0.42.0
 zope.interface==5.4.0
 
+git+https://github.com/krkn-chaos/krkn-lib.git@node_timing
```
chaitanyaenr, line 39: Switch to latest krkn-lib release

chaitanyaenr (approved):
PR is good for merge after the krkn-lib version is bumped to include the enhancements needed.

PR #752, opened 2025-01-27
Adding collecting metrics

paigerube14:
Removing redundant true/false for collecting metrics. I think that if we define that we want to enable metric collection and enable elastic search and give a index, we should go ahead and push the metrics. 

I was not able to fully get the push metrics to work though. Working on adding in changes into krkn-lib

reliant on krkn-lib changes: https://github.com/krkn-chaos/krkn-lib/pull/147

### DIFF
--- a/krkn/prometheus/client.py
+++ b/krkn/prometheus/client.py
@@ -2,10 +2,11 @@ from __future__ import annotations
 
 import datetime
 import os.path
+import math
 from typing import Optional, List, Dict, Any
 
-import urllib3
 import logging
+import urllib3
 import sys
 
 import yaml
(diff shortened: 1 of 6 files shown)


krkn/prometheus/client.py
```diff
@@ -26,16 +26,17 @@ def alerts(
     end_time,
     alert_profile,
     elastic_alerts_index,
+    logging
 ):
 
     if alert_profile is None or os.path.exists(alert_profile) is False:
-        logging.error(f"{alert_profile} alert profile does not exist")
+        print(f"{alert_profile} alert profile does not exist")
```
tsebastiani: Just out curiosity, why are we replacing logging with plain `print`?

tsebastiani (commented):
/lgtm

tsebastiani (approved):
/lgtm

PR #753, opened 2025-01-28
Add auto detecting distribution

jtydlack:
Used is_openshift function from krkn lib

Remove distribution parameter from config


chaitanyaenr:
Nice work @jtydlack!

paigerube14 (approved):
/lgtm

PR #754, opened 2025-02-03
Adding node id to affected node

paigerube14:
Need to add affected node node id to the details. In the cluster shut down scenario we only have knowledge of the instance id that is being stopped not the node name and we don't have a way to easily get it during the run 

Correlating PR in krkn-lib: https://github.com/krkn-chaos/krkn-lib/pull/146

See error below:
https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-redhat-chaos-prow-scripts-main-4.18-nightly-krkn-hub-node-tests/1886263731898290176


chaitanyaenr (approved):
LGTM

paigerube14:
We need to create a krkn-lib release to use with this code

PR #761, opened 2025-02-23
initial version of health checks

kattameghana:


chaitanyaenr:
Nice work @kattameghana!

config/config.yaml
```diff
@@ -111,7 +111,8 @@ telemetry:
     oc_cli_path: /usr/bin/oc                                # optional, if not specified will be search in $PATH
     events_backup: True                                     # enables/disables cluster events collection
 
-
-
-
-
+health_checks:
```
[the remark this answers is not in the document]
kattameghana: Made the respective changes

run_kraken.py
```diff
@@ -362,12 +373,16 @@ def main(cfg) -> int:
...
             iteration += 1
+            health_checker.current_iterations += 1
 
         # telemetry
         # in order to print decoded telemetry data even if telemetry collection
         # is disabled, it's necessary to serialize the ChaosRunTelemetry object
         # to json, and recreate a new object from it.
         end_time = int(time.time())
+        health_check_worker.join()
+        health_check_telemetry = health_check_telemetry_queue.get()
```
tsebastiani, line 384: Tests are failing because the queue is empty and the `get()` method will wait forever. Please use `get_nowait()` instead catching `Empty` exception.

krkn/utils/HealthChecker.py
```diff
@@ -0,0 +1,87 @@
...
+                            start_timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+                            health_check_tracker[config["url"]] = {
+                                "status_code": response["status_code"],
+                                "start_timestamp": start_timestamp
+                            }
+                            if response_tracker[config["url"]] != False: response_tracker[config["url"]] = False
+                            if config["exit_on_failure"] and config["exit_on_failure"] == True and self.ret_value==0: self.ret_value = 2
+                    else:
+                        if config["url"] in health_check_tracker:
+                            end_timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
```
paigerube14, line 52: with the changes to the corresponding krkn-lib pr, lets leave this as a datetime object

kattameghana: Implemented changes to pass `start_timestamp` and `end_timestamp` as strings for JSON serialization. Added logic in `krkn-lib` to convert the string back to a `datetime` object before pushing telemetry data to Elasticsearch.

chaitanyaenr (approved):
LGTM!

kattameghana:
The health check response follows the below format.

Success case:
"health_checks" : [
{
"url" : "http://example.com",
"status" : true,
"status_code" : 200,
"start_timestamp" : "2025-03-11T16:56:31.407462",
"end_timestamp" : "2025-03-11T16:59:56.903850",
"duration" : 205.496388
}
]

Failure case:
"health_checks" : [
{
"url" : "http://example.com",
"status" : false,
"status_code" : "503",
"start_timestamp" : "2025-03-11T17:56:07.677704",
"end_timestamp" : "2025-03-11T17:56:12.656610",
"duration" : 4.978906
}
]

When the health check configuration is empty, no data will be pushed to Elasticsearch. the following data will be displayed in telemetry.
"health_checks": null

paigerube14:
@kattameghana are you able to also add these parameters to krkn-hub config https://github.com/krkn-chaos/krkn-hub/blob/main/config.yaml.template

PR #765, opened 2025-03-04
moving ibm node to non native

paigerube14:


paigerube14:
@chaitanyaenr @tsebastiani PTAL, these updates are blocking from properly adding power outages to ibm cloud in the Prow ci

chaitanyaenr (approved):
LGTM! We might want to add pointers to the example config in the docs: https://github.com/krkn-chaos/krkn/blob/main/docs/node_scenarios.md#ibmcloud.

PR #771, opened 2025-03-11
Adding cluster events into file

paigerube14:
Adding cluster events to print to file instead of in telemetry. Will want to also take out cluster event key from scenario_telemetry in krkn-lib at some point


tsebastiani (commented):
/lgtm

chaitanyaenr (approved):

PR #779, opened 2025-03-20
Smaller metrics

paigerube14:
There were a lot of extra metrics we are collecting and not needed. Triming them down a lot

### DIFF
--- a/config/metrics-aggregated.yaml
+++ b/config/metrics-aggregated.yaml
@@ -1,133 +1,126 @@
 metrics:
 # API server
-  - query: histogram_quantile(0.99, sum(rate(apiserver_request_duration_seconds_bucket{apiserver="kube-apiserver", verb!~"WATCH", subresource!="log"}[2m])) by (verb,resource,subresource,instance,le)) > 0
-    metricName: API99thLatency
-
   - query: sum(irate(apiserver_request_total{apiserver="kube-apiserver",verb!="WATCH",subresource!="log"}[2m])) by (verb,instance,resource,code) > 0
     metricName: APIRequestRate
+    instant: True
 
   - query: sum(apiserver_current_inflight_requests{}) by (request_kind) > 0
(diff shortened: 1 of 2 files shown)


config/metrics-aggregated.yaml
```diff
@@ -1,217 +1,119 @@
...
-  - query: histogram_quantile(0.99, sum(rate(apiserver_request_duration_seconds_bucket{apiserver="kube-apiserver", verb!~"WATCH", subresource!="log"}[2m])) by (verb,resource,subresource,instance,le)) > 0
-    metricName: API99thLatency
-
   - query: sum(irate(apiserver_request_total{apiserver="kube-apiserver",verb!="WATCH",subresource!="log"}[2m])) by (verb,instance,resource,code) > 0
     metricName: APIRequestRate
+    instant: True
 
   - query: sum(apiserver_current_inflight_requests{}) by (request_kind) > 0
     metricName: APIInflightRequests
+    instant: True
```
chaitanyaenr, line 9: Should we capture peak and 95th percentile for the metrics to be able to identify spikes while still capturing less number of samples? Same for master node cpu/mem. Thoughts?

paigerube14: I'll add in extra queries for those, but I don't think we should auto put that into the code to generate if instant is specified. Thoughts?

chaitanyaenr: Agreed @paigerube14. It's needed only for the metrics where it makes sense.

paigerube14: added cpu/memory maxes for workers and master nodes. any others that are important? I couldn't figure out how to get the one you commented on here

chaitanyaenr (approved):

chaitanyaenr (approved):
LGTM

tsebastiani (commented):
/lgtm

PR #785, opened 2025-04-07
RBAC for running privileged vs non-privileged scenarios

yogananth-subramanian:
Document privileged and non-privileged RBAC  permissions needed for each of the scenarios for the users to use and run scenarios based on their access level.


yogananth-subramanian:
Moved the changes to https://github.com/krkn-chaos/website/pull/37

Issue #762, opened 2025-02-28
Add support for triggering kubevirt VM outages


chaitanyaenr:
Users can benefit from being able to trigger VM outages ( when kubevirt/CNV is enabled on the cluster ) to ensure availability and also understand/improve recovery timing to avoid extended downtime.

We can leverage https://github.com/kubevirt/client-python to talk to the kubevirt APIs and trigger the outages.

abhinavs1920:
Heyyy!!  @chaitanyaenr @paigerube14 @tsebastiani 

I plan to add a new kubevirt_vm_outage scenario plugin using the kubevirt/client-python SDK to delete a VirtualMachineInstance and let Kubernetes handle recovery. 

I’ll pin the SDK in requirements.txt, register the plugin in the loader, and cover both unit tests (with mocks) and real-cluster integration tests. Are there any coding conventions(currently I am taking reference from existing plugins) or telemetry/logging requirements I should be aware of before diving in?

I’ll update the sample YAML, add docs, and share a PR for review soon.

Thanks!!!

paigerube14:
@abhinavs1920 awesome! thanks so much for your interest in contributing. This is a great PR to get started on 

All your steps/thinking sound good to me! Just a note that the docs should all live in [this repo](https://github.com/krkn-chaos/website) (we're in the process of taking them all out of here into one common place) 

It would be great if you could add actions 1-6 [documented here](https://krkn-chaos.dev/docs/scenarios/node-scenarios/) for the kubevirt configuration. This is what we support on all the other clouds that it would be awesome to stay consistent. If you want to take a look at any of them. Also can start with just one to get started as well

abhinavs1920:
> [@abhinavs1920](https://github.com/abhinavs1920) awesome! thanks so much for your interest in contributing. This is a great PR to get started on
> 
> All your steps/thinking sound good to me! Just a note that the docs should all live in [this repo](https://github.com/krkn-chaos/website) (we're in the process of taking them all out of here into one common place)
> 

Sure, will add the docs accordingly.

> It would be great if you could add actions 1-6 [documented here](https://krkn-chaos.dev/docs/scenarios/node-scenarios/) for the kubevirt configuration. This is what we support on all the other clouds that it would be awesome to stay consistent. If you want to take a look at any of them. Also can start with just one to get started as well

Taking a look on them.

Thanks!!

PR #816, opened 2025-05-16
Add support for triggering kubevirt VM outages

abhinavs1920:
# Description

This PR introduces a new scenario plugin for simulating KubeVirt VM outages. Currently this PR draft PR in which work will be done:

Core thing to be implemented:

- Delete KubeVirt Virtual Machine Instances (VMIs) to simulate VM crashes

## Documentation

Documentation to be added:
- `kubevirt_vm_outage.md in website repo`
- Plugin documentation within the code to be added

https://github.com/krkn-chaos/website/pull/68

## Changes Made

1. Added new scenario plugin:
   - `krkn/scenario_plugins/kubevirt_vm_outage/kubevirt_vm_outage_scenario_plugin.py`

2. Added requirements

### Issue 

This PR will resolve the issue #762


krkn/scenario_plugins/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Import plugin to make it available for the plugin factory
+from krkn.scenario_plugins.kubevirt_vm_outage.kubevirt_vm_outage_scenario_plugin import KubevirtVmOutageScenarioPlugin
```
paigerube14, line 2: instead of importing this here, I think this should be set up similarly to the nod action scenario in: 

https://github.com/krkn-chaos/krkn/blob/main/krkn/scenario_plugins/node_actions/node_actions_scenario_plugin.py

abhinavs1920: Sure, thanks for the feedback will update it as per the existing norms.

abhinavs1920:
@paigerube14
I believe the initial implementation of the KubeVirt VM outage scenario is in a good state. I am currently working on writing the unit tests. Should I raise a separate PR for merging the unit tests?

Also, I think this implementation can be improved further, so I would appreciate your honest feedback. :)

Thanks.

abhinavs1920:
Hey @paigerube14, @tsebastiani @chaitanyaenr 

I’ve updated the requirements.txt to pin the KubeVirt Python client to a specific commit (git+https://github.com/kubevirt/client-python.git@868766b). This is because we’re facing some version compatibility issues while fetching KubeVirt directly from the Git repo, as mentioned in this KubeVirt PR: [kubevirt/kubevirt#14634](https://github.com/kubevirt/kubevirt/pull/14634). 

Once this issue is sorted out, I’ll update the dependency to use the stable version.

krkn/scenario_plugins/kubevirt_vm_outage/kubevirt_vm_outage_scenario_plugin.py
```diff
@@ -0,0 +1,276 @@
...
+            k8s_config = kubernetes.client.Configuration().get_default_copy()
+            k8s_config.api_key = k8s_client.api_client.configuration.api_key
+            k8s_config.api_key_prefix = k8s_client.api_client.configuration.api_key_prefix
+            k8s_config.host = k8s_client.api_client.configuration.host
+            k8s_config.ssl_ca_cert = k8s_client.api_client.configuration.ssl_ca_cert
+            k8s_config.verify_ssl = k8s_client.api_client.configuration.verify_ssl
+            k8s_config.cert_file = k8s_client.api_client.configuration.cert_file
+            k8s_config.key_file = k8s_client.api_client.configuration.key_file
+            
+            api_client = ApiClient(k8s_config)
```
paigerube14, line 81: api client is already set up using krkn-lib here: https://github.com/krkn-chaos/krkn-lib/blob/63e58983997ff062fe4bd1ccff1fca4ac4c63b69/src/krkn_lib/k8s/krkn_kubernetes.py#L91

Can you use that init instead of creating again here? The kubvirt api can also be added to krkn-lib instead with all get, list, etc functions should live in krkn-lib and called from here

abhinavs1920: Hi, I have made some changes, I have refactored to use KrknKubernetes client for KubeVirt operations instead of a separate client. 

Also I have removed KrknKubevirt class and updated tests. All functionality remains the same, just made it using the existing client now.

Thanks!! :)

krkn/scenario_plugins/kubevirt_vm_outage/kubevirt_vm_outage_scenario_plugin.py
```diff
@@ -0,0 +1,294 @@
...
+        Recover a deleted VMI, either by waiting for auto-recovery or manually recreating it.
+        
+        :param vm_name: Name of the VMI to recover
+        :param namespace: Namespace of the VMI
+        :return: 0 for success, 1 for failure
+        """
+        try:
+            logging.info(f"Attempting to recover VMI {vm_name} in namespace {namespace}")
+            
+            timeout = 300  # seconds
```
yogananth-subramanian: This wait might not be required, since your are already waiting for the "duration"

krkn/scenario_plugins/kubevirt_vm_outage/kubevirt_vm_outage_scenario_plugin.py
```diff
@@ -0,0 +1,294 @@
...
+            return 1
+            
+        except Exception as e:
+            logging.error(f"Error deleting VMI {vm_name}: {e}")
+            log_exception(e)
+            return 1
+
+    def recover(self, vm_name: str, namespace: str) -> int:
+        """
+        Recover a deleted VMI, either by waiting for auto-recovery or manually recreating it.
```
yogananth-subramanian, line 290: we could have an option to disable auto restart, by patching the vm spec, as one of the test parameter.

scenarios/kubevirt/kubevirt-vm-outage.yaml
```diff
@@ -0,0 +1,7 @@
+scenarios:
+  - name: "kubevirt outage test"
+    scenario: kubevirt_vm_outage
```
tsebastiani, line 3: Even if you plan to have multiple `kubevirt_*` sub-scenario , that I understand you might have seen from other scenarios, it's better to do not write code that [you ain't gonna need it](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). This behavior is already mapped in the API itself through the `get_scenario_types()` and you implemented it correctly,

abhinavs1920:
> The dependency installation fails with the following error:
> 
> ` × python setup.py egg_info did not run successfully. │ exit code: 1 ╰─> [16 lines of output] Traceback (most recent call last): File "<string>", line 2, in <module> File "<pip-setuptools-caller>", line 34, in <module> File "/private/var/folders/fw/52twjtjx14l2grt88ryv7djw0000gn/T/pip-req-build-5z03nfw_/setup.py", line 28, in <module> setup( File "/Users/tsebasti/Projects/krkn/venv/lib/python3.9/site-packages/setuptools/__init__.py", line 117, in setup return distutils.core.setup(**attrs) File "/Users/tsebasti/Projects/krkn/venv/lib/python3.9/site-packages/setuptools/_distutils/core.py", line 148, in setup _setup_distribution = dist = klass(attrs) File "/Users/tsebasti/Projects/krkn/venv/lib/python3.9/site-packages/setuptools/dist.py", line 334, in __init__ self.metadata.version = self._normalize_version(self.metadata.version) File "/Users/tsebasti/Projects/krkn/venv/lib/python3.9/site-packages/setuptools/dist.py", line 370, in _normalize_version normalized = str(Version(version)) File "/Users/tsebasti/Projects/krkn/venv/lib/python3.9/site-packages/packaging/version.py", line 202, in __init__ raise InvalidVersion(f"Invalid version: {version!r}") packaging.version.InvalidVersion: Invalid version: 'v1.6.0-beta.0-100-g8e69ed103f'`
> 
> how did you solve this? Looks like there's also an [issue](https://github.com/kubevirt/client-python/issues/60) opened on the client repo

Yes you are right it gives us the error, this is because we’re facing some version compatibility issues while fetching KubeVirt directly from the Git repo, as you mentioned PR: https://github.com/kubevirt/kubevirt/pull/14634. Initially I used the kubevirt python plugin for this feature initially, but soon realized it would be really hard for the to get the plugin manually each time, so the current implementation is only based on the kubernetes client, as it meets all requirements without adding unnecessary complexity. 

The kubevirt-python-client would be more beneficial for complex KubeVirt operations, but for basic VMI management, the standard Kubernetes client is perfectly sufficient I guess, so in my next commit I will remove this dependency from the `requirement.txt`.

krkn/scenario_plugins/kubevirt_vm_outage/kubevirt_vm_outage_scenario_plugin.py
abhinavs1920: Great, I would work on this. It will be really helpful to predict the behavior when the vm doesnt get restarted automatically.

krkn/scenario_plugins/kubevirt_vm_outage/kubevirt_vm_outage_scenario_plugin.py
abhinavs1920: Yes correct, I will remove it.

scenarios/kubevirt/kubevirt-vm-outage.yaml
abhinavs1920: Yep, I have got your point, it makes sense to just provide the things required by the current usage requirements.
I have resolved this in my latest commit.

krkn/scenario_plugins/kubevirt_vm_outage/kubevirt_vm_outage_scenario_plugin.py
```diff
@@ -0,0 +1,335 @@
...
+            return True
+            
+        except ApiException as e:
+            logging.error(f"Failed to patch VM {vm_name}: {e}")
+            return False
+        except Exception as e:
+            logging.error(f"Unexpected error patching VM {vm_name}: {e}")
+            return False
+            
+    def inject(self, vm_name: str, namespace: str, disable_auto_restart: bool = False) -> int:
```
paigerube14, line 215: it might make more sense here to call the function delete_vm instead of inject. inject is too broad and could mean a lot of things

krkn/scenario_plugins/kubevirt_vm_outage/kubevirt_vm_outage_scenario_plugin.py
```diff
@@ -0,0 +1,335 @@
...
+        """
+        Validate that KubeVirt is installed and the specified VM exists.
+        
+        :param vm_name: Name of the VM to validate
+        :param namespace: Namespace of the VM
+        :return: True if environment is valid, False otherwise
+        """
+        try:
+            # Check if KubeVirt CRDs exist
+            crd_list = self.k8s_client.list_custom_resource_definition()
```
paigerube14: doesn't look like this function exists in [krkn-lib](https://github.com/search?q=repo%3Akrkn-chaos%2Fkrkn-lib%20list_custom_resource_definition&type=code)? Do you have a PR to add it? You might want to also pass the kubevirt.io type to it and do the looping in the krkn-lib function itself

paigerube14: think you want to use: https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CustomObjectsApi.md#list_namespaced_custom_object

krkn/scenario_plugins/kubevirt_vm_outage/kubevirt_vm_outage_scenario_plugin.py
```diff
@@ -0,0 +1,335 @@
...
+
+from krkn.scenario_plugins.abstract_scenario_plugin import AbstractScenarioPlugin
+
+
+class KubevirtVmOutageScenarioPlugin(AbstractScenarioPlugin):
+    """
+    A scenario plugin that injects chaos by deleting a KubeVirt Virtual Machine Instance (VMI).
+    This plugin simulates a VM crash or outage scenario and supports automated or manual recovery.
+    """
+
```
paigerube14, line 20: be sure to add a **get_scenario_types** function that returns the name of the scenario plugin 
https://krkn-chaos.dev/docs/contribution-guidelines/scenario_plugin_api/#get_scenario_types

like this format: https://github.com/krkn-chaos/krkn/blob/630dbd805b9540e2445422dcea10dfcdc390a827/krkn/scenario_plugins/application_outage/application_outage_scenario_plugin.py#L92-L93

krkn/scenario_plugins/kubevirt_vm_outage/kubevirt_vm_outage_scenario_plugin.py
```diff
@@ -0,0 +1,335 @@
...
+        :param disable_auto_restart: Whether auto-restart was disabled during injection
+        :return: 0 for success, 1 for failure
+        """
+        try:
+            logging.info(f"Attempting to recover VMI {vm_name} in namespace {namespace}")
+            
+            # Check current state once since we've already waited for the duration
+            vmi = self.get_vmi(vm_name, namespace)
+            
+            if vmi:
```
paigerube14: let's maybe pull this wait for running code into a separate function

krkn/scenario_plugins/kubevirt_vm_outage/kubevirt_vm_outage_scenario_plugin.py
```diff
@@ -0,0 +1,335 @@
...
+            logging.info(f"Captured initial state of VMI: {vm_name}")
+            
+            result = self.inject(vm_name, namespace, disable_auto_restart)
+            if result != 0:
+                return 1
+                
+            logging.info(f"Waiting for {duration} seconds before attempting recovery")
+            time.sleep(duration)
+            
+            result = self.recover(vm_name, namespace)
```
paigerube14: instead of recover here, I think it would be better to just call a wait until running function and then if that fails, add an option in the config if you want to recover. I'm not sure if we'll want to do this everytime

krkn/scenario_plugins/kubevirt_vm_outage/kubevirt_vm_outage_scenario_plugin.py
```diff
@@ -0,0 +1,335 @@
...
+                return 1
+                
+            self.original_vmi = vmi
+            logging.info(f"Captured initial state of VMI: {vm_name}")
+            
+            result = self.inject(vm_name, namespace, disable_auto_restart)
+            if result != 0:
+                return 1
+                
+            logging.info(f"Waiting for {duration} seconds before attempting recovery")
```
paigerube14: I think this wait duration might be overkill, I ran on my system and then vmi came back and was running within 10 seconds. I think instead of duration in the config, we could add a timeout to wait until running

krkn/scenario_plugins/kubevirt_vm_outage/kubevirt_vm_outage_scenario_plugin.py
```diff
@@ -0,0 +1,335 @@
...
+                    return 1
+                else:
+                    logging.error(f"API error during VMI deletion: {e}")
+                    return 1
+            
+            # Wait for the VMI to be deleted
+            timeout = 120  # seconds
+            start_time = time.time()
+            while time.time() - start_time < timeout:
+                if not self.get_vmi(vm_name, namespace):
```
paigerube14: in my testing, the vmi was actually non existent for about 1 second and this code didn't catch it. Can you add in a second "or" check here if the metadata.creationTimestamp of the original vmi doesn't match the current vmi to also see that it was successfully deleted.

krkn/scenario_plugins/kubevirt_vm_outage/kubevirt_vm_outage_scenario_plugin.py
paigerube14: along with that, the abstract plugin couldn't find your file because there's not an init file in the new folder you added. Are you able to add a blank ___init__.py_ file to this folder

krkn/scenario_plugins/kubevirt_vm_outage/kubevirt_vm_outage_scenario_plugin.py
paigerube14: We do have the [custom resource client](https://github.com/krkn-chaos/krkn-lib/blob/7ab3b0588ba8ab2fac4ba3f0c0abe8ae9743b94d/src/krkn_lib/k8s/krkn_kubernetes.py#L87-L89) already configured in krkn-lib, I think you'll want to use that client instead of k8s_client on all of the function calls here

abhinavs1920:
Hi @paigerube14, apologies for the delay, I’ve been a bit tied up over the past couple of weeks. I’ll start working on my PR as soon as I get some time to properly go through the issues.

Thanks!

chaitanyaenr (approved):

chaitanyaenr:
Thanks for the contribution @abhinavs1920! This is a key feature for testing the resilience of kube-virt!

PR #818, opened 2025-05-16
Implementation of Centralized Clean-up feature for krkn

abhinavs1920:
## Description  

Implement cleanup manager for Krkn to handle termination and resource cleanup. This PR adds a centralized mechanism for registering and executing cleanup callbacks when Krkn terminates due to signals, exceptions, or normal exit.

Key changes:

In progress

- Create CleanupManager class for signal and exit handling

To be done

- Integrate cleanup into main execution flow
- Implement artifact collection for diagnostic purpose
- Add BaseScenario with cleanup method for scenario plugins
- Add documentation for cleanup mechanism
- Add tests  for cleanup functionality

## Documentation  

I will be parallelly working on the documentation.

### DIFF
--- a/krkn/core/cleanup.py
+++ b/krkn/core/cleanup.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python
+"""
+Cleanup manager for Krkn. This module provides a centralized mechanism for registering and executing
+cleanup callbacks when Krkn terminates due to signals, exceptions, or normal exit.
+"""
+
+import signal
+import atexit
+import sys
+import os
+import subprocess
+import logging
+import datetime
+from typing import Callable, List, Optional
+
+
+class CleanupManager:
+    """
+    Manages cleanup operations for Krkn.
+    
+    This class handles registering signal handlers, exit handlers, and exception hooks
+    to ensure cleanup functions are called during termination.
+    
+    Implementted capturing of diagnostic information to help with debugging.
+    """
+    
+    def __init__(self, artifacts_dir: Optional[str] = None):
+        """
+        Initialization ofthe cleanup manager.
+        
+        Args:
+            artifacts_dir: Directory where diagnostic artifacts will be stored.
+                           If None, a default timestamped directory will be created.
+        """
+        self._cleanup_funcs = []
+        self._done = False
+        
+        if artifacts_dir is None:
+            run_id = datetime.datetime.now().strftime("%Y%m%d-%H%M%S")
+            self.artifacts_dir = os.path.abspath(os.path.join("artifacts", f"krkn-run-{run_id}"))
+        else:
+            self.artifacts_dir = os.path.abspath(artifacts_dir)
+            
+        os.makedirs(self.artifacts_dir, exist_ok=True)
+        logging.info(f"Artifacts will be stored in: {self.artifacts_dir}")
+        
+        signal.signal(signal.SIGINT, self._signal_handler)
+        signal.signal(signal.SIGTERM, self._signal_handler)
+        
+        atexit.register(self._run_cleanup)
+        
+        sys.excepthook = self._exception_handler
+
+   


abhinavs1920:
@paigerube14 @chaitanyaenr @tsebastiani 
I have some questions regarding this feature:

1. Does a centralized cleanup handler a good idea, or would you prefer cleanup to be handled per-scenario?
2. Should I move the the cleanup class to krkn.utils?
3. Is there any existing internal logic I should avoid duplicating or refactor into this tool/feature?

krkn/core/cleanup.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        Capture diagnostic info about the cluster.
+        
+        This method runs kubectl commands to gather information about pods, nodes, 
+        and events, and saves them to files in the artifacts directory.
+        """
+        logging.info("Capturing diagnostic information...")
+        
+        try:
+            self._run_and_save_command(["kubectl", "get", "pods", "-A", "-o", "wide"], "all-pods.txt")
```
tsebastiani, line 125: we do not use `kubectl` or any other binaries to collect cluster data, we use kubernetes API instead.

tsebastiani (commented):
First of all @abhinavs1920 thank you very much for your contribution and interest around the project.
Even if the overall idea of a centralized error handler might be partially interesting I lean more for an hybrid approach: a per scenario cleanup method for complex cleanup scenarios that in most cases is unimplemented and a fallback centralized for all the common objects defined by a `name` a `namespace` an original yaml object collected before the resource is modified/deleted and an action to perform (`delete` `revert` `recreate` ). The problem that I see in your design is that in your manager there's a lot of complexity for signal handlers and callbacks and no trace of this cleanup strategies. If you see the [API](https://github.com/krkn-chaos/krkn/blob/0dcb901da11c7b3760e6fe0ab2288b222745ac57/krkn/scenario_plugins/abstract_scenario_plugin.py#L77) no exception must escape this block, so the magic should happen all here, the signal handler could be registered in the constructor of the AbstractScenarioPlugin, and could be a private method of the class itself that simply calls `self.cleanup()`in case a ctrl+c or an os interrupt is sent and the same for the exceptions that should be handled in that block. This might represent the basic structure of the handling, then there is all the `per-scenario` logic for tracking the resource creation both in a shared thread safe structure like a queue and in the filesystem in order to be able to recover also from a severe disruption (eg. unexpected shutdown). I'm sorry about that but this is currently the task that we assigned for the LFX internship and must be subject of a wider discussion unless the solution proposed is complete and aligns with the project design.

abhinavs1920:
Thanks a lot @tsebastiani for the detailed feedback and for taking the time to walk through the design concerns.

It makes sense now why handling cleanup within `AbstractScenarioPlugin` is the better approach, especially with the focus on keeping things aligned with the plugin structure and avoiding signal handling or command execution outside of that context. The hybrid model you described, where each scenario has its own cleanup() while still supporting a simple, centralized fallback for shared resources—sounds like a solid direction.

I'll go ahead and refactor this PR accordingly. Planning to move the signal handling and cleanup logic into the plugin base class, drop the external CleanupManager, and make sure everything stays consistent with the current scenario framework. I'll also switch away from using kubectl and stick with the Kubernetes API as suggested.

Thanks again for the guidance, I'll update the PR soon with these changes.

abhinavs1920:
> I'm sorry about that but this is currently the task that we assigned for the LFX internship and must be subject of a wider discussion unless the solution proposed is complete and aligned with the project design.

I just have to share some context: I’ve applied to the upcoming LFX internship as well, and my motivation for working on this cleanup mechanism was both personal interest and to start contributing early to the project in areas which are priority.

I understand that this feature is part of the LFX scope, and I don’t want to create parallel or conflicting work. That said, I’m happy to rework my PR to better align with the architecture you’ve described—mainly shifting cleanup into AbstractScenarioPlugin with scenario-specific logic and a centralized fallback handler. Even if it’s not merged right away, I hope it can serve as a constructive step in the direction you’re envisioning.

If selected, I’d love to continue refining this as part of the internship effort; and if not, I’m still happy to contribute in a way that supports the LFX intern or complements the official workstream.

Thanks again for the guidance—I'll begin updating my code accordingly.

Issue #827, opened 2025-05-28
Prometheus CI deploy keeps failing with startup timeout error


Transcendental-Programmer:
Hi,

I saw everyone running into the ci check fail : one of my PRs (https://github.com/krkn-chaos/krkn-lib/pull/163) fails when installing Prometheus, but then i noticed  @tsebastiani PR https://github.com/krkn-chaos/krkn-lib/pull/160 worked just fine—and neither of us changed the CI setup. 

**What’s going on?**
When the Helm chart tries to install `kube-prometheus-stack`, it spits out:

```
Run # nodePort mapping 30000 -> http://localhost:9090
namespace/monitoring created
Error: INSTALLATION FAILED: 1 error occurred:
	* Prometheus.monitoring.coreos.com "kind-prometheus-kube-prome-prometheus" is invalid: spec.maximumStartupDurationSeconds: Invalid value: 0: spec.maximumStartupDurationSeconds in body should be greater than or equal to 60


Error: Process completed with exit code 1.
```

In short: the CRD now insists that `maximumStartupDurationSeconds` must be at least 60 seconds, but by default it’s 0, so it errors out.

**Why did his PR pass but mine fail?**

* maybe the Helm chart got updated between the two CI runs
* maybe our CI runners updated Kubernetes or the operator CRD
* either way, the default value (0) is now invalid under the new rules

**How to fix**

figured out we need to change code in [`.github/workflows/tests.yml` ](https://github.com/krkn-chaos/krkn/blob/main/.github/workflows/tests.yml) after line 38 

adding 

```
  --set prometheus.prometheusSpec.maximumStartupDurationSeconds=300
```

**References**

https://github.com/prometheus-community/helm-charts/issues/4585

Let me know if I’ve missed anything or i can go ahead and update the workflow. Thanks!

PR #828, opened 2025-05-28
Prometheus ci startup timeout error

Transcendental-Programmer:
## Description

This PR includes the following changes:

1. Ensures the CI workflow fails when any test fails, improving reliability.
2. Sets the correct Git email for automated commits to avoid identity issues in logs.
3. Fixes Prometheus installation by setting the required `maximumStartupDurationSeconds` parameter, preventing startup errors during deployment.

## Documentation

* [ ] **Is documentation needed for this update?**
  (If checked, a documentation PR must be created and merged in the [[website repository](https://github.com/krkn-chaos/website/)](https://github.com/krkn-chaos/website/).)

---


Transcendental-Programmer:
@tsebastiani does this fix the issue in https://github.com/krkn-chaos/krkn/issues/827 and i also want to solve the same issue in other repos if you could review this pr i will proceed for the other repos

paigerube14:
I approved your functional test run but I'm not sure why it's not pulling up in the checks list but it looks like you can only set one "run" in your workflow 

https://github.com/krkn-chaos/krkn/actions/runs/15304867377

Transcendental-Programmer:
> I approved your functional test run but I'm not sure why it's not pulling up in the checks list but it looks like you can only set one "run" in your workflow
> 
> https://github.com/krkn-chaos/krkn/actions/runs/15304867377

yes i fixed that and also squashed my commits to a single one.

.github/workflows/tests.yml
```diff
@@ -36,6 +36,7 @@ jobs:
           --set alertmanager.service.type=NodePort \
           --set prometheus-node-exporter.service.nodePort=32001 \
           --set prometheus-node-exporter.service.type=NodePort
+          --set prometheus.prometheusSpec.maximumStartupDurationSeconds=300
```
priyansh-saxena1, line 39: this line fixes the ci error

Transcendental-Programmer:
> @Transcendental-Programmer thanks for the updates on this PR but something still doesn't look quite right. Can you take another look here. TIA
> 
> ```
> namespace/prometheus-k8s created
> Error: INSTALLATION FAILED: 1 error occurred:
> 	* Prometheus.monitoring.coreos.com "kind-prometheus-kube-prome-prometheus" is invalid: spec.maximumStartupDurationSeconds: Invalid value: 0: spec.maximumStartupDurationSeconds in body should be greater than or equal to 60
> ```

hi @paigerube14 that's the exact error i am trying to fix through the pr, since the current ci file which does not have my changes will be checking the correctness of the ci tests so it is intended to fail.

paigerube14:
@Transcendental-Programmer if you look at [this run](https://github.com/krkn-chaos/krkn/actions/runs/15305953427/job/43059278807#step:5:1) I think it is using your changes but still giving error

Transcendental-Programmer:
Hi @paigerube14 , thanks for pointing that out. I took another look, and it turns out the error was due to a super tiny issue : a missing backslash :) , which is why the command 

```
--set prometheus-node-exporter.service.type=NodePort (\) --> missing backslash 
          --set prometheus.prometheusSpec.maximumStartupDurationSeconds=300
```

was being treated as a separate command rather than being part of the helm install.

PR #833, opened 2025-05-30
Add local metrics persistence when Elasticsearch is disabled

### LINKED ISSUE #829: Be able to save metrics locally
Currently, if a user set enable_metrics to True but does not set enable_elastic to True (False), the metrics are collected and nothing is done with them. If the following set up occurs, I think we should save the metrics to a file at a tmp location (maybe based on the run_uuid to be specific to a run) 

```
performance_monitoring:
....                                  
    enable_alerts: False                                  # Runs the queries specified in the alert profile and displays the info or exits 1 when severity=error
    enable_metrics: True
    alert_profile: config/alerts.yaml                          # Path or URL to alert profile with the prometheus queries
    metrics_profile: config/metrics-report.yaml
    check_critical_alerts: False                          # When enabled will check prometheus for critical alerts firing post chaos
elastic:
    enable_elastic: False
    verify_certs: False
    elastic_url: ""                                         # To track results in elasticsearch, give url to server here; will post telemetry details when url and index not blank
    elastic_port: 32766
    username: "elastic"
    password: "test"
    metrics_index: "krkn-metrics"
    alerts_index: "krkn-alerts"
    telemetry_index: "krkn-telemetry"
```

Code block associated with issue

https://github.com/krkn-chaos/krkn/blob/0dcb901da11c7b3760e6fe0ab2288b222745ac57/krkn/prometheus/client.py#L253-L259

abhinavs1920:
## Description
Adds local metrics persistence when Elasticsearch is disabled. Now when `enable_metrics=true` but `enable_elastic=false`, metrics will be saved to `/tmp/krkn_<run_uuid>.json`.

## Documentation
- [ ] **Is documentation needed for this update?** No

## Testing
- Verified metrics save locally when `enable_metrics=true` and `enable_elastic=false`
- Confirmed existing Elasticsearch integration remains functional

Resolves #829

### DIFF
--- a/krkn/prometheus/client.py
+++ b/krkn/prometheus/client.py
@@ -15,6 +15,7 @@ from krkn_lib.elastic.krkn_elastic import KrknElastic
 from krkn_lib.models.elastic.models import ElasticAlert
 from krkn_lib.models.krkn import ChaosRunAlertSummary, ChaosRunAlert
 from krkn_lib.prometheus.krkn_prometheus import KrknPrometheus
+from krkn.utils.functions import write_json_to_tmp
 
 
 urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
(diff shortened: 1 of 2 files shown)


krkn/prometheus/client.py
```diff
@@ -251,11 +252,22 @@ def metrics(
                         metric[k] = v
                         metric['timestamp'] = str(datetime.datetime.now())
                     metrics_list.append(metric.copy())
-        if elastic:
+        # Save metrics based on elastic availability
+        if elastic is not None and hasattr(elastic, 'upload_metrics_to_elasticsearch'):
```
paigerube14, line 256: not sure on what _upload_metrics_to_elasticsearch_ attribute would be. Instead, we can add in check to see if `elastic_metrics_index` is not none instead

https://github.com/krkn-chaos/krkn/blob/0dcb901da11c7b3760e6fe0ab2288b222745ac57/krkn/prometheus/client.py#L256C42-L256C63

krkn/prometheus/client.py
```diff
@@ -251,11 +252,22 @@ def metrics(
...
+            # Upload to elasticsearch if elastic client is available
             result = elastic.upload_metrics_to_elasticsearch(
                 run_uuid=run_uuid, index=elastic_metrics_index, raw_data=metrics_list
             )
             if result == -1:
                 logging.error("failed to save metrics on ElasticSearch")
+        else:
+            # Otherwise save locally
+            local_path = os.path.join(tempfile.gettempdir(), f"krkn_metrics_{run_uuid}.json")
```
paigerube14, line 265: you imported the write_json_to_tmp function but didn't actually use it. Lets replace this code block with using that function

paigerube14:
@abhinavs1920 thanks so much for this contribution, since there was another PR that had the similar changes and was being routinely updated, we decided to merge that one. Hope to have more contributions from you in the future, thanks!

PR #834, opened 2025-06-03
Ensure metrics are always saved with improved local fallback

levi178u:
fixes: #829 
- trying `ElasticSearch` if available and save locally if it fails.
- Local saves go to a organized location with good metadata
- direct file writing to save the metrics and metadata in JSON format at `/tmp/krkn_<run_uuid>.json.`

### DIFF
--- a/krkn/prometheus/client.py
+++ b/krkn/prometheus/client.py
@@ -9,6 +9,7 @@ import logging
 import urllib3
 import sys
 import json
+import tempfile
 
 import yaml
 from krkn_lib.elastic.krkn_elastic import KrknElastic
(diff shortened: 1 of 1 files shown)


yogananth-subramanian (commented):
Great, looks like a very useful feature.  Does it feature overlap with #833 ?

krkn/prometheus/client.py
```diff
@@ -251,11 +252,31 @@ def metrics(
...
-        if elastic:
+        if elastic is not None and elastic_metrics_index is not None:
             result = elastic.upload_metrics_to_elasticsearch(
                 run_uuid=run_uuid, index=elastic_metrics_index, raw_data=metrics_list
             )
             if result == -1:
                 logging.error("failed to save metrics on ElasticSearch")
+                save_metrics = True
+            else:
+                save_metrics = False
```
paigerube14: maybe instead of this setting to false, let's set save_metrics to false before the if elastic line

krkn/prometheus/client.py
```diff
@@ -251,11 +252,31 @@ def metrics(
...
+            save_metrics = True
+        if save_metrics:
+            local_dir = os.path.join(tempfile.gettempdir(), "krkn_metrics")
+            os.makedirs(local_dir, exist_ok=True)
+            local_file = os.path.join(local_dir, f"{elastic_metrics_index}_{run_uuid}.json")
 
+            try:
+                with open(local_file, "w") as f:
+                    json.dump({
+                        "metadata": {
```
paigerube14: I think this metadata key is a little redundant, I think we can just have 

```
{
  "run_uuid": "2b949e8f-d065-4a1a-9c89-54df598b2cd1"
  "metrics": [
    {...}
  ]
}
```

paigerube14 (commented):
minor nits, but overall this PR looks good to me! thanks so much for the PR

krkn/prometheus/client.py
```diff
@@ -251,11 +252,29 @@ def metrics(
...
+        if elastic is not None and elastic_metrics_index is not None:
             result = elastic.upload_metrics_to_elasticsearch(
                 run_uuid=run_uuid, index=elastic_metrics_index, raw_data=metrics_list
             )
             if result == -1:
                 logging.error("failed to save metrics on ElasticSearch")
+                save_metrics = True
+        else:
+            save_metrics = True
+        if save_metrics or elastic is None:
```
paigerube14: here, I think the `elastic is None` is overkill. I think the if statement can just be`if save_metrics`

levi178u: Ok i'll change it

Issue #836, opened 2025-06-03
Replace invoke of oc debug with function in krkn-lib


paigerube14:
For security reasons its better to limit the amount of time we use the _invoke_ function. One of the remaining places is [here](https://github.com/krkn-chaos/krkn/blob/5bdbf622c32282e1978cc17036afd4096546354a/krkn/scenario_plugins/node_actions/abstract_node_scenarios.py#L126), we already have a function in krkn-lib to do that exact functionality. This invoke just needs to be replaced with the krkn-lib calling of [this function](https://github.com/krkn-chaos/krkn-lib/blob/63e58983997ff062fe4bd1ccff1fca4ac4c63b69/src/krkn_lib/k8s/krkn_kubernetes.py#L1071)

khushiiagrawal:
@paigerube14  I see the issue is not being fixed in the previous PR. I would like to contribute to this issue.
thanks!

PR #838, opened 2025-06-03
Refactor: use of K8s APi from krkn-lib instead of oc debug cmd

levi178u:
fixes: #836 
- Using K8s APi directly from `krkn-lib` instead of `oc debug` cmd
- generating unique pod names to avoid conflicts using `krkn-lib`'s controlled pod creation
- removed `invoke` dependency

### DIFF
--- a/krkn/scenario_plugins/node_actions/abstract_node_scenarios.py
+++ b/krkn/scenario_plugins/node_actions/abstract_node_scenarios.py
@@ -6,7 +6,6 @@ import krkn.scenario_plugins.node_actions.common_node_functions as nodeaction
 from krkn_lib.k8s import KrknKubernetes
 from krkn_lib.models.k8s import AffectedNode, AffectedNodeStatus
 
-k8s_client = KrknKubernetes()
 
 # krkn_lib
 class abstract_node_scenarios:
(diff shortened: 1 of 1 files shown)


krkn/scenario_plugins/node_actions/abstract_node_scenarios.py
```diff
@@ -6,6 +6,8 @@
 from krkn_lib.k8s import KrknKubernetes
 from krkn_lib.models.k8s import AffectedNode, AffectedNodeStatus
 
+k8s_client = KrknKubernetes()
```
tsebastiani: check #838 for the same issue with this approach. Please Inject!

tsebastiani (commented):
**Context**: I've tested this PR on kind with `docker` cloud
While testing this PR I've found some issues not strictly related to it, but is better to keep track of them:
- docker lib needs to be updated to 7.1.0 
- needs a rebase because affected_nodes_status has not been injected [in the docker constructor](https://github.com/krkn-chaos/krkn/blame/d47286ae21ebbe5c224fa8d8ebc869d1f0377e49/krkn/scenario_plugins/node_actions/node_actions_scenario_plugin.py#L112)

Regarding this PR these are my observations:
- the scenario seems to not have effect on kind nodes
- the pod used to inject the scenario is not removed (that's the main problem)

PR #841, opened 2025-06-05
Fix: network scenario timing issue

yogananth-subramanian:
## Description  
Introduce a delay in network scenarios prior to imposing restrictions. This ensures that chaos test case jobs are scheduled before any restrictions are put in place.

### DIFF
--- a/krkn/scenario_plugins/native/network/ingress_shaping.py
+++ b/krkn/scenario_plugins/native/network/ingress_shaping.py
@@ -661,7 +661,7 @@ def get_ingress_cmd(
             )
         tc_set += ";"
 
-    exec_cmd = "{0} {1} sleep {2};{3} sleep 20;{4}".format(
+    exec_cmd = "sleep 30;{0} {1} sleep {2};{3} sleep 20;{4}".format(
         tc_set, tc_ls, duration, tc_unset, tc_ls
     )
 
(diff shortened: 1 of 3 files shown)


krkn/scenario_plugins/native/network/ingress_shaping.py
```diff
@@ -661,7 +661,7 @@ def get_ingress_cmd(
             )
         tc_set += ";"
 
-    exec_cmd = "{0} {1} sleep {2};{3} sleep 20;{4}".format(
+    exec_cmd = "sleep 30;{0} {1} sleep {2};{3} sleep 20;{4}".format(
```
chaitanyaenr, line 664: Should we check the status of the job instead of sleep @yogananth-subramanian? Thoughts?

yogananth-subramanian: Hello @chaitanyaenr , good point, thought of introducing  a quick fix for the existing code.
May be in the long run, we might need to replace shell script code, with a python one that is able to  use the k8s api to query the status of other jobs before it applies the restrictions.

chaitanyaenr: Merging the PR. Let's create a issue for the long term fix to keep track of it @yogananth-subramanian. Thanks!

PR #842, opened 2025-06-05
Adding kube_check as option into node scenarios

paigerube14:
## Description  
Sometimes during node scenarios we may lose connection to the cluster. If that happens we currently are stuck waiting on the kubernetes api to report that the node goes to a certain state. The user should be able to get past this error if they know they may lose connection to the api server

## Documentation  
- [X] **Is documentation needed for this update?**


chaitanyaenr (approved):
LGTM!

Issue #852, opened 2025-06-25
[Rollback Scenarios] Meta Issue


jason810496:
Here is the breakdown of sub-issues for the [Scenario rollback feature #804](https://github.com/krkn-chaos/krkn/issues/804) to improve task tracking.

### Design
- [x] Design Doc: [CNCF - Krkn: Chaos Scenario Rollback Feature (2025 Term 2)](https://docs.google.com/document/d/11KwSi_c_WkkD6I2jbBEMQR-GEjIqOVVu0fviBVQ-9cQ/edit)

### Core
- [ ] Implement `RollbackHandler` #853
- [ ] Add unit tests for `RollbackHandler` and serializer
- [ ] Ensure graceful shutdown behavior

### CLI
- [ ] Execute-Rollback cmd
- [ ] Dry-Run cmd
- [ ] List/ Delete Versions cmd

### Plugin Scenarios
- [ ] [Application Outage Scenarios](https://github.com/jason810496/krkn/blob/feature/rollback-scenario-docs/ROLLBACK_SCENARIOS.md#application-outage-scenarios) https://github.com/krkn-chaos/krkn/pull/853
- [ ] [Container Scenarios](https://github.com/jason810496/krkn/blob/feature/rollback-scenario-docs/ROLLBACK_SCENARIOS.md#container-scenarios) ( No need rollback )
- [ ] [Hog Scenarios](https://github.com/jason810496/krkn/blob/feature/rollback-scenario-docs/ROLLBACK_SCENARIOS.md#hog-scenarios) https://github.com/krkn-chaos/krkn/pull/853
- [ ] kubevirt-vm-outage https://github.com/krkn-chaos/krkn/pull/887
- [ ] [Managed Cluster Scenarios](https://github.com/jason810496/krkn/blob/feature/rollback-scenario-docs/ROLLBACK_SCENARIOS.md#managed-cluster-scenarios) ( Still try to setup with Kind )
- [ ] [Native Scenario Plugin](https://github.com/jason810496/krkn/blob/feature/rollback-scenario-docs/ROLLBACK_SCENARIOS.md#native-scenario-plugin) -> question about`arcaflow_plugin_sdk` 
- [ ] [Network Chaos Scenarios](https://github.com/jason810496/krkn/blob/feature/rollback-scenario-docs/ROLLBACK_SCENARIOS.md#network-chaos-scenarios) https://github.com/krkn-chaos/krkn/pull/887
[...]

PR #853, opened 2025-06-25
[Rollback Scenarios] Core Implementation

jason810496:
## Description  

Core Implementation for Rollback Scenarios
- **New Krkn Configs**
```yaml
    auto_rollback: False                                    # Enable auto rollback when a scenario fails
    rollback_versions_directory: /tmp/kraken-rollback            # The rollback version files will be stored in 
```
- **Components**
  - `RollbackHandler`
    - previous named as `RollbackCallableTracer`, but it's more like a _logger-like Handler_, so rename to `RollbackHandler`
    - Inject as attribute instead of parameter in `run` method
      - Make it centerlize, only need to change the initialization of `AbstractScenarioPlugin`
  - `set_rollback_context_decorator` decorator
    - Since the `RollbackHandler` ( `Serializer` underhood ) use the `run_uuid` as current context, we need a decorator for proper setup/ teardown  
    - Will set `run_uuid` as the context of `RollbackHandler` before the `run` method execute, and clean the context after the `run` method call
  - `Serializer`
    - Serialize callable (function) to version files (a executable python file) based on given context (`run_uuid`)
    - Written by ourself to reduce the project dependency, since for the most case, the most complex type we need to handle is the krkn-lib related class
      - So we handle krkn-lib related class as special case
      - IMO, the rest of the parameters should be simple enough to serialize.

## Documentation  
  - Yes, I will raise the follow-up PR for the docs after most of the PR are done, the task is tracked at https://github.com/krkn-chaos/krkn/issues/852


config/config.yaml
```diff
@@ -5,6 +5,8 @@ kraken:
     signal_state: RUN                                      # Will wait for the RUN signal when set to PAUSE before running the scenarios, refer docs/signal.md for more details
     signal_address: 0.0.0.0                                # Signal listening address
     port: 8081                                             # Signal port
+    auto_rollback: False                                    # Enable auto rollback when a scenario fails
+    rollback_versions_directory: /tmp/kraken-rollback            # The rollback version files will be stored in {versions_directory}/{scenario_type}/{run_uuid} directory
```
jason810496: I had also thought about leveled structure, but left it as single level now. As we don't have much config for rollback feature.

```yaml
rollback:
  auto: False
  versions_directory: /tmp/kraken-rollback 
```

krkn/rollback/handler.py
```diff
@@ -0,0 +1,129 @@
...
+        sig = inspect.signature(func)
+        param_names = (name for name in sig.parameters.keys() if name != 'self')  # Exclude 'self'
+        # Initialize kwargs with all param_names as key
+        kwargs: dict = {name: None for name in param_names}
+        kwargs.update(orig_kwargs) # Update kwargs with original kwargs, if the parameter is set by args will be None
+        args = list(orig_args)
+        # Fill the missing kwargs with original args
+        for param_name in param_names:
+            if not kwargs.get(param_name, None) and args:
+                kwargs[param_name] = args.pop(0)
```
jason810496: After merging the `orig_kwargs` and `orig_args`, no matter the `run_uuid` is passed via kwargs or args way, we should be able to get `run_uuid` from the `kwargs` dict.

Then we call call the `set_context` and `clear_context` for the `self.rollback_handler` in the decorator.

krkn/rollback/handler.py
```diff
@@ -0,0 +1,129 @@
...
+        scenario_type: str,
+        rollback_config: RollbackConfig
+    ):
+        self.rollback_config = rollback_config
+        self.run_uuid = None
+        self.scenario_type = scenario_type
+        self.serializer = Serializer(
+            scenario_type=scenario_type,
+            versions_directory=self.rollback_config.versions_directory
+        )
```
jason810496: Each rollback handler is bound with provided `scenario_type` as part of version file full path, but `run_uuid` will be set with the `set_rollback_context_decorator` before each `scenario_plugin.run` method call.

krkn/rollback/serialization.py
```diff
@@ -0,0 +1,331 @@
...
+
+@dataclass
+class Version:
+    rollback_callable_code: str
+    has_lib_telemetry: bool = False
+    has_kubecli: bool = False
+    has_scenario_telemetry: bool = False
+    serialized_arguments: Optional[list[Any]] = None
+    serialized_kwargs: Optional[dict[str, Any]] = None
+    timestamp: int = time.time_ns() # Get current timestamp in nanoseconds
```
jason810496: I set the timestamp to nanosecond, it should be precise enough for individual call.

krkn/rollback/handler.py
jason810496: I'm also thinking about having `RollbackConfig` be initialized in the module level, instead of having it be initialized in each scenario plugin class instances. Since the `RollbackConfig` should be the same for all the `RollbackHandler` instances.

krkn/rollback/serialization.py
```diff
@@ -0,0 +1,331 @@
...
+
+        if not (
+            hasattr(callable, "__qualname__")
+            and callable.__qualname__.endswith(".rollback_callable")
+            and hasattr(callable, "__func__")
+            and callable.__func__.__name__ == "rollback_callable"
+        ):
+            logger.warning(
+                "The callable may not be an AbstractScenarioPlugin.rollback_callable staticmethod"
+            )
```
jason810496: Maybe we can remove this validation. IMO, we don't need to restrict the name of the actual rollback function, as long as it's a `staticmethod` or a normal function definition, it should be fine to be set via `self.rollback_handler.set_rollback_callable`.

krkn/scenario_plugins/scenario_plugin_factory.py
```diff
@@ -29,11 +31,12 @@ def create_plugin(self, scenario_type: str) -> AbstractScenarioPlugin:
         :param scenario_type: the scenario type defined in the config.yaml
             e.g. `arcaflow_scenarios`, `network_scenarios`, `plugin_scenarios`
             etc.
+        :param rollback_config: the configuration for the rollback handler
         :return: an instance of the class that implements this scenario and
             inherits from the AbstractScenarioPlugin abstract class
         """
         if scenario_type in self.loaded_plugins:
-            return self.loaded_plugins[scenario_type]()
+            return self.loaded_plugins[scenario_type](scenario_type, rollback_config)
```
jason810496: So with `RollbackHandler`, we need to provide `scenario_type` and `rollback_config` for constructing all the scenario plugin instances.

Fortunately, the initialization of scenario plugin is centralized here, so we only need to change this factory to adapt the `RollbackHandler` for all scenario plugins.

tests/rollback_scenario_plugins/simple.py
```diff
@@ -0,0 +1,55 @@
...
+from krkn_lib.models.telemetry import ScenarioTelemetry
+
+from krkn.scenario_plugins.abstract_scenario_plugin import AbstractScenarioPlugin
+from krkn.rollback.handler import set_rollback_context_decorator
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+logger.addHandler(logging.StreamHandler())
+
+class SimpleRollbackScenarioPlugin(AbstractScenarioPlugin):
```
jason810496, line 15: Here is the example usage of new `AbstractScenarioPlugin`, this class is also used in `test_rollback.py` integration test.

config/config.yaml
tsebastiani: I would keep it simple (as it is) unless it's not absolutely needed.

krkn/rollback/handler.py
tsebastiani: I like the decorator approach, well done.

tests/test_rollback.py
```diff
@@ -0,0 +1,100 @@
...
+            handlers=[
+                logging.FileHandler(os.path.join(TEST_LOGS_DIR, "test_rollback.log")),
+                logging.StreamHandler(),
+            ],
+        )
+
+    @pytest.fixture(scope="module")
+    def kubeconfig_path(self):
+        # Provide the path to the kubeconfig file for testing
+        return os.getenv("KUBECONFIG", "/path/to/your/kubeconfig")
```
tsebastiani: I would set it to `~/.kube/config`

tsebastiani (approved):
As we discussed on slack I approve the overall design of the solution, but for the end-to-end implementation on the application outage scenario, highlight how the `KrknTelemetryOpenshift`object is injected in `rollback_callable` context. Please convert this PR to Draft and push the implementation in this PR (even if it becomes bloated I prefer to have the core + the first scenario on the same branch for a while so it can be validated). Great Job @jason810496!

krkn/rollback/serialization.py
tsebastiani: From the discussion had this morning on slack I'd like to reiterate on this concept:
`any scenario executes any possible action only using the AbstractPluginAPI run method parameters`
This means that, even if I **extremely** appreciate your python skills that you're demonstrating from your implementation, from my angle, as project mantainer I worked hard in the past years to remove all the not needed complexities and to try to uniform the objects interactions using clear interfaces.
In my opinion the `rollback_callable` method should always have the following `AbstractPluginScenario::run` subset of parameters (plus anything else if strictly needed of course, but always the same number and type of params):

- `run_uuid: str`
- `scenario: str`
- `krkn_config: dict[str, any]`
- `lib_telemetry: KrknTelemetryOpenshift`

and all the `serialization.py` complexities relative to the `rollback_callable` parameters construction and inference should be ditched. 
If you want to keep this logic you've to provide a convincing argument to justify such complexity in favor of a more simple and streamlined approach.

tsebastiani: cc @chaitanyaenr @paigerube14

jason810496: > In my opinion the `rollback_callable` method should always have the following `AbstractPluginScenario::run` subset of parameters (plus anything else if strictly needed of course, but always the same number and type of params):
> 
> * `run_uuid: str`
> * `scenario: str`
> * `krkn_config: dict[str, any]`
> * `lib_telemetry: KrknTelemetryOpenshift`
> 
> and all the `serialization.py` complexities relative to the `rollback_callable` parameters construction and inference should be ditched. If you want to keep this logic you've to provide a convincing argument to justify such complexity in favor of a more simple and streamlined approach.

Agree to have simpler interface.

However here are my concerns.
I will take **application outage scenario** as example and describe my concerns:

### 1. The dynamic resources created by krkn in runtime are not visible for user

  The name of network policy is generated in runtime:
  https://github.com/krkn-chaos/krkn/blob/b8cf1d059703123d541469bcef36392164b13230/krkn/scenario_plugins/application_outage/application_outage_scenario_plugin.py#L36
  And we don't log the  `policy_name` currently, which means the only way for user to track the resource is using kubectl and list the network policies with `krkn-deny-` prefix, then looked up the created time and identify by themself to know which network policy is created in current run.

  Even we store these information passed from `AbstractPluginScenario::run`
  > * `run_uuid: str`
  > * `scenario: str`
  > * `krkn_config: dict[str, any]`
  
  we still **don't** have the ability to get those variables generated in runtime ( `policy_name` in this example ).
  IMO, as a rollback version files, they should provide information of exact resources created by current run.

### 2. The rollback version files need exact snapshot of all yaml files

  > * `run_uuid: str`
  > * `scenario: str`
  > * `krkn_config: dict[str, any]`

Even we store these parameters in rollback version files, we still rely on `config.yaml` and all the sub-yamls ( `scenarios/openshift/app_outage.yaml` in this exmaple ) referenced by root `config.yaml`.

Since we get the yaml content of current scenario in runtime currently.

https://github.com/krkn-chaos/krkn/blob/b8cf1d059703123d541469bcef36392164b13230/krkn/scenario_plugins/application_outage/application_outage_scenario_plugin.py#L23-L33

If user modify the `namespace` in `scenarios/openshift/app_outage.yaml` after the run, then run the rollback version files that only store the parameters passed by `AbstractPluginScenario::run` solely. This will result in deleting wrong resources in rollback execution.

IMO, there is the need to store the `config.yaml` related informations ( `namespace` in this example ) that require by rollback function even if these informations can be retrieve from the `config.yaml`.

jason810496: Combining our opinions, to simplify the `rollback_handler.set_rollback_callable` interface, I would make it **`**kwargs, lib_telemetry: KrknTelemetryOpenshift`**. What do you think?

That `lib_telemetry` will always be constructed in version file and pass to `rollback_callable`, the rest of `**kwargs` are the required parameters of  `rollback_callable` and could show the exact identifiers ( `namspace` and `policy_name` in above example )  for users to rollback the resources created by krkn.

tsebastiani: Hey @jason810496 that's exactly the point, what I don't like is to have a variadic function but the exact opposite, I don't mind how many parameters we're passing to the rollback_callable method (even if I think that my subset + the resource name and type eventually should be enough) as I said here
>  following AbstractPluginScenario::run subset of parameters (**plus anything else if strictly needed of course**, but always the same number and type of params):

My proposal is to define the minimum set of parameters needed and let's stick to that.

Thank you @jason810496 for having this conversation with me, it's really enjoyable 😊✌️.

jason810496: Nice! I'm glad to hear that we're aligned on the idea.

> Thank you @jason810496 for having this conversation with me, it's really enjoyable 😊✌️.

I'm happy to have the conversation to dive deeper and rethink whole structure as well. It's a nice to talk with you!

I will refactor the interface we just discussed these day tomorrow, and we can go through the implementation of new interface ( and some changes related to rollback config ) in tomorrow meeting.

tsebastiani: Just to clarify **kwargs represents the definition of [variadic function](https://en.m.wikipedia.org/wiki/Variadic_function) in python and my idea is to have a definite set of arguments to have a clear interface (the opposite), I'm sorry to enforce on concepts that do not belong to python programming (as type hints for example) but I come from a different background and I feel more confident having these concept in my code base. Also sorry if the clarification is redundant.

jason810496 (commented):
Hi @tsebastiani,

I just refactor the `RollbackContent` interface and `Serializer` today.

The `RollbackContent`
- will only have the following 2 fields for or all scenarios
  - `resource_identifier: str`
  - `namespace: Optional[str]`
- and adding `__str__` method for serializing it will be enough.

For `Serializer` side
- it will only parse the code of rollback callable and render it with the Jinja2-based version template file
  - Having a Jinja template for version file will be better for us to see how will the version file looks like.
- and we don't need to handle the dynamic amount of arguments and kwargs any more.

Sorry I didn't get the the concept of "how having limited amount of parameters can still work" until you mentioning "they are just a name" in the meeting yesterday. I stuck with the exact naming for those keyword arguments, e.g. `pod_name`, `job_name`, `policy_name` ... etc previously 😓 and totally agree to have a "unified field" ( I called it `resource_identifier` now ) for all scenarios.

 If we are agree with the new interface, I can go on finish the unit test for the core implementation of rollback feature.

Thanks and look forward to your feedback!

krkn/rollback/config.py
```diff
@@ -0,0 +1,150 @@
...
+    def __new__(cls, run_uuid: str):
+        timestamp = int(time.time())
+        return super().__new__(cls, f"{timestamp}-{run_uuid}")
+
+
+class RollbackConfig(metaclass=SingletonMeta):
+    """Configuration for the rollback scenarios."""
+
+    def __init__(self):
+        self._auto = False
```
jason810496, line 55: _auto -> True instead of False

run_kraken.py
```diff
@@ -62,6 +63,18 @@ def main(cfg) -> int:
             config["kraken"], "publish_kraken_status", False
         )
         port = get_yaml_item_value(config["kraken"], "port", 8081)
+        RollbackConfig.register(
+            auto=get_yaml_item_value(
+                config["kraken"],
+                "auto_rollback",
+                False
```
jason810496, line 69: False -> True

run_kraken.py
```diff
@@ -62,6 +63,18 @@ def main(cfg) -> int:
...
+            ),
+            versions_directory=get_yaml_item_value(
+                config["kraken"],
+                "rollback_versions_directory",
+                "/tmp/kraken-rollback"
```
jason810496, line 74: TODO: cleanup version file:
if rollback successfully -> we should delete the corresponding version files.

krkn/rollback/config.py
```diff
@@ -0,0 +1,150 @@
...
+    @classmethod
+    def register(cls, auto=False, versions_directory=""):
+        """Initialize and return the singleton instance with given configuration."""
+        instance = cls()
+        instance.auto = auto
+        instance.versions_directory = versions_directory
+        instance._registered = True
+        return instance
+
+    @classmethod
```
jason810496, line 88: new structure:
`{versions_directory}/<timestamp(ns)>-<run_uuid>/{scenario_type}-<timestamp(ns)>-<random_hash>.py`

krkn/scenario_plugins/hogs/hogs_scenario_plugin.py
```diff
@@ -70,6 +71,13 @@ def run_scenario_worker(self, config: HogConfig,
             pod_name = f"{config.type.value}-hog-{get_random_string(5)}"
             node_resources_start = lib_k8s.get_node_resources_info(node)
             lib_k8s.deploy_hog(pod_name, config)
+            self.rollback_handler.set_rollback_callable(
```
jason810496, line 86: make the `set_rollback_callable` before the execution of chaos for all scenarios.

e.g. hog pod is pending and krkn-cli timeout. the hog pod will still in the cluster.

krkn/rollback/handler.py
```diff
@@ -0,0 +1,127 @@
...
+    return wrapper
+
+
+class RollbackHandler:
+    def __init__(
+        self,
+        scenario_type: str,
+    ):
+        self.scenario_type = scenario_type
+        self.serializer = Serializer(
```
tsebastiani: Don't you think that would be better injecting Serializer as a constructor parameter instead of initializing it in the constructor avoiding coupling between the two classes?

krkn/rollback/handler.py
```diff
@@ -0,0 +1,127 @@
...
+from typing import cast, TYPE_CHECKING
+
+from krkn.rollback.config import RollbackConfig, RollbackContext, Version
+from krkn.rollback.serialization import Serializer
+
+logger = logging.getLogger(__name__)
+
+
+if TYPE_CHECKING:
+    from krkn.scenario_plugins.abstract_scenario_plugin import AbstractScenarioPlugin
```
tsebastiani, line 12: unused import

krkn/rollback/serialization.py
```diff
@@ -0,0 +1,122 @@
...
+class Serializer:
+    def __init__(self, scenario_type: str):
+        self.scenario_type = scenario_type
+        # Set up Jinja2 environment to load templates from the rollback directory
+        template_dir = os.path.join(os.path.dirname(__file__))
+        env = Environment(loader=FileSystemLoader(template_dir))
+        self.template = env.get_template("version_template.j2")
+
+    def _parse_rollback_callable_code(
+        self, callable: "RollbackCallable"
```
tsebastiani: `callable` as parameter name shadows built-in name callable, what do you think about renaming to `rollback_callable`

krkn/rollback/serialization.py
```diff
@@ -0,0 +1,122 @@
...
+                cleaned_lines.append(line.lstrip())
+
+        # Reconstruct the code and clean up trailing whitespace
+        function_code = "\n".join(cleaned_lines).rstrip()
+
+        return function_name, function_code
+
+    def serialize_callable(
+        self,
+        callable: "RollbackCallable",
```
tsebastiani: same here ^

krkn/rollback/serialization.py
```diff
@@ -0,0 +1,122 @@
...
+        rollback_content: "RollbackContent",
+        version: "Version",
```
tsebastiani, line 93: add version to docstring

krkn/scenario_plugins/scenario_plugin_factory.py
```diff
@@ -61,7 +63,9 @@ def __load_plugins(self, base_class: Type):
                             continue
 
                         cls = getattr(module, name)
-                        instance = cls()
+                        # To construct an instance of ScenarioPlugin, we need scenario_type.
+                        # Since we only call get_scenario_types() here, using placeholder values is sufficient.
+                        instance = cls("placeholder_scenario_type")
```
tsebastiani, line 68: This is not exactly clear, can you explain better what's going on here?

krkn/rollback/handler.py
jason810496: The `AbstractScenarioPlugin` is imported under `if TYPE_CHECKING` block.

This is intended for further `self = cast("AbstractScenarioPlugin", self)` to make IDE aware of  `self` is actually `AbstractScenarioPlugin` interface in `def wrapper(self, *args, **kwargs):` decorator.

krkn/rollback/handler.py
jason810496: Just moved initialization of `Serializer` to `AbstractScenarioPlugin` constructor and pass it as parameter for `RollbackHandler`.

PR #868, opened 2025-07-22
Update network scenario image

chaitanyaenr:
## Description  
This commit updates fedora tools image reference used by the network scenarios to the one hosted in the krkn-chaos quay org. This also fixes the issues with RHACS flagging runs when using latest tag by using tools tag instead:

```
Container 'modtools' has image with tag 'latest'\n\n\nIn case of emergency, add the annotation {\"[admission.stackrox.io/break-glass\](http://admission.stackrox.io/break-glass%5C)": \"ticket-xyz\"} to your deployment with an updated ticket number\n\n","reason":"Failed currently enforced policies from RHACS","code":400}
```

### DIFF
--- a/CI/templates/outage_pod.yaml
+++ b/CI/templates/outage_pod.yaml
@@ -8,7 +8,7 @@ spec:
   hostNetwork: true
   containers:
   - name: fedtools
-    image: docker.io/fedora/tools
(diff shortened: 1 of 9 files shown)


paigerube14 (approved):
/lgtm

PR #888, opened 2025-08-19
Adding node_label_selector for pod scenarios

shahsahil264:
## Description  
  
Adding this feature, so that we can be more specific on only targeting pods on worker/master nodes. 

## Documentation  

### DIFF
--- a/krkn/scenario_plugins/pod_disruption/pod_disruption_scenario_plugin.py
+++ b/krkn/scenario_plugins/pod_disruption/pod_disruption_scenario_plugin.py
@@ -188,14 +188,7 @@ class PodDisruptionScenarioPlugin(AbstractScenarioPlugin):
         if not namespace: 
             logging.error('Namespace pattern must be specified')
 
-        # Handle comma-separated node names
-        node_names = config.node_names
-        if isinstance(node_names, str) and node_names:
-            node_names = node_names.split(",")
-        elif not isinstance(node_names, list):
-            node_names = []
-
-        pods = self.get_pods(config.name_pattern,config.label_selector,config.namespace_pattern, kubecli, field_selector="status.phase=Running", node_label_selector=config.node_label_selector, node_names=node_names)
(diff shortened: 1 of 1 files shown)


scenarios/openshift/openshift-apiserver.yml
```diff
@@ -3,5 +3,6 @@
   config:
     namespace_pattern: ^openshift-apiserver$
     label_selector: app=openshift-apiserver-a
+    node_label_selector: node-role.kubernetes.io/master=  # Target master nodes only
```
paigerube14: think lets take off the node labels for the base pod scenarios, can keep in the custom app pod one

krkn/scenario_plugins/pod_disruption/pod_disruption_scenario_plugin.py
```diff
@@ -115,6 +115,22 @@ def get_pods(self, name_pattern, label_selector,namespace, kubecli: KrknKubernet
             pods = kubecli.select_pods_by_name_pattern_and_namespace_pattern(pod_name_pattern=name_pattern, namespace_pattern=namespace, field_selector=field_selector)
         else: 
             logging.error('Name pattern or label pattern must be specified ')
+        
+        # Filter pods by node label selector if specified
+        if node_label_selector and pods:
+            filtered_pods = []
+            nodes_with_label = kubecli.list_nodes(label_selector=node_label_selector)
```
paigerube14, line 158: let's maybe add an option to list nodes to add in field selector of nodeName and do multiple calls if there are multiple node names (couldn't find a way to add multiple node names in one call)

CI/tests/test_customapp_pod.sh
```diff
@@ -0,0 +1,18 @@
+set -xeEo pipefail
```
paigerube14, line 1: @shahsahil264 you'll need to add a line to add this test to the list of tests run in the test workflow: https://github.com/krkn-chaos/krkn/blob/8c05e44c23f7725ad8f6b01de6b7cdaf65c2d58d/.github/workflows/tests.yml#L75

PR #889, opened 2025-08-21
[Rollback Scenarios] Fix cleanup_rollback_version_files error

jason810496:
## Why

Since most of the scenarios haven't integrated rollback feature, but `cleanup_rollback_version_files` will assert the length of the rollback context directory be `1.
This cause the following failure CI: https://github.com/krkn-chaos/krkn/actions/runs/17102009865/job/48500866187

## What  

Instead of raising `ValueError` when length is not `1`, we should just log warning and return empty list.


tsebastiani (commented):
Please add the rollback options to the `config.yaml` with a brief description.

PR #909, opened 2025-10-01
Resiliency Score Draft

abhinavs1920:
## Description  
This PR introduces a comprehensive Resiliency Scoring System that evaluates and quantifies system stability during chaos testing. The implementation includes:

- **Scoring Engine**: Core logic for calculating resiliency scores
- **Scenario-wise Analysis**: Granular scoring per test scenario
- **Telemetry & Reporting**: Detailed metrics 
- **Configurable SLOs**: Support for custom Service Level Objectives

## Type of Change  
- [x] New feature (non-breaking change which adds functionality)

[Documentation PR ](https://github.com/krkn-chaos/website/pull/167)
Require Updates

### DIFF
--- a/run_kraken.py
+++ b/run_kraken.py
@@ -31,6 +31,7 @@ from krkn_lib.ocp import KrknOpenshift
 from krkn_lib.telemetry.k8s import KrknTelemetryKubernetes
 from krkn_lib.telemetry.ocp import KrknTelemetryOpenshift
 from krkn_lib.models.telemetry import ChaosRunTelemetry
+from krkn_lib.models.k8s import ResiliencyReport
 from krkn_lib.utils import SafeLogger
 from krkn_lib.utils.functions import get_yaml_item_value, get_junit_test_case
 
@@ -498,7 +499,13 @@ def main(options, command: Optional[str]) -> int:
         telemetry_json = chaos_telemetry.to_json()
         decoded_chaos_run_telemetry = ChaosRunTelemetry(json.loads(telemetry_json))
         if resiliency_obj and hasattr(resiliency_obj, "summary"):
-            decoded_chaos_run_telemetry.overall_resiliency_report = resiliency_obj.get_summary()
+            summary_dict = resiliency_obj.get_summary()
+            decoded_chaos_run_telemetry.overall_resiliency_report = ResiliencyReport(
+                json_object=summary_dict,
+                resiliency_score=summary_dict.get("resiliency_score", 0),
+                passed_slos=summary_dict.get("passed_slos", 0),
+                total_slos=summary_dict.get("total_slos", 0)
+            )
         chaos_output.telemetry = decoded_chaos_run_telemetry
         logging.info(f"Chaos data:\n{chaos_output.to_json()}")
         if enable_elastic:
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ __pycache__/*
 kube-burner*
 kube_burner*
 recommender_*.json
+resiliency*.json
 
 # Project files
 .ropeproject
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ ibm_vpc==0.26.3  # Requires ibm_cloud_sdk_core
 jinja2==3.1.6
 lxml==5.1.0
 kubernetes==34.1.0
-krkn-lib==6.0.3
+krkn-lib==6.0.4
 numpy==1.26.4
 pandas==2.2.0
 openshift-client==1.0.21


run_kraken.py
```diff
@@ -99,6 +100,17 @@ def main(options, command: Optional[str]) -> int:
         enable_metrics = get_yaml_item_value(
             config["performance_monitoring"], "enable_metrics", False
         )
+
+        # Ensure resiliency scoring can obtain a Prometheus handle even when
+        # metrics/alerts collection is disabled.
+        resiliency_enabled = get_yaml_item_value(
```
paigerube14: let's add a variable into the config/config.yaml with a default variable

run_kraken.py
```diff
@@ -427,8 +448,64 @@ def main(options, command: Optional[str]) -> int:
...
+                    prometheus,
+                    datetime.datetime.fromtimestamp(start_time),
+                    datetime.datetime.fromtimestamp(end_time),
+                )
+                resiliency_obj.calculate_score()
+                resiliency_dict = resiliency_obj.to_dict()
+                chaos_telemetry.resiliency_report = resiliency_dict
+
+                try:
+                    import json, dataclasses
```
paigerube14: lets maybe move this to the resiliency class instead of directly in the run_kraken

tsebastiani: Not sure about this import in this position, json is already imported above, so simply import `dataclasses` in the import section.

run_kraken.py
abhinavs1920: Sure will do that, will add `enabled: true` in the resiliency config

run_kraken.py
abhinavs1920: Yepp, working on it.

abhinavs1920: Yes it should be removed from here, will do that

abhinavs1920:
> things ran well with this first default run. Definitely would be good to add a file that has a different configuration with weights so its clear how to configure it
> 
> couple of small notes:
> 
> * definitely will need documnetation on what failed vs points lost means
> * for slos results, might be cleaner if we only show failures. (definitely also need to clean up the descriptions)
> 
> ```
>  "resiliency_report": {
>             "score": 95,
>             "breakdown": {
>                 "total_points": 45,
>                 "points_lost": 2,
>                 "passed": 25,
>                 "failed": 2
>             },
>             "slo_results": {
>                 "10 minutes avg. 99th etcd fsync latency on {{$labels.pod}} higher than 10ms. {{$value}}s": true,
>                 "10 minutes avg. 99th etcd fsync latency on {{$labels.pod}} higher than 1s. {{$value}}s": true,
>                 "10 minutes avg. 99th etcd commit latency on {{$labels.pod}} higher than 30ms. {{$value}}s": true,
>                 "etcd leader changes 
> ```

Yes working on it, will update the branch as soon I successfully test it out.

run_kraken.py
```diff
@@ -427,8 +448,30 @@ def main(options, command: Optional[str]) -> int:
...
+                )
+
+            except Exception as e:
+                logging.error("Failed to compute resiliency score: %s", e)
+
+
         telemetry_json = chaos_telemetry.to_json()
         decoded_chaos_run_telemetry = ChaosRunTelemetry(json.loads(telemetry_json))
+        # Propagate resiliency report attribute after re-instantiation
+        if hasattr(chaos_telemetry, "resiliency_report"):
```
paigerube14: The telemetry structure is defined in Krkn-lib, thinking let's add this resiliency score attribute to chaos telemetry and also add it to the elastic telemetry so we can track it in the future

krkn/resiliency/resiliency.py
```diff
@@ -0,0 +1,196 @@
...
+                        status = hc.get("status", True)
+                except Exception:
+                    status = False
+                health_results[str(name)] = bool(status)
+
+        resiliency_obj.calculate_score(health_check_results=health_results)
+        resiliency_report = resiliency_obj.to_dict()
+        chaos_telemetry.resiliency_report = resiliency_report
+
+        from krkn_lib.models.telemetry import ChaosRunTelemetry  # late import to avoid cycle
```
tsebastiani: please try to import the dependencies on top of the file.

krkn/resiliency/resiliency.py
```diff
@@ -0,0 +1,196 @@
...
+
+        if not hasattr(ChaosRunTelemetry, "_with_resiliency_patch"):
+            _orig_to_json = ChaosRunTelemetry.to_json
+
+            def _to_json_with_resiliency(self):  
+                import json  # local import to keep module deps minimal
```
tsebastiani: please try to import the dependencies on top of the file.

config/config.yaml
```diff
@@ -125,4 +125,8 @@ kubevirt_checks:                                            # Utilizing virt che
     namespace:                                              # Namespace where to find VMI's
     name:                                                   # Regex Name style of VMI's to watch, optional, will watch all VMI names in the namespace if left blank
     only_failures: False                                    # Boolean of whether to show all VMI's failures and successful ssh connection (False), or only failure status' (True) 
-    disconnected: False                                     # Boolean of how to try to connect to the VMIs; if True will use the ip_address to try ssh from within a node, if false will use the name and uses virtctl to try to connect; Default is False
\ No newline at end of file
+    disconnected: False     
+    
+resiliency:
+    enabled: true
+    prometheus_url: "http://localhost:9090"  # Prometheus URL specifically for resiliency scoring                                  # Boolean of how to try to connect to the VMIs; if True will use the ip_address to try ssh from within a node, if false will use the name and uses virtctl to try to connect; Default is False
```
paigerube14: let's replace this URL with the url we set here: https://github.com/krkn-chaos/krkn/blob/97f4c1fd9c8377738169457735422f499b88b4c0/config/config.yaml#L62

krkn/prometheus/collector.py
```diff
@@ -0,0 +1,152 @@
...
+import math
+import os
+import logging
+from typing import Dict, Any, List
+
+import yaml
+from krkn_lib.prometheus.krkn_prometheus import KrknPrometheus
+
+
+def collect_prometheus_metrics(
```
paigerube14: this function and [this one](https://github.com/krkn-chaos/krkn/blob/97f4c1fd9c8377738169457735422f499b88b4c0/krkn/prometheus/client.py#L170-L243) seem very similar. Let's try to reuse/edit the existing function

krkn/scenario_plugins/scenario_plugin_factory.py
```diff
@@ -47,7 +47,11 @@ def __load_plugins(self, base_class: Type):
         ):
 
             if not is_pkg:
-                module = importlib.import_module(module_name)
```
abhinavs1920: Should we keep this exception handling case, or should I remove it?

krkn/resiliency/resiliency.py
```diff
@@ -0,0 +1,405 @@
...
+        slo_defs = {slo["name"]: slo["severity"] for slo in self._slos}
+        score, breakdown = calculate_resiliency_score(
+            slo_definitions=slo_defs,
+            prometheus_results=self._results,
+            health_check_results=health_check_results or {},
+            weights=weights,
+        )
+        self._score = score
+        self._breakdown = breakdown
+        self._health_check_results = health_check_results or {}
```
tsebastiani, line 111: This should be done in the \_\_init\_\_ method

krkn/resiliency/resiliency.py
```diff
@@ -0,0 +1,405 @@
...
+        )
+        slo_defs = {slo["name"]: slo["severity"] for slo in self._slos}
+        system_stability_score, full_breakdown = calculate_resiliency_score(
+            slo_definitions=slo_defs,
+            prometheus_results=full_slo_results,
+            health_check_results={},
+            weights=weights,
+        )
+
+        self.summary = {
```
tsebastiani, line 214: same as above, I'd suggest to define and initialize all the class members in the \_\_init\_\_ method

krkn/resiliency/resiliency.py
```diff
@@ -0,0 +1,405 @@
...
+            "scenarios": {rep["name"]: rep["score"] for rep in self.scenario_reports},
+            "resiliency_score": resiliency_score,
+            "system_stability_score": system_stability_score,
+            "passed_slos": full_breakdown.get("passed", 0),
+            "total_slos": full_breakdown.get("passed", 0) + full_breakdown.get("failed", 0),
+        }
+
+        self.detailed_report = {
```
tsebastiani, line 222: same as above, I'd suggest to define and initialize all the class members in the \_\_init\_\_ method

krkn/resiliency/resiliency.py
```diff
@@ -0,0 +1,405 @@
...
+                    }
+                if name in score_map:
+                    item_dict["resiliency_report"] = score_map[name]
+                new_scenarios.append(item_dict)
+        chaos_telemetry.scenarios = new_scenarios
+
+    # ------------------------------------------------------------------
+    # Internal helpers
+    # ------------------------------------------------------------------
+    @staticmethod
```
tsebastiani, line 285: this is a duplicate

krkn/resiliency/resiliency.py
```diff
@@ -0,0 +1,405 @@
...
+            if not os.path.exists(alerts_yaml_path):
+                raise FileNotFoundError(f"alerts file not found: {alerts_yaml_path}")
+            with open(alerts_yaml_path, "r", encoding="utf-8") as fp:
+                raw_yaml_data = yaml.safe_load(fp)
+            logging.info("Loaded SLO configuration from %s", alerts_yaml_path)
+            self.prometheus_url = None
+
+        self._slos = self._normalise_alerts(raw_yaml_data)
+        self._results: Dict[str, bool] = {}
+        self._score: int | None = None
```
tsebastiani: use `Optional[int]` instead of `int | None`

krkn/resiliency/resiliency.py
```diff
@@ -0,0 +1,405 @@
...
+        self._breakdown: Dict[str, int] | None = None
```
tsebastiani: use `Optional[Dict[str,int]]` instead of `Dict[str,int]| None`

tsebastiani: same approach at line:
- 101
- 102
- 138
- 139
- 140
- 194
- 333
- 335
- 336

krkn/resiliency/resiliency.py
```diff
@@ -0,0 +1,405 @@
...
+from typing import Dict, List, Any
+
+import yaml
+import json
+import dataclasses
+from krkn_lib.models.telemetry import ChaosRunTelemetry
+
+from krkn_lib.prometheus.krkn_prometheus import KrknPrometheus
+from krkn.prometheus.collector import evaluate_slos
+from .score import calculate_resiliency_score
```
tsebastiani: do not use relative imports, add an \_\_init\_\_.py in the folder so it becomes a legit package, and import the package correctly with `from krkn.resiliency.score import calculate_resiliency_score`

config/config.yaml
paigerube14: Abhinav, are you able to remove this in your next push?

abhinavs1920: In the main resiliency code we will be using the prometheus_url we get from the performance_monitoring? I made changes in that way only will be visible in the next push.

abhinavs1920:
Hey @tsebastiani @paigerube14 @chaitanyaenr @yogananth-subramanian,

The core resiliency feature implementation is ready for a final look.

**What's included in this PR:**
* **Config Flexibility:** It defaults to `alerts.yaml` but also accepts custom SLOs via the environment variable if provided.
In this user need to make a resiliency config yaml file and export it to this env `KRKN_ALERTS_YAML_CONTENT`.
It generates the detailed `resiliency-report.json` separately, keeping the main telemetry lightweight.

Resiliency is enabled once you provide the prometheus url in the config.

This PR handles the raw, unweighted scoring logic. I've implemented the weighted average in the `krknctl` PR.

Please test it out and let me know the feedback!

config/config.yaml
```diff
@@ -128,4 +128,8 @@ kubevirt_checks:                                            # Utilizing virt che
     disconnected: False                                     # Boolean of how to try to connect to the VMIs; if True will use the ip_address to try ssh from within a node, if false will use the name and uses virtctl to try to connect; Default is False
     ssh_node: ""                                            # If set, will be a backup way to ssh to a node. Will want to set to a node that isn't targeted in chaos
     node_names: ""
-    exit_on_failure:                                        # If value is True and VMI's are failing post chaos returns failure, values can be True/False
\ No newline at end of file
+    exit_on_failure:                                        # If value is True and VMI's are failing post chaos returns failure, values can be True/False
+    disconnected: False
```
paigerube14: might be a rebase issue here, there are 2 disconnected fields

abhinavs1920: Yes, my mistake... will fix it.

krkn/prometheus/collector.py
```diff
@@ -0,0 +1,117 @@
...
+    prom_cli: KrknPrometheus,
+    metrics_profile: str | dict,
+    start_time: datetime.datetime,
+    end_time: datetime.datetime,
+    granularity: int = 30,
+) -> list[dict[str, Any]]:
+    """Collect Prometheus metrics described by a metrics profile for a given time window.
+
+    Args:
```
paigerube14: do we need this function at all since its just calling a pre-existing function? this seems not needed

krkn/prometheus/collector.py
```diff
@@ -0,0 +1,117 @@
...
+from krkn_lib.prometheus.krkn_prometheus import KrknPrometheus
+from krkn.prometheus.client import metrics as client_metrics
+
+
+def collect_prometheus_metrics(
+    prom_cli: KrknPrometheus,
+    metrics_profile: str | dict,
+    start_time: datetime.datetime,
+    end_time: datetime.datetime,
+    granularity: int = 30,
```
paigerube14: granularity doesn't seem to be used here

krkn/resiliency/resiliency.py
```diff
@@ -0,0 +1,409 @@
...
+        self._breakdown: Optional[Dict[str, int]] = None
+        self._health_check_results: Dict[str, bool] = {}
+        self.scenario_reports: List[Dict[str, Any]] = []
+        self.summary: Optional[Dict[str, Any]] = None
+        self.detailed_report: Optional[Dict[str, Any]] = None
+
+    # ---------------------------------------------------------------------
+    # Public API
+    # ---------------------------------------------------------------------
+    def evaluate_slos(
```
paigerube14: do we need this function at all since its just calling a pre-existing function? this seems not needed/redundant

run_kraken.py
```diff
@@ -52,6 +53,14 @@ def main(options, command: Optional[str]) -> int:
     print(pyfiglet.figlet_format("kraken"))
     logging.info("Starting kraken")
 
+    # Determine execution mode (standalone, controller, or disabled)
+    run_mode = os.getenv("RESILIENCY_ENABLED_MODE", "standalone").lower()
```
paigerube14: lets add this variable to the resiliency part of the config instead of an enviornment variable. We use enviornment variables in krkn-hub and krknctl but everything for krkn should be set in the config itself

run_kraken.py
```diff
@@ -88,17 +97,24 @@ def main(options, command: Optional[str]) -> int:
         iterations = get_yaml_item_value(config["tunings"], "iterations", 1)
         daemon_mode = get_yaml_item_value(config["tunings"], "daemon_mode", False)
 
-        prometheus_url = config["performance_monitoring"].get("prometheus_url")
-        prometheus_bearer_token = config["performance_monitoring"].get(
-            "prometheus_bearer_token"
-        )
+        # Prometheus configuration: environment variables take precedence over config file
```
paigerube14: these variables are already set in the config by [these lines](https://github.com/krkn-chaos/krkn/blob/4527d073c6a69af86fcbc9f7c3be19b064d40e4e/run_kraken.py#L91-L94). Let's remove these

run_kraken.py
paigerube14: would recommend using get_yaml_item_value to get the value and set the default to "standalone", should be other examples in the run_kraken file

run_kraken.py
```diff
@@ -363,12 +388,51 @@ def main(options, command: Optional[str]) -> int:
...
 
+                        batch_window_start_dt = datetime.datetime.utcnow()
                         failed_post_scenarios, scenario_telemetries = (
                             scenario_plugin.run_scenarios(
                                 run_uuid, scenarios_list, config, telemetry_ocp
                             )
                         )
                         chaos_telemetry.scenarios.extend(scenario_telemetries)
+                        batch_window_end_dt = datetime.datetime.utcnow()
+                        if resiliency_obj:
```
paigerube14, line 405: lets maybe add this for loop into a function in the resiliency/resiliency file. The run_kraken file has almost all calls to other files and very large already. Would be good to reduce as much code as possible in this file

run_kraken.py
```diff
@@ -430,12 +494,56 @@ def main(options, command: Optional[str]) -> int:
...
             telemetry_k8s.collect_cluster_metadata(chaos_telemetry)
 
+        if resiliency_obj:
+            try:
+                resiliency_obj.attach_compact_to_telemetry(chaos_telemetry)
+            except Exception as exc:
+                logging.error("Failed to embed per-scenario resiliency in telemetry: %s", exc)
+
+        if resiliency_obj:
+            try:
```
paigerube14, line 485: similar to [this comment](https://github.com/krkn-chaos/krkn/pull/909/files#r2557181995) lets move this into resiliency/resiliency

krkn/resiliency/resiliency.py
```diff
@@ -0,0 +1,409 @@
...
+                decoded_yaml = base64.b64decode(env_yaml).decode('utf-8')
+                raw_yaml_data = yaml.safe_load(decoded_yaml)
+            except (yaml.YAMLError, base64.binascii.Error, UnicodeDecodeError) as exc:
+                raise ValueError(
+                    f"Invalid base64 or YAML in environment variable {self.ENV_VAR_NAME}: {exc}"
+                ) from exc
+            logging.info("Loaded SLO configuration from environment variable %s", self.ENV_VAR_NAME)
+            # Store optional Prometheus URL if provided
+            if isinstance(raw_yaml_data, dict):
+                self.prometheus_url = raw_yaml_data.get("prometheus_url")  # may be None
```
paigerube14: the prometheus url should always come from the config/config.yaml file, don't think this will be used

krkn/prometheus/collector.py
abhinavs1920: Yes, I just rechecked the implementation, by default it is kept at 10, so I will remove it from here.
Should we increase the granularity for SLO evaluation queries? Thoughts??

paigerube14:
> Also to remove the DCO error I will squash all my commits to 1 or 2 major signed commits, I guess this will resolve the CI errors.

Yes, need to squash to 1 and sign off on the commit after you get the tests passing as well

abhinavs1920:
I’ve fixed the issue, but I’m now seeing another error in the CI related to Docker authentication. I don’t believe this is caused by the krkn code.

run_kraken.py
```diff
@@ -89,16 +102,22 @@ def main(options, command: Optional[str]) -> int:
...
+        prometheus_bearer_token = config["performance_monitoring"].get("prometheus_bearer_token")
         run_uuid = config["performance_monitoring"].get("uuid")
         enable_alerts = get_yaml_item_value(
             config["performance_monitoring"], "enable_alerts", False
         )
         enable_metrics = get_yaml_item_value(
             config["performance_monitoring"], "enable_metrics", False
         )
+
+        # Disable resiliency if Prometheus URL is not available
```
paigerube14: Need to move this to below [this if block](https://github.com/krkn-chaos/krkn/blob/a13150b0f55eb1c773fe2c67437f1d3c46b09f96/run_kraken.py#L206-L223), if the user doesn't set the prometheus_url but it is an openshift distribution it'll try finding the prometheus URL. the resiliency score was not running even though prometheus was set on OpenShift

krkn/prometheus/collector.py
```diff
@@ -0,0 +1,77 @@
...
+from krkn_lib.prometheus.krkn_prometheus import KrknPrometheus
+
+
+# -----------------------------------------------------------------------------
+# SLO evaluation helpers (used by krkn.resiliency)
+# -----------------------------------------------------------------------------
+
+
+def slo_passed(prometheus_result: List[Any]) -> Optional[bool]:
+    if not prometheus_result:
```
paigerube14, line 16: here, if the response is None or a blank list coming from prometheus alerts that means the alert is passing. For example the below, it execute the query and since the sum of etcd cpu is less than 5 it doesn't return a repsonse and the alert should "Pass". We don't want to ignore these

```
# etcd CPU and usage increase
- expr: sum(rate(container_cpu_usage_seconds_total{image!='', namespace='openshift-etcd', container='etcd'}[1m])) * 100 / sum(machine_cpu_cores)  > 5
  description: Etcd CPU usage increased significantly by {{$value}}%
  severity: critical
```

krkn/prometheus/collector.py
```diff
@@ -0,0 +1,77 @@
...
+        name = slo["name"]
+        try:
+            response = prom_cli.process_prom_query_in_range(
+                expr,
+                start_time=start_time,
+                end_time=end_time,
+            )
+
+            passed = slo_passed(response)
+            if passed is None:
```
paigerube14, line 70: getting a lot of these warnings when we should not, see other comment in this file on why

paigerube14:
Going to add the resiliency score details to telemetry that it can push to elasticsearch 

https://github.com/krkn-chaos/krkn-lib/pull/256

paigerube14:
@abhinavs1920 thank you so much for this PR! I used your code as the base and added some custom weight to the calculation and got it merged in this PR: https://github.com/krkn-chaos/krkn/pull/1173 
Appreciate your contributions and hope for more work together in the future

PR #932, opened 2025-10-30
Adding resetting new ip address when restarted

paigerube14:
## Description  
When a vm is killed it is improperly reporting a false status even when it comes back up because it has been assigned a new ip. This is only applicable in the disconnected environment as its using the ip address so check ssh connection 

Reliant on: https://github.com/krkn-chaos/krkn-lib/pull/208

## Documentation  

### DIFF
--- a/krkn/utils/VirtChecker.py
+++ b/krkn/utils/VirtChecker.py
@@ -14,12 +14,12 @@ from krkn_lib.utils.functions import get_yaml_item_value
 class VirtChecker:
     current_iterations: int = 0
     ret_value = 0
-    def __init__(self, kubevirt_check_config, iterations, krkn_lib: KrknKubernetes, threads_limt=20):
+    def __init__(self, kubevirt_check_config, iterations, krkn_lib: KrknKubernetes, threads_limit=20):
         self.iterations = iterations
         self.namespace = get_yaml_item_value(kubevirt_check_config, "namespace", "")
         self.vm_list = []
         self.threads = []
-        self.threads_limit = threads_limt
+        self.threads_limit = threads_limit
         if self.namespace == "":
             logging.info("kube virt checks config is not defined, skipping them")
             return
@@ -28,6 +28,7 @@ class VirtChecker:
         self.disconnected =  get_yaml_item_value(kubevirt_check_config, "disconnected", False)
         self.only_failures =  get_yaml_item_value(kubevirt_check_config, "only_failures", False)
         self.interval = get_yaml_item_value(kubevirt_check_config, "interval", 2)
+        self.ssh_node = get_yaml_item_value(kubevirt_check_config, "ssh_node", "")
         try:
             self.kube_vm_plugin = KubevirtVmOutageScenarioPlugin()
             self.kube_vm_plugin.init_clients(k8s_client=krkn_lib)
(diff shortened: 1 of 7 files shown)


chaitanyaenr (approved):

krkn/utils/VirtChecker.py
```diff
@@ -91,7 +108,14 @@ def run_virt_check(self, vm_list_batch, virt_check_telemetry_queue: queue.Queue)
...
                     else:
-                        vm_status = self.check_disconnected_access(vm.ip_address, vm.node_name)
+                        # if new ip address exists use it 
+                        if vm.new_ip_address: 
+                            vm_status, new_ip_address = self.check_disconnected_access(vm.new_ip_address, vm.node_name, vm.vm_name)
+                            # since we already set the new ip address, we don't want to reset to none each time
+                        else: 
+                            vm_status, new_ip_address = self.check_disconnected_access(vm.ip_address, vm.node_name, vm.vm_name)
+                            vm.new_ip_address = new_ip_address
+                    print('status ' + str(vm_status))
```
yogananth-subramanian: Hello @paigerube14 , seeing status message in a loop, may be due to this line.
If so, maybe we have exclude it from final commit.

krkn/utils/VirtChecker.py
```diff
@@ -91,7 +108,14 @@ def run_virt_check(self, vm_list_batch, virt_check_telemetry_queue: queue.Queue)
                     if not self.disconnected: 
                         vm_status = self.get_vm_access(vm.vm_name, vm.namespace)
                     else:
-                        vm_status = self.check_disconnected_access(vm.ip_address, vm.node_name)
+                        # if new ip address exists use it 
+                        if vm.new_ip_address: 
+                            vm_status, new_ip_address = self.check_disconnected_access(vm.new_ip_address, vm.node_name, vm.vm_name)
+                            # since we already set the new ip address, we don't want to reset to none each time
+                        else: 
+                            vm_status, new_ip_address = self.check_disconnected_access(vm.ip_address, vm.node_name, vm.vm_name)
```
yogananth-subramanian: Hello @paigerube14 , using vm.node_name as worker_name in check_disconneted_access may not work in tests were the VM.node is offline(network scenario, bm reboot), since node is offline, ssh might not work.

yogananth-subramanian: one option would be use one the master node, it will work in most of the scenario, except when doing master outage or setup where VMs are also  hosted on master. 
Other options would be to dynamically check the node availability before using the node for ssh or 
have a option to pass the preferred worker node  to use for ssh, from the user.

paigerube14: I think any option is good, think we can maybe let the user set the master or a node they know they are not targeting in the config?

yogananth-subramanian: Hello @paigerube14 , as you might have hinted, we could have existing code , but add an option to to let the user specify the node in the config.

krkn/utils/VirtChecker.py
```diff
@@ -42,13 +42,30 @@ def __init__(self, kubevirt_check_config, iterations, krkn_lib: KrknKubernetes,
             ip_address = vmi.get("status",{}).get("interfaces",[])[0].get("ipAddress")
             self.vm_list.append(VirtCheck({'vm_name':vmi_name, 'ip_address': ip_address, 'namespace':self.namespace, 'node_name':node_name}))
 
-    def check_disconnected_access(self, ip_address: str, worker_name:str = ''):
-
+    def check_disconnected_access(self, ip_address: str, worker_name:str = '', vmi_name: str = ''):
+        
+        virtctl_vm_cmd = f"ssh core@{worker_name} 'ssh -o BatchMode=yes -o ConnectTimeout=2 -o StrictHostKeyChecking=no root@{ip_address}'"
```
yogananth-subramanian: Hello @paigerube14 , may be we can increase the ssh timeout to 10 or higher, to have some buffer in case the vm is slow to respond and the ssh options for both hosts, something like
ssh  -o BatchMode=yes -o ConnectTimeout=10 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no core@{worker_name} 'ssh -o BatchMode=yes -o ConnectTimeout=10 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@{ip_address}'

paigerube14: updated!

krkn/utils/VirtChecker.py
paigerube14: added a last if statement if "ssh_node" is set to try sshing using that node name

PR #967, opened 2025-11-20
feat:Add exclude label to application outage

### LINKED ISSUE #930: Add exclude label to application outage
Add exclude label to application outage to exclude apps found when pod selector is given

FAUST-BENCHOU:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization

## Description  
Convert the label expression provided by the user in exclude_label into a Kubernetes NetworkPolicy matchExpressions structure using NotIn, so that the specified Pods are excluded from being isolated by the network policy.

- Related Issue #930 
- Closes #930 

## Documentation  

image

### DIFF
--- a/containers/krknctl-input.json
+++ b/containers/krknctl-input.json
@@ -520,6 +520,26 @@
     "required": "false"
   },
   {
+    "name": "kubevirt-exit-on-failure",
+    "short_description": "KubeVirt fail if failed vms at end of run",
+    "description": "KubeVirt fails run if vms still have false status",
+    "variable": "KUBE_VIRT_EXIT_ON_FAIL",
+    "type": "enum",
+    "allowed_values": "True,False,true,false",
+    "separator": ",",
+    "default": "False",
+    "required": "false"
+  },
+  {
+    "name": "kubevirt-node-node",
+    "short_description": "KubeVirt node to filter vms on",
+    "description": "Only track VMs in KubeVirt on given node name",
+    "variable": "KUBE_VIRT_NODE_NAME",
+    "type": "string",
+    "default": "",
+    "required": "false"
+  },
+  {
     "name": "krkn-debug",
     "short_description": "Krkn debug mode",
     "description": "Enables debug mode for Krkn",
--- a/krkn/rollback/config.py
+++ b/krkn/rollback/config.py
@@ -209,6 +209,10 @@ class RollbackConfig(metaclass=SingletonMeta):
             rollback_context_dir = os.path.join(cls().versions_directory, rollback_context_dir)
 
             for file in os.listdir(rollback_context_dir):
+                # Skip known non-rollback files/directories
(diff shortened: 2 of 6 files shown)


scenarios/openshift/app_outage.yaml
```diff
@@ -3,3 +3,4 @@ application_outage:                                  # Scenario to create an out
   namespace: <namespace-with-application>            # Namespace to target - all application routes will go inaccessible if pod selector is empty
   pod_selector: {app: foo}                            # Pods to target
   block: [Ingress, Egress]                           # It can be Ingress or Egress or Ingress, Egress
+  exclude_label: ""                                  # Optional label selector to exclude pods (e.g. "tier=gold" or "tier=gold,env=prod")
```
paigerube14: small nit, because pod selector is set up as a dict here, it would be nice to also have the exclude label as a dict not a string. (or both options would be amazing too)

FAUST-BENCHOU: Of course, I’ll consider supporting dict and string, to lay the groundwork for future improvements.

FAUST-BENCHOU: In fact, my current implementation of exclude_label supports three types: string, dict, and list. You can check the specific implementation in the build_exclude_expressions function, particularly the parts with if isinstance(exclude_label, dict) and if isinstance(exclude_label, list). However, my comments might not be clear enough, so I will update the comments in both the YAML file and the function itself.😃

krkn/scenario_plugins/application_outage/application_outage_scenario_plugin.py
```diff
@@ -122,3 +146,48 @@ def rollback_network_policy(
 
     def get_scenario_types(self) -> list[str]:
         return ["application_outages_scenarios"]
+
+    @staticmethod
+    def _build_exclude_expressions(exclude_label) -> list[dict]:
```
tsebastiani, line 154: LGTM, my two cents, why don't you move this method in the utils package and you make it public?  Might be absolutely reusable for something else.

FAUST-BENCHOU: yeah sure，I plan to refactor it into 
krkn/utils/label_utils.py
                         └── def build_exclude_expressions(exclude_label) -> list[dict]:
What do you think?

tsebastiani: you already have a functions.py that should be enough, for the moment I don't see the need of a new file for a single method, let's keep them here and then, if in the future we'll have multiple label related methods, we'll refactor. Thanks again!

FAUST-BENCHOU: 👌Thx for your feedback！

PR #975, opened 2025-11-24
feat:enable node-network chaos for list of nodes

### LINKED ISSUE #969: enable node-network chaos for list of nodes
**Is your feature request related to a problem? Please describe.**
With krkn-hub:node-network-filter scenario, in the config I can pass in 
instance-count : 1 
target: ip-xx-xxx-xx-x.eu-west-1.compute.internal
This scenario creates a privileged pod that blocks the port 2049 for a particular node.

**Describe the solution you'd like**
I want to extend it to cover a list of nodes
If the target can take a wild card, that should mean on all nodes do this
we might want to skip master nodes
if i can pass in a list of nodes then i can customise the scenario down to relevant nodes and pods

**Additional context**
The idea is to block EFS access for all pods that uses EFS and see the behaviour

FAUST-BENCHOU:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization

## Description  
  
Supports single value, CSV string, list, '*', 'regex:pattern' for target
Supprot exclude_label: Optional[Union[str, List[str]]] for node-network chaos

- Related Issue #
- Closes #969 

## Documentation  


krkn/scenario_plugins/network_chaos_ng/models.py
```diff
@@ -44,7 +44,7 @@ class NetworkFilterConfig(BaseNetworkChaosConfig):
     ingress: bool
     egress: bool
     interfaces: list[str]
-    target: str
+    target: str | list[str] | None
```
tsebastiani: this notation is not supported in python 3.9 (our current python version) you can use `target: Union[str, list[str], None]`

tsebastiani (commented):
Go ahead implementing the exclude label as discussed in the related issue. Thanks a lot for your super valuable contribution!🙏🏆

krkn/scenario_plugins/network_chaos_ng/models.py
tsebastiani: https://github.com/krkn-chaos/krkn/actions/runs/19629220963/job/56209563830?pr=975 tests are failing for the same reason

FAUST-BENCHOU: Have changed it to target: Union[str, List[str], None]

tsebastiani (approved):
/lgtm

scenarios/kube/node-network-filter.yml
```diff
@@ -5,12 +5,13 @@
...
   service_account: ""
   taints: [] # example ["node-role.kubernetes.io/master:NoSchedule"]
+  exclude_label: "" # optional: string or list of selectors for nodes to exclude
   namespace: 'default'
   instance_count: 1
   execution: parallel
   ingress: false
   egress: true
-  target: '<node_name>'
+  target: '<node_name>' # Supports single value, CSV string, list, '*', 'regex:pattern'
```
paigerube14, line 14: instead of 'target' can we use 'node_names' here for the key? Its more aligned with other scenario keys

tsebastiani: There's a constraint related to the generic NetworkChaosNG plugin architecture, there's a `NetworkFilterConfig` common to all the the filter scenarios, that's way we're using a generic `target`property. If we want to change that there will be several objects affected.

Issue #916, opened 2025-10-01
Node Outage


paigerube14:
Want to expand and continue the work started around rolling back the cluster to its original state after a scenario fails in krkn

The current Scenario Plugin API has a new feature in the [plugin abstract class](https://github.com/krkn-chaos/krkn/blob/a072f0306a1b21df2b790a7a38f381d75da4de13/krkn/scenario_plugins/abstract_scenario_plugin.py#L8) to restore all the changes performed by krkn in the cluster in case run method encounters some kind of exception before ending the execution.

Related issues: 
- https://github.com/krkn-chaos/krkn/issues/804
- https://github.com/krkn-chaos/krkn/issues/852

Documentation: 
- https://krkn-chaos.dev/docs/rollback-scenarios/
- https://krkn-chaos.dev/docs/scenarios/node-scenarios/

Helpful code links: 
- https://github.com/krkn-chaos/krkn/blob/main/krkn/rollback/handler.py
- https://github.com/krkn-chaos/krkn/blob/main/krkn/scenario_plugins/node_actions/node_actions_scenario_plugin.py

cynox-66:
I’ve been reviewing the rollback architecture introduced in PR #927 and comparing it with the existing open implementations for #916.

From what I can tell, the current open PRs appear to follow an older rollback pattern and don’t use the newer RollbackContent/cloud-native rollback flow introduced by #927.

I’d like to work on a fresh implementation that mirrors the rollback framework pattern used in #927 for node outage scenarios.

Please let me know if there are any architectural considerations I should be aware of before opening a PR.

viktoriavikulya:
This behavior changed in v2.4 — the migration guide mentions it but it's easy to miss. Adding a note to the docs PR.

PR #1022, opened 2025-12-24
Implements rollback functionality for NodeActionsScenarioPlugin

kaffie-1517:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization

## Description  
This PR implements rollback functionality for the `NodeActionsScenarioPlugin` to automatically restore cluster nodes to their original state when scenarios fail or are interrupted.

### Key Features

 1. **Rollback Registration**: Rollback callables are registered immediately after node actions complete, ensuring automatic restoration on failure or interruption.

 2. **State-Based Recovery**: Before attempting recovery, the implementation checks node Ready status to:

   - Skip unnecessary operations on already healthy nodes
   - Identify nodes that actually need recovery
   - Handle edge cases where nodes may be inaccessible

 3. **Multiple Recovery Strategies**: For docker/generic cloud types, the rollback uses a tiered recovery approach:

   - First attempts less invasive kubelet restart
   - Falls back to node start if kubelet restart fails
   - For other cloud providers, uses direct node start

 4. **Action-Specific Rollback**:

   -  **Actions with rollback enabled**: `node_stop_scenario`, `stop_kubelet_scenario`, `node_stop_start_scenario`, `stop_start_kubelet_scenario`, and other actions that leave nodes in recoverable bad states
   -  **Actions skipped**: `node_start_scenario`, `node_termination_scenario`, `node_reboot_scenario`

 5. **Cloud Provider Support**: Rollback is implemented for AWS, Azure, GCP, OpenStack, Docker, VMware/vSphere, IBM Cloud, and IBM Power cloud providers.

- Closes Node Outage Rollback Feature #916

## Documentation  

N/A

### DIFF
--- a/krkn/scenario_plugins/node_actions/node_actions_scenario_plugin.py
+++ b/krkn/scenario_plugins/node_actions/node_actions_scenario_plugin.py
@@ -225,11 +225,11 @@ class NodeActionsScenarioPlugin(AbstractScenarioPlugin):
             cloud_type = get_yaml_item_value(node_scenario, "cloud_type", "generic")
             
             # Skip rollback for actions that complete successfully and don't leave nodes in bad state
+            # Note: node_stop_scenario is NOT skipped - if interrupted, nodes are stopped and need restoration
             skip_rollback_actions = [
-                "node_start_scenario",      
-                "node_stop_scenario",       
-                "node_termination_scenario", 
-                "node_reboot_scenario",      
+                "node_start_scenario",      # Node is started - completes successfully, no rollback needed
+                "node_termination_scenario", # Node is terminated - can't rollback anyway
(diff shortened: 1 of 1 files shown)


tsebastiani (changes requested):
The handler registration needs to be placed where you're sure that is always registered, I've tried the scenario with `node_stop_start_scenario` action and the handler is not registered in the right moment. What I suggest you is to include the logic of the skipped rollback actions inside the handler, and *always* register it, also I wouldn't put it under any conditional statement.
The purpose of the rollback method is to rollback the cluster even if the scenario is suddenly interrupted, so the rollback script must be prepared *before* any action in the cluster is performed by the scenario, not after.

PR #1059, opened 2026-01-16
fix: use list comprehension to avoid skipping nodes during exclusion

### LINKED ISSUE #1058: Node exclusion skips nodes due to list modification during iteration
# Bug Description

## Describe the bug

In `node_actions_scenario_plugin.py` (lines 203-209), the code modifies a list while iterating over it, which causes some excluded nodes to be skipped.

**The problematic code:**
```python
for node in nodes:
    if node in exclude_nodes:
        logging.info(f"excluding node {node} with exclude label {exclude_labels}...")
        nodes.remove(node)
```

**The issue:** When `nodes.remove(node)` is called during iteration, the list is modified in place. This shifts all subsequent elements left by one index, but the iterator continues to the next index position, skipping the element that just moved into the current position.

**Result:** Consecutive nodes in `exclude_nodes` won't all be excluded - every other one gets skipped.

## To Reproduce

1. Configure node actions with `label_selector` that matches multiple nodes
2. Set `exclude_labels` to match 2+ consecutive nodes in the list
3. Run krkn with node actions scenario
4. Observe that not all excluded nodes are actually excluded

## Minimal reproduction script:
```python
nodes = ["node-A", "node-B", "node-C", "node-D"]
exclude_nodes = ["node-A", "node-B"]

for node in nodes:
    if node in exclude_nodes:
        nodes.remove(node)

print(nodes)  # ['node-B', 'node-C', 'node-D'] - node-B NOT excluded!
```

## Scenario File
```yaml
node_scenarios:
  - actions:
      - node_stop_start_scenario
    node_count: 1
    label_selector: node-role.kubernetes.io/worker
    timeout: 120
```

## Config File
```yaml
kraken:
  kubeconfig_path: ~/.kube/config
  exit_on_failure: False
  publish_kraken_status: True
  signal_state: RUN

tunings:
  wait_duration: 60
```

## Expected behavior

All nodes matching `exclude_labels` should be excluded from chaos testing. If nodes A and B both match the exclude criteria, neither should be targeted.

**Expected output:** `['node-C', 'node-D']`
**Actual output:** `['node-B', 'node-C', 'node-D']`

## Krkn Output

Only `node-A` will show the exclusion log message. `node-B` will silently remain in the target list and may be subjected to chaos actions despite being configured for exclusion.

### Reference https://stackoverflow.com/questions/6260089/strange-result-when-removing-item-from-a-list-while-iterating-over-it-in-python

AR21SM:
# Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization

## Description

Fixed list modification during iteration in node exclusion logic. The previous code called `nodes.remove(node)` while iterating over `nodes`, causing consecutive excluded nodes to be skipped.

**Before:** `['node-B', 'node-C', 'node-D']` (node-B not excluded)  
**After:** `['node-C', 'node-D']` (both excluded correctly)

- Related Issue #: 1058
- Closes #1058 

### Is documentation needed for this update?
[ ] Is documentation needed for this update?
No documentation needed - this is a bug fix with no API changes.

See [testing your changes](link) and run on any Kubernetes or OpenShift cluster to validate your changes

### REQUIRED:
```bash
python bug.py
...
=================================================================
REPRODUCING LIST MODIFICATION DURING ITERATION BUG
=================================================================

1. Running BUGGY code (nodes will be skipped)...
-----------------------------------------------------------------
   Initial nodes: ['node-A', 'node-B', 'node-C', 'node-D']
   Nodes to exclude: ['node-A', 'node-B']

INFO: excluding node node-A with exclude label ['node-A', 'node-B']

   Final nodes after exclusion: ['node-B', 'node-C', 'node-D']

    BUG CONFIRMED! node-B was NOT excluded!

2. Running FIXED code (works correctly)...
-----------------------------------------------------------------
   Initial nodes: ['node-A', 'node-B', 'node-C', 'node-D']
   Nodes to exclude: ['node-A', 'node-B']

INFO: excluding node node-A
INFO: excluding node node-B

   Final nodes after exclusion: ['node-C', 'node-D']

    CORRECT! Both node-A and node-B were excluded.

EXPECTED: ['node-C', 'node-D']
BUGGY:    ['node-B', 'node-C', 'node-D']
FIXED:    ['node-C', 'node-D']
```
```

### CODE REVIEW

File: krkn/scenario_plugins/node_actions/node_actions_scenario_plugin.py [hunk 1]
```diff
@@ -197,12 +197,12 @@ def inject_node_scenario(
                     exclude_label, 0, kubecli
                 )
 
-                for node in nodes:
-                    if node in exclude_nodes:
+                for node in exclude_nodes:
```
paigerube14: since this loop only does logging now and we do another loop below, I think we can remove this loop and just make one log statement, with the list of exclude_nodes

AR21SM: Done! Simplified to a single log statement with the full exclude_nodes list. Thanks for the review!

AR21SM:
@paigerube14

PR #1181, opened 2026-03-06
Add memory hog v2 pytest tests to CI/tests_v2

ddjain:
### Summary
Adds memory hog scenario to the CI/tests_v2 pytest framework, migrating coverage from the shell script CI/tests/test_memory_hog.sh.

### What’s included

- test_scenario_execution_success – Run scenario, assert exit 0 and no hog pods left.
- test_stress_pod_lifecycle – Background run; assert hog pod appears, goes Running, then is removed after Krkn exits.
- test_node_selector_targeting – Hog pod is scheduled on the node from node-selector.
- test_duration_configuration – Scenario run respects configured duration (elapsed time check).
- test_memory_size_configuration – Scenario runs successfully with memory-vm-bytes: 50%.
- test_invalid_node_selector_fails – Invalid node selector (no matching nodes) causes Krkn to fail.
- test_invalid_scenario_fails – Invalid scenario YAML causes Krkn to fail.


ddjain:
This is already done in separate PR

PR #1183, opened 2026-03-09
fix: use sorted() instead of sort() (#1182)

### LINKED ISSUE #1182: [BUG] `list.sort()` returns None causing Prometheus alert and metric key validation to be non functional
## **Describe the bug**

While reading through `krkn/prometheus/client.py`, I came across a subtle but interesting issue. In two separate places, the code uses `.sort()` on lists to compare dictionary keys but the thing is, `.sort()` in Python doesn't return the sorted list. It sorts in-place and returns `None`. So both comparisons end up being `None != None` or `None == None`, which completely defeats the purpose of those checks.

It's one of those easy-to-miss Python gotchas that can silently break things.

**First one, alert key validation on line 49:**

```python
if list(alert.keys()).sort() != ["expr", "description", "severity"].sort():
    logging.error(f"wrong alert {alert}, skipping")
    continue
```

Because both `.sort()` calls return `None`, this condition is always `False`. That means even if someone has a totally broken alert config (wrong keys, typos, whatever), it'll just sail right through without a single warning.

**Second one, metric query routing on lines 208–209:**

```python
elif (
    list(metric_query.keys()).sort()
    == ["query", "metricName"].sort()
):
```

Same thing here, but the other way around `None == None` is always `True`. So every non-instant metric query ends up in this branch regardless of what keys it actually has, and the `else` case that logs `"didn't match keys"` can never be reached.

### Scenario File

Nothing special needed, any scenario with Prometheus monitoring turned on will go through this code:

```yaml
- config:
    scenario_type: pod_scenarios
    scenarios:
      - scenarios/etcd.yml
```

### Config File

```yaml
kraken:
  chaos_scenarios:
    - pod_scenarios:
        - scenarios/etcd.yml

performance_monitoring:
  deploy_dashboards: False
  alert_profile: config/alerts_invalid.yaml
  metrics_profile: config/metrics.yaml
  prometheus_url: http://prometheus:9090
  prometheus_bearer_token: ""
  enable_alerts: True
  enable_metrics: True
  check_critical_alerts: False
```

The simplest way to see it is to make a bad alert profile on purpose (`alerts_invalid.yaml`):

```yaml
- wrong_key: "up == 0"
  another_bad_key: "test"
```

Then just run krkn with that config. You'd expect the logs to say something like `"wrong alert {...}, skipping"` but instead, the bad config slips through and you eventually get hit with a `KeyError` when the code tries to access `alert["expr"]` down the line. Not great.

## **Expected behavior**

Honestly, just what the original code clearly intended, alerts that don't have the right keys (`expr`, `description`, `severity`) should get flagged and skipped with a helpful log message. And for metrics, queries should be routed based on what keys they actually have, not just blindly funneled into one branch.

## **Krkn Output**

What the logs *should* show:

```
ERROR - wrong alert {'wrong_key': 'up == 0', 'another_bad_key': 'test'}, skipping
```

What you actually get:

```
KeyError: 'expr'
```

I also ran a quick test in the Python shell to make sure I wasn't imagining things:

```python
>>> [3, 1, 2].sort()
# returns None, not [1, 2, 3]

>>> list({"WRONG": 1}.keys()).sort() != ["expr", "description", "severity"].sort()
False  # yep, None != None is always False

>>> sorted({"WRONG": 1}.keys()) != sorted(["expr", "description", "severity"])
True   # this is what we actually want
```

## **Additional context**

This is a pretty well-known Python pitfall `.sort()` vs `sorted()`. Easy to mix up, easy to miss in code review. The fix is literally just swapping `.sort()` for `sorted()` in those two spots:

```python
# Line 49
if sorted(alert.keys()) != sorted(["expr", "description", "severity"]):

# Lines 208–209
elif sorted(metric_query.keys()) == sorted(["query", "metricName"]):
```

That's it, two lines, no behavior change other than making the validation actually work as intended. I'm happy to put up a PR for this if that works for you all!

Ivan-houzhiwen:
## Summary

Fixed Python pitfall in `krkn/prometheus/client.py`:

- `.sort()` returns `None`, not the sorted list
- Changed to `sorted()` for actual key comparison

### Before
```python
if list(alert.keys()).sort() != [...].sort():
```

### After
```python
if sorted(alert.keys()) != sorted([...]):
```

Fixes #1182

**PrimeFix Tech**

### DIFF
--- a/krkn/prometheus/client.py
+++ b/krkn/prometheus/client.py
@@ -46,7 +46,7 @@ def alerts(
             sys.exit(1)
 
         for alert in profile_yaml:
-            if list(alert.keys()).sort() != ["expr", "description", "severity"].sort():
+            if sorted(alert.keys()) != sorted(["expr", "description", "severity"]):
                 logging.error(f"wrong alert {alert}, skipping")
                 continue
 
@@ -205,7 +205,7 @@ def metrics(
                    query
                 )
             elif (
-                list(metric_query.keys()).sort()
+                sorted(metric_query.keys())
                 == ["query", "metricName"].sort()
             ):
                 metrics_result = prom_cli.process_prom_query_in_range(


krkn/prometheus/client.py
```diff
@@ -205,7 +205,7 @@ def metrics(
                    query
                 )
             elif (
-                list(metric_query.keys()).sort()
+                sorted(metric_query.keys())
                 == ["query", "metricName"].sort()
```
paigerube14, line 209: should it be sorted here to match if first half of the if statement?

paigerube14:
@Ivan-houzhiwen thanks so much for the help on this issue. One small change requested to align the 2 parts of the if statement. Could you please add some unit tests around this change if possible, TIA

Also please signoff your commit using the steps: https://github.com/krkn-chaos/krkn/pull/1183/checks?check_run_id=66259695604

PR #1219, opened 2026-03-31
fix(scenarios): fix network_chaos_ng variable shadowing and instance_count condition

NETIZEN-11:
# Description  

This PR fixes two bugs in `krkn/scenario_plugins/network_chaos_ng/network_chaos_ng_scenario_plugin.py`:

- Fixed incorrect condition `len(config) > 1`, which was checking the length of a dictionary instead of the scenario list. Updated to `len(scenario_config) > 1` so inter-scenario wait is applied only when multiple scenarios are configured  
- Corrected inverted `instance_count` guard from `instance_count > len(targets)` to `len(targets) > instance_count`, ensuring sampling only occurs when there are more targets than requested  

These fixes ensure correct scenario execution flow and prevent invalid sampling behavior.

---

## Type of change

- [ ] Refactor  
- [ ] New feature  
- [x] Bug fix  
- [ ] Optimization  

---

- Related Issue #:  #1208 (fix/scenario-plugins-and-utils) 
- Closes #: #1208

---

- [x] Ensure the changes and proposed solution have been discussed in the relevant issue  
- [ ] Added unit tests with above 80% coverage  

---

## Tests Performed

```bash
python run_kraken.py

# Output:
No runtime errors observed.
Scenario execution behaves correctly with proper delay handling.
Instance selection logic works as expected.

### DIFF
--- a/krkn/scenario_plugins/network_chaos/network_chaos_scenario_plugin.py
+++ b/krkn/scenario_plugins/network_chaos/network_chaos_scenario_plugin.py
@@ -192,6 +192,12 @@ class NetworkChaosScenarioPlugin(AbstractScenarioPlugin):
         pods_list = kubecli.list_pods(
             label_selector=pod_label_selector, namespace="default"
         )
+        if not pods_list:
+            raise Exception(
+                f"No pods found matching label selector '{pod_label_selector}' "
+                f"in namespace 'default'. The job pod may not have started or "
+                f"the label selector may be incorrect."
+            )
         return pods_list[0]
 
     # krkn_lib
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: krkn/scenario_plugins/network_chaos_ng/network_chaos_ng_scenario_plugin.py [hunk 1]
```diff
@@ -52,8 +68,9 @@ def run(
                         )
 
                     if (
-                        network_chaos_config.instance_count != 0
-                        and network_chaos_config.instance_count > len(targets)
+                        isinstance(network_chaos_config.instance_count, int)
```
paigerube14: do we need this check since the model should first check if its an int already?

NETIZEN-11: Good catch, thanks for pointing this out! Since BaseNetworkChaosConfig already defines `instance_count: int`, the isinstance check was indeed redundant. I've removed it and simplified the guard to `instance_count > 0` and `len(targets) > instance_count`. Really appreciate the review!

PR #1233, opened 2026-04-17
feat(ci): add squash-commits bot workflow

paigerube14:
## Summary

- Adds a workflow that detects PRs with more than one commit and labels them `squash-commits`
- Posts a comment with two methods for squashing (interactive rebase or soft reset)
- Updates the comment with the current commit count on each new push
- Removes the label and comment automatically once the PR is down to a single commit

## How it works

**Trigger:** `pull_request_target` on open/sync/reopen only — no push-to-main needed since commit count is only affected by the PR itself

**Per-PR logic:**
- Paginates `pulls.listCommits` to get the exact commit count
- If count > 1 → adds `squash-commits` label + posts comment (or updates existing comment with new count)
- If count == 1 → removes label and deletes comment automatically
- Creates the `squash-commits` label on first run if it doesn't exist

## Test plan
- [ ] Open a PR with multiple commits — label and comment should appear with correct count
- [ ] Push another commit — comment should update with the new count
- [ ] Squash down to one commit and force-push — label and comment should be removed automatically

### DIFF
--- a/.github/workflows/squash-commits.yml
+++ b/.github/workflows/squash-commits.yml
@@ -1,0 +1,156 @@
+name: Squash Commits
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+
+permissions:
+  pull-requests: write
+  issues: write
+  contents: read
+
+jobs:
+  squash-commits:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


paigerube14:
closing as I think this would be too noisy

PR #1250, opened 2026-04-25
feat: implement Time to Recovery (TTR) measurement in HealthChecker

### LINKED ISSUE #1235: Calculate and report "Time to Recovery" (TTR) for chaos scenarios
Hi Team,
While I was setting up and testing krkn locally to get familiar with the codebase I was running some basic pod-kill scenarios to see how my test cluster would react. I noticed that while it was doing  a great job of verifying if the system recovered it doesn't tell me how long that recovery took. For example, I wanted to see if my deployment took 5 seconds or 45 seconds to stabilize after the pods were deleted. Right now, to figure that out, I have to manually scroll through the console output and do math based on the log timestamps.

**Describe the solution you'd like**
I propose adding an automated TTR calculation directly into the scenario execution flow.Once the chaos injection Krkn should grab a timestamp Once the post chaos health checks return a healthy status it grabs a second timestamp hould then calculate the difference and explicitly log the Time to Recovery as well as include it in the final telemetry payload

How I can fix it:
If assigned to this issue my plan of action is:
1)Modify the Scenario Lifecycles: In the core scenario scripts I will introduce variables to capture chaos end time and recovery success tme 
2)Simply calculate TTR = recovery success time - chaos end time
3)If the health checks fail and the system never recovers within the configured window, the TTR should be logged as Failed/Timeout

Aditya200247:
Resolves #1235. This implements the TTR timer in the HealthChecker. Note: This PR requires the telemetry model update submitted in krkn-lib PR [#266](https://github.com/krkn-chaos/krkn-lib/pull/266)

### DIFF
--- a/krkn/utils/HealthChecker.py
+++ b/krkn/utils/HealthChecker.py
@@ -24,6 +24,7 @@ from krkn_lib.models.telemetry.models import HealthCheck
 class HealthChecker:
     current_iterations: int = 0
     ret_value = 0
+    time_to_recovery = None
     def __init__(self, iterations):
         self.iterations = iterations
 
@@ -76,6 +77,15 @@ class HealthChecker:
                                 start_timestamp = health_check_tracker[config["url"]]["start_timestamp"]
                                 previous_status_code = str(health_check_tracker[config["url"]]["status_code"])
                                 duration = (end_timestamp - start_timestamp).total_seconds()
+                                
+                                if response["status_code"] == 200:
+                                    ttr_seconds = round(duration, 2)
+                                    if self.time_to_recovery != -1:
+                                        if self.time_to_recovery is None:
+                                            self.time_to_recovery = ttr_seconds
+                                        else:
+                                            self.time_to_recovery = max(self.time_to_recovery, ttr_seconds)
+
                                 change_record = {
                                     "url": config["url"],
                                     "status": False,
@@ -101,6 +111,13 @@ class HealthChecker:
                     "duration": duration
                 }
                 health_check_telemetry.append(HealthCheck(success_response))
+                
+                if health_check_tracker[url]["status_code"] != 200:
+                    logging.warning(f"Health check timed out before system recovered for URL:%s" % url)
+                    self.time_to_recovery = -1
+
+            if self.time_to_recovery is not None and self.time_to_recovery != -1:
+                logging.info(f"All health checks recovered. Maximum Time to Recovery (Worst-Case TTR): %ss" % self.time_to_recovery)
 
             health_check_telemetry_queue.put(health_check_telemetry)
         else:
--- a/krkn/utils/HealthChecker.py
+++ b/krkn/utils/HealthChecker.py
@@ -113,11 +113,11 @@ class HealthChecker:
                 health_check_telemetry.append(HealthCheck(success_response))
                 
                 if health_check_tracker[url]["status_code"] != 200:
-                    logging.warning(f"Health check timed out before system recovered for URL:%s" % url)
+                    logging.warning(f"Health check timed out before system recovered for URL: {url}")
                     self.time_to_recovery = -1
 
             if self.time_to_recovery is not None and self.time_to_recovery != -1:
-                logging.info(f"All health checks recovered. Maximum Time to Recovery (Worst-Case TTR): %ss" % self.time_to_recovery)
+                logging.info(f"All health checks recovered. Maximum Time to Recovery (Worst-Case TTR): {self.time_to_recovery}s")
 
             health_check_telemetry_queue.put(health_check_telemetry)
         else:
--- a/tests/test_health_checker.py
+++ b/tests/test_health_checker.py
@@ -13,7 +13,7 @@ This test file provides comprehensive coverage for the main functionality of Hea
 Usage:
     python -m coverage run -a -m unittest tests/test_health_checker.py -v
 
-Assisted By: Claude Code
+
 """
 
 import queue
--- a/tests/test_health_checker.py
+++ b/tests/test_health_checker.py
@@ -498,6 +498,108 @@ class TestHealthChecker(unittest.TestCase):
         # Verify sleep was called with custom interval
         mock_sleep.assert_called_with(5)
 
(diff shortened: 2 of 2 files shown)


krkn/utils/HealthChecker.py
```diff
@@ -76,6 +77,9 @@ def run_health_check(self, health_check_config, health_check_telemetry_queue: qu
                                 start_timestamp = health_check_tracker[config["url"]]["start_timestamp"]
                                 previous_status_code = str(health_check_tracker[config["url"]]["status_code"])
                                 duration = (end_timestamp - start_timestamp).total_seconds()
+                                ttr_seconds = round(duration, 2)
```
paigerube14: I'm not sure of the value here when it is the same as the duration that is already printed

krkn/utils/HealthChecker.py
```diff
@@ -101,6 +105,10 @@ def run_health_check(self, health_check_config, health_check_telemetry_queue: qu
                     "duration": duration
                 }
                 health_check_telemetry.append(HealthCheck(success_response))
+                
+                if health_check_tracker[url]["status_code"] != 200:
+                    logging.warning("Health check timed out before system recovered.")
+                    self.time_to_recovery = "TIMEOUT"
```
paigerube14: there can be multiple health checker urls defined. I'm not sure if having one self.time_to_recovery will work here

krkn/utils/HealthChecker.py
yogananth-subramanian: agree, and how about only calculating  TTR if  status changes only to status 200 and not any other status?

krkn/utils/HealthChecker.py
yogananth-subramanian: agree,  how about reporting  maximum TTR (worst-case recovery time), when multiple URLs are involved.

Aditya200247:
Thanks @paigerube14 and @yogananth-subramanian! That makes perfect sense. I will update the logic to only calculate TTR when the status transitions to 200, and I'll remove the redundant print statement. I will also implement a running maximum to capture the worst-case TTR across all URLs, and ensure that if any single URL fails to recover, the final TTR reports as TIMEOUT. I'll push these changes shortly!

krkn/utils/HealthChecker.py
```diff
@@ -101,6 +111,13 @@ def run_health_check(self, health_check_config, health_check_telemetry_queue: qu
                     "duration": duration
                 }
                 health_check_telemetry.append(HealthCheck(success_response))
+                
+                if health_check_tracker[url]["status_code"] != 200:
+                    logging.warning(f"Health check timed out before system recovered for URL:%s" % url)
+                    self.time_to_recovery = "TIMEOUT"
```
paigerube14: for being able to compare/track recovery times over multiple different runs, it might be helpful to have this be -1 or some sort of float instead of a string

Aditya200247:
Will do! I'm fetching the latest upstream, squashing my commits, and adding the DCO sign-off right now. I'll force-push the updated branch shortly.

Aditya200247:
@paigerube14 @yogananth-subramanian All requested updates are complete! Let me know if everything looks good to go.
Thanks again for your guidance!

krkn/utils/HealthChecker.py
```diff
@@ -101,6 +111,13 @@ def run_health_check(self, health_check_config, health_check_telemetry_queue: qu
                     "duration": duration
                 }
                 health_check_telemetry.append(HealthCheck(success_response))
+                
+                if health_check_tracker[url]["status_code"] != 200:
+                    logging.warning(f"Health check timed out before system recovered for URL:%s" % url)
```
yogananth-subramanian: logging.warning(f"Health check timed out before system recovered for URL:{url}")

krkn/utils/HealthChecker.py
```diff
@@ -101,6 +111,13 @@ def run_health_check(self, health_check_config, health_check_telemetry_queue: qu
...
+                    self.time_to_recovery = -1
+
+            if self.time_to_recovery is not None and self.time_to_recovery != -1:
+                logging.info(f"All health checks recovered. Maximum Time to Recovery (Worst-Case TTR): %ss" % self.time_to_recovery)
```
yogananth-subramanian: logging.info(f"All health checks recovered. Maximum Time to Recovery (Worst-Case TTR): {self.time_to_recovery}s")

yogananth-subramanian: @Aditya200247 tests need to be added for the new TTR feature
Test TTR calculation when a single URL fails and recovers ,tracking the maximum across multiple URLs,    tracking the maximum across multiple URLs etc.

Aditya200247: @yogananth-subramanian Thanks for the feedback! I've updated the logging statements to use modern f-strings as requested. I have also added new unit tests to test_health_checker.py that specifically cover the new TTR logic, including:

TTR calculation when a single URL fails and recovers.

Timeout handling (-1) when a URL never recovers.

Tracking the maximum worst-case TTR across multiple URLs recovering at different speeds.

Let me know if these test cases look good to you

Aditya200247:
Hello @paigerube14 and @yogananth-subramanian
I just wanted to gently follow up on this PR All the requested updates have been implemented, including adding the new unit tests for the TTR logic and updating the logging statements to use f-strings Please let me know if you need any further adjustments Thank you again for your time and guidance.

tests/test_health_checker.py
```diff
@@ -0,0 +1,605 @@
...
+                },
+            ],
+            "interval": 0.01,
+        }
+
+        self.checker.iterations = 5
+        self.checker.run_health_check(config, self.health_check_queue)
+
+        self.assertIsNotNone(self.checker.time_to_recovery)
+        self.assertNotEqual(self.checker.time_to_recovery, -1)
```
yogananth-subramanian, line 601: Current code doesn't strongly prove the max logic works correctly. May be mock datetime.now() to return controlled timestamps.Create meaningfully different recovery times (e.g., 2s vs 5s) and assert the exact TTR value matches the longer recovery time.

PR #1217, opened 2026-03-31
fix(cerberus): fix call signature, config key typo, and add legacy key fallback

NETIZEN-11:
# Description  

This PR fixes multiple issues in `krkn/cerberus/setup.py` and improves test coverage:

- Remove stale `cerberus_url` positional argument from `application_status()` call to fix runtime TypeError  
- Fix shadowed local variable `check_application_routes` in `get_status()` so route checks work correctly  
- Read correct key `check_application_routes` first and fall back to legacy misspelled key `check_applicaton_routes` with a deprecation warning for backward compatibility  
- Simplify `publish_kraken_status` logic to exit only when the cluster is unhealthy (previous logic incorrectly exited when healthy)  
- Add Apache 2.0 license header to `setup.py` and corresponding test file  
- Update tests to cover legacy key fallback and corrected healthy-cluster behavior  

This PR is scoped to cerberus setup fixes and tests, making it easier to review and validate.

---

## Type of change

- [ ] Refactor  
- [ ] New feature  
- [x] Bug fix  
- [ ] Optimization  

---

## Related Tickets & Documents

- Related Issue #:  #1208 (fix/scenario-plugins-and-utils) 
- Closes #: #1208

## Checklist before requesting a review

- [x] Ensure the changes and proposed solution have been discussed  
- [x] I have performed a self-review of my code by running krkn and specific scenario  
- [ ] Added unit tests with above 80% coverage  

---

## Tests Performed

```bash
python run_kraken.py

# Output:
No runtime errors observed.
Cerberus setup correctly handles application status, route checks, and exit logic.
Legacy config fallback works as expected.

PR #1253, opened 2026-04-25
 fix(cerberus): fix call signature, config key typo, and add legacy key fallback

NETIZEN-11:
## Summary
  - Remove stale `cerberus_url` positional argument from `application_status()` call to fix runtime TypeError
  - Fix shadowed local variable `check_application_routes` in `get_status()` so route checks work correctly
  - Read correct config key `check_application_routes` first and fall back to legacy misspelled
  `check_applicaton_routes` with a deprecation warning
  - Only exit on failure when `exit_on_failure: true` AND cluster is unhealthy
  - Fix `publish_kraken_status()` incorrect else block that would exit when cluster is healthy

  ## Related Issue
  Fixes the issues identified in #1217


paigerube14:
this pr looks to have both cerberus changes but also node rollback, lets make sure there's only 1 related change in each PR. Thanks

krkn/scenario_plugins/node_actions/node_actions_scenario_plugin.py
```diff
@@ -41,12 +48,20 @@
...
-     ibmcloud_power_node_scenarios,
+    ibmcloud_power_node_scenarios,
 )
+
 node_general = False
 
+# Maps each reversible action to its compensating (inverse) action.
+# Only these actions register a rollback entry; irreversible actions
+# (terminate, crash, reboot, disk-detach) are intentionally excluded.
+REVERSIBLE_ACTIONS = {
```
paigerube14, line 60: lets remove this code so that each PR has one related change per PR, thanks

PR #1261, opened 2026-05-01
Skip cluster metadata collection when telemetry is disabled

kdelee:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [x] Optimization

## Description

`collect_cluster_metadata` runs unconditionally after every scenario, even when `telemetry.enabled: false`. On large OpenShift clusters (e.g. ROSA with many namespaces), this takes up to 15 minutes because it:

- Lists all Pods, Secrets, ConfigMaps, Routes, and Builds across the entire cluster via `get_all_kubernetes_object_count`
- Queries `get_nodes_infos()` for every node
- Attempts to list VirtualMachineInstances (hangs when KubeVirt is not installed)

None of this data is used when telemetry is disabled — it's only needed for the telemetry payload sent to the S3 backend.

This change guards `collect_cluster_metadata` with a `config["telemetry"].get("enabled", False)` check so it only runs when the data will actually be sent.

N/A — discovered during chaos testing on ROSA clusters with `telemetry.enabled: false`.

## Documentation

- [ ] Yes
- [x] No

No documentation changes needed — this is an internal optimization that doesn't change any user-facing behavior or configuration.

- [x] I have discussed and acknowledged this PR per our contributing guidelines
- [x] I have performed a self-review by running the application
- [ ] I have added unit tests with 80%+ coverage for core features
- [x] I have described the test combinations I performed with output results

### Test results

Before fix — `telemetry.enabled: false`, simple no-op scenario:
```
23:24:25 [INFO] Evaluating 6 SLOs over window ...
23:24:30 [INFO] collecting OCP cluster metadata, this may take few minutes....
# ... hangs for 15+ minutes collecting metadata that is never sent ...
23:42:03 [INFO] telemetry collection disabled, skipping.
23:42:05 [INFO] Successfully finished running Kraken.
```

After fix — same scenario:
```
23:24:25 [INFO] Evaluating 6 SLOs over window ...
23:24:26 [INFO] telemetry disabled, skipping cluster metadata collection
23:24:26 [INFO] telemetry collection disabled, skipping.
23:24:27 [INFO] Successfully finished running Kraken.
```

Total runtime reduced from ~18 minutes to ~30 seconds for a no-op test scenario.

### DIFF
--- a/run_kraken.py
+++ b/run_kraken.py
@@ -497,17 +497,19 @@ def main(options, command: Optional[str]) -> int:
         
         post_kubevirt_check = kubevirt_checker.gather_post_virt_checks(kubevirt_check_telem)
         chaos_telemetry.post_virt_checks = post_kubevirt_check
-        # if platform is openshift will be collected
-        # Cloud platform and network plugins metadata
-        # through OCP specific APIs
-        if distribution == "openshift":
-            logging.info(
-                "collecting OCP cluster metadata, this may take few minutes...."
-            )
-            telemetry_ocp.collect_cluster_metadata(chaos_telemetry)
+        # Collect cluster metadata only when telemetry is enabled
+        # (listing all k8s objects is very slow on large clusters)
+        if config["telemetry"].get("enabled", True):
+            if distribution == "openshift":
+                logging.info(
+                    "collecting OCP cluster metadata, this may take few minutes...."
+                )
+                telemetry_ocp.collect_cluster_metadata(chaos_telemetry)
+            else:
+                logging.info("collecting Kubernetes cluster metadata....")
+                telemetry_k8s.collect_cluster_metadata(chaos_telemetry)
         else:
-            logging.info("collecting Kubernetes cluster metadata....")
-            telemetry_k8s.collect_cluster_metadata(chaos_telemetry)
+            logging.info("telemetry disabled, skipping cluster metadata collection")
 
         # Collect error logs from handler
         error_logs = error_collection_handler.get_error_logs()


paigerube14:
@kdelee thanks so much for this PR. The telemetry config parameter that you're changing here, I thought was to save (or not) the telemetry details in S3 for long term storage. Better explanation of https://krkn-chaos.dev/docs/krkn/telemetry/
But taking a second look I think it should turn on or off the collection of those details as well 

I think we'll need to work on better reducing the time it takes for the details to be printed. If we don't have those details it could be unclear of the overall status of the run and what ran.

paigerube14 (approved):
/lgtm

Think we should look into reducing collection time in future, opened this issue: https://github.com/krkn-chaos/krkn-lib/issues/267

run_kraken.py
```diff
@@ -497,17 +497,19 @@ def main(options, command: Optional[str]) -> int:
...
-        # Cloud platform and network plugins metadata
-        # through OCP specific APIs
-        if distribution == "openshift":
-            logging.info(
-                "collecting OCP cluster metadata, this may take few minutes...."
-            )
-            telemetry_ocp.collect_cluster_metadata(chaos_telemetry)
+        # Collect cluster metadata only when telemetry is enabled
+        # (listing all k8s objects is very slow on large clusters)
+        if config["telemetry"].get("enabled", False):
```
paigerube14: think lets maybe actually set this default to true so a base run can show the json output at the end but you'll be able to set to false in your runs. Thoughts ?

paigerube14:
@kdelee are you able to change the default value and signoff on your commit and I think we can get this merged, TIA

kdelee:
Thanks @paigerube14! Updated the default to `True` and added sign-off.

Re: the Qodo concern about Elasticsearch missing cluster metadata — in our case we use ES for resiliency scores and SLO results, not cluster metadata. We don't have anything consuming that field from ES. Anyone who does need it can leave `telemetry.enabled` at the default (`true`). The skip only kicks in when you explicitly opt out.

paigerube14 (approved):
/lgtm

kdelee:
@paigerube14 thanks for review! If it looks good to you, please merge -- doesn't appear to be option for me

paigerube14:
> @paigerube14 thanks for review! If it looks good to you, please merge -- doesn't appear to be option for me

will do! waiting for test to finish, but will merge after

PR #1265, opened 2026-05-04
fix: replace bare except with Exception handler in Kubernetes client initialization

### LINKED ISSUE #1264: fix: bare except clause causes NameError during Kubernetes client initialization failure in run_kraken.py
## Description

In `run_kraken.py`, the Kubernetes client initialization block uses a bare `except:` clause that causes a secondary `NameError` crash when the primary initialization fails, completely masking the original error.

## Location

- **File:** `run_kraken.py`
- **Lines:** 224–231

## Current Behavior

When the Kubernetes client initialization fails at line 228 (before `kubecli` is assigned), the bare `except:` handler on line 230 immediately attempts to call a method on `kubecli` — which is not yet defined. This raises a `NameError`, masking the original exception entirely.

Additionally, the code following the try/except block references `ocpcli` unconditionally, which is also undefined if the exception occurred before line 229.

## Impact

- Any failure during Kubernetes client initialization causes a `NameError` crash instead of a meaningful error message
- The bare `except:` also catches `SystemExit` and `KeyboardInterrupt`, preventing clean shutdowns
- The original root cause (e.g., invalid kubeconfig path) is completely hidden from the user
- Violates PEP 8 and Python best practices for exception handling

## Proposed Fix

- Replace bare `except:` with `except Exception as e`
- Log the actual initialization error for visibility
- Initialize both `kubecli` and `ocpcli` with `None` as fallback so subsequent code does not crash with `NameError`

parth-agrawall:
## Description

The Kubernetes client initialization block in `run_kraken.py` used a bare `except:` clause that caused a `NameError` crash whenever initialization failed, completely hiding the original error from the user.

## Root Cause

If `KrknKubernetes()` raised an exception before `kubecli` was assigned, the except handler immediately tried to call `kubecli.initialize_clients(None)` on an undefined variable. This replaced the real error with an unrelated `NameError`. The bare `except:` also unintentionally caught `SystemExit` and `KeyboardInterrupt`.

## Changes

- `run_kraken.py`: Replaced bare `except:` with `except Exception as e`
- Added `logging.error()` to surface the actual initialization failure
- Initialized both `kubecli` and `ocpcli` with `kubeconfig_path=None` as fallback so subsequent code referencing these variables does not crash with `NameError`

## Testing

- When kubeconfig initialization fails, the error is now logged clearly
- `kubecli` and `ocpcli` are always defined after the try/except block
- `SystemExit` and `KeyboardInterrupt` are no longer swallowed

## Related Issue

Closes #1264

---

**Signed-off-by:** Parth Agrawal <parth.agrawal4002@gmail.com>


yogananth-subramanian (approved):
LGTM!

PR #1272, opened 2026-05-06
added the badge and the workflow for scorecard badge

### LINKED ISSUE #1271: Adding openssf scorecard badge in readme
Image

On CLOMonitor, openssf scorecard badge check is failing.

antedotee:
# Description  
I have added the yaml file from the workflow file template mentioned in openssf repo (https://github.com/ossf/scorecard/blob/main/.github/workflows/scorecard-analysis.yml). To make sure my implementation is well sound, I have read the same from the cloudnative-pg project.
Also, I have added the badge within the same PR to prevent raising one unnecessary PR. In case seperation of concerns in required, I can raise that in another PR.

- Closes #1271

### DIFF
--- a/.github/workflows/scorecard-analysis.yml
+++ b/.github/workflows/scorecard-analysis.yml
@@ -1,0 +1,60 @@
+name: Scorecard analysis workflow
+on:
+  workflow_dispatch: # For Branch-Protection check. Only the default branch is supported. See
+
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
+  branch_protection_rule: # To guarantee Maintained check is occasionally updated. See
+
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
+  schedule:
+    - cron: "17 9 * * 1"
+
+permissions: read-all
+
+jobs:
+  analysis:
+    name: Scorecard analysis
+    runs-on: ubuntu-latest
+    # `publish_results: true` only works when run from the default branch. conditional can be removed if disabled.
+    if: github.event.repository.default_branch == github.ref_name ||
+      github.event_name == 'pull_request'
+    permissions:
+      # Needed for Code scanning upload
+      security-events: write
+      # Needed for GitHub OIDC token if publish_results is true
+      id-token: write
+
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: "Run analysis"
+        uses: ossf/scorecard-action@4eaacf0543bb3f2c246792bd56e8cdeffafb205a # v2.4.3
+        with:
+          results_file: results.sarif
+          results_format: sarif
+          # Scorecard team runs a weekly scan of public GitHub repos,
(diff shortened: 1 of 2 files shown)


.github/workflows/scorecard-analysis.yml
```diff
@@ -0,0 +1,62 @@
+name: Scorecard analysis workflow
+on:
+  workflow_dispatch: # For Branch-Protection check. Only the default branch is supported. See
+
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
+  branch_protection_rule: # To guarantee Maintained check is occasionally updated. See
+
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
+  schedule:
+    - cron: "17 9 * * 1"
```
paigerube14, line 10: wondering if do we want this on a schedule or just on a push to main?

antedotee: I think scheduling is better as we dont want to run it unnecessarily and almost all of the workflows that I have read(around 4) have scheduled it.It won't be a problem if the scorecard is updated daily instead of every push,I think. I can change this to push if you want also. What are your thoughts?

.github/workflows/scorecard-analysis.yml
```diff
@@ -0,0 +1,62 @@
...
+  push:
```
paigerube14: agree, we can run periodic but this will run on pushes to main

antedotee: I was going through the documentation and reached upon this answer by chatting with claude:

Scorecard's Maintained check looks at recent repo activity (commits, issues) over a rolling window. The OpenSSF infrastructure that publishes your score at api.scorecard.dev and on the badge consumes the SARIF you upload. If the repo has a quiet week with no pushes to main, the published score goes stale and the Maintained signal can drift. The weekly cron is the canonical OpenSSF-recommended schedule precisely to keep the badge fresh independently of commit cadence.


What do you think? Should I change anything😅?

paigerube14: yes, think we should use the cron schedule instead of the push to main branch here. Let's remove lines 11-13

antedotee: Done. Also, I had some generic doubts about the project, not related specifically to any project but general doubts. Can I ask you in the slack if you are free currently? 😅 😅

paigerube14: @antedotee sure can ping on slack, I'm about to end my day but will get back to you when I can

antedotee: Tysm :)

paigerube14:
@antedotee are you able to squash your commits into one and rebase your PR with upstream/main, think we can merge after

antedotee:
I just thought of something. Can this be helpful in our org, what do you think? https://github.com/marketplace/actions/automatic-rebase
I remember it adding it for one of the organizations earlier where I was contributing.

If it is no significant value, ignore this message 😅 .
Hope you have a great day after work.

paigerube14:
@antedotee can you take a look at this, the score card is getting invalid path image

antedotee:
Yes, I was about to tell you, this workflow runs every monday. Changing it to daily to fix this. Glad you saw this 😅 .

PR #1273, opened 2026-05-07
fix: fail node actions when actions list is empty

1PoPTRoN:
# Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization

# Description

This PR fixes a case where node chaos could silently do nothing.

If a node scenario was configured with `actions: []`, `NodeActionsScenarioPlugin` would skip the action loop and still return success. That meant Krkn could report a passing run even though no node chaos was actually injected.

With this change, an empty or missing `actions` field is treated as an invalid scenario config. The plugin now logs a warning and returns `1` before creating any node scenario object.

I also added a regression test to make sure this does not slip back in.

# Documentation

- [ ] Ensure the changes and proposed solution have been discussed in the relevant issue and have received acknowledgment from the community or maintainers. See [contributing guidelines](https://krkn-chaos.dev/docs/contribution-guidelines/)
- [ ] See [testing your changes](https://krkn-chaos.dev/docs/developers-guide/testing-changes/) and run on any Kubernetes or OpenShift cluster to validate your changes

*REQUIRED*:

```bash
python -m unittest tests.test_node_actions_scenario_plugin -v
...
Ran 44 tests in 0.160s

OK
```

### DIFF
--- a/krkn/scenario_plugins/node_actions/node_actions_scenario_plugin.py
+++ b/krkn/scenario_plugins/node_actions/node_actions_scenario_plugin.py
@@ -58,23 +58,29 @@ class NodeActionsScenarioPlugin(AbstractScenarioPlugin):
     ) -> int:
         with open(scenario, "r") as f:
             node_scenario_config = yaml.safe_load(f)
-            for node_scenario in node_scenario_config["node_scenarios"]:
+            for index, node_scenario in enumerate(node_scenario_config["node_scenarios"]):
                 try:
+                    actions = node_scenario.get("actions")
+                    if not actions:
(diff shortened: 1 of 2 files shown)


1PoPTRoN:
Hi @paigerube14 @chaitanyaenr , I fixed the node actions false-positive case where `actions: []` silently skipped chaos and still returned success. Added a regression test as well; the Python test checks are passing, while the current Docker/FOSSA failures look unrelated to this change. Looking forward if any changes needed.

paigerube14 (approved):
/lgtm

PR #1291, opened 2026-05-11
implemented the fix for service hijacking along with modifying tests

### LINKED ISSUE #1230: [BUG] Service Hijacking doesn't reset label on pod
# Bug Description

## **Describe the bug**
After service hijacking is ran, the labels on the pod that was effected is not reset to the original label that was on the pod. We should reset the pod labels as it was before the test so we are able to rerun easily and do a clean cleanup after chaos

## **To Reproduce**

Any specific steps used to reproduce the behavior

### Scenario File
Scenario file(s) that were specified in your config file (can be starred (*) with confidential information )
```yaml
# these settings are changed by the service_hijacking.sh script
service_target_port: 80
service_name: nginx
service_namespace: default
image: quay.io/krkn-chaos/krkn-service-hijacking:v0.1.3
chaos_duration: 60
# these settings are changed by the service_hijacking.sh script

plan:
  - resource: "/"
    steps:
      GET:
        - duration: 15
          status: 200
          mime_type: "application/json"
          payload: |
            {
              "status":"First Step: Service Hijacked by Krkn"
            }
        - duration: 15
          status: 401
          mime_type: "application/json"
          payload: |
            {
              "status":"Second step unauthorized by Krkn"
            }

```

### Config File 
Config file you used when error was seen (the default used is config/config.yaml)

```yaml
- service_hijacking_scenarios:
              - scenarios/kube/service_hijacking.yaml

```

## **Expected behavior**

Krkn cleans up labels after service hijackig is ran

## **Krkn Output**
Krkn output was successful

## **Additional context**

```
kind: Service
apiVersion: v1
metadata:
  name: nginx
  namespace: default
  uid: 0f***
  resourceVersion: '286908'
  creationTimestamp: '2026-04-10T17:42:11Z'
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"Service","metadata":{"annotations":{},"name":"nginx","namespace":"default"},"spec":{"ports":[{"name":"http","port":80,"targetPort":80}],"selector":{"app":"nginx"}}}
spec:
  clusterIP: 0.0.0
  ipFamilies:
    - IPv4
  ports:
    - name: http
      protocol: TCP
      port: 80
      targetPort: 80
  internalTrafficPolicy: Cluster
  clusterIPs:
    - 0.0.0
  type: ClusterIP
  ipFamilyPolicy: SingleStack
  sessionAffinity: None
  selector:
    service-hijacking: sh-pzbiv
status:
  loadBalancer: {}
```

antedotee:
## Description

Fixes #1230.

The service hijacking scenario was leaving residue in the target namespace after a chaos run completed. While the existing rollback logic correctly restored the original service selectors and deleted the hijacker pod, the ConfigMap that `deploy_service_hijacking` creates alongside the pod (which holds the encoded response plan mounted into the hijacker container) was never cleaned up. Every run therefore orphaned a ConfigMap, which interfered with rerunning the same scenario back to back and left the namespace in a dirty state.

This change captures the ConfigMap name returned by `deploy_service_hijacking` at the point the hijacker is deployed and persists it inside the base64 encoded rollback payload alongside the existing pod name and original selectors. During rollback, the plugin reads that field and deletes the matching namespaced ConfigMap through the Kubernetes client, mirroring how the hijacker pod is already torn down. The deletion is wrapped in a `try`/`except` so a missing or already removed ConfigMap is logged as a warning rather than aborting the rest of the rollback. The new field is also read with `.get()` so any rollback data persisted by an older build of Krkn (where the key does not exist) still drives a complete rollback of the selector and the pod without raising.

## Verification

To confirm the fix end to end I deployed a standard `nginx` Deployment and Service into the `default` namespace of a kind cluster, then ran the service hijacking scenario using `scenarios/kube/service_hijacking.yaml` with the configuration shown in the issue. During the run I observed that the service selector switched to the hijacker selector and `curl` against the service returned the configured payloads for the duration of the chaos. After the scenario completed I inspected the namespace with `kubectl get svc nginx -o yaml`, `kubectl get pods` and `kubectl get configmaps`. The service selector was back to `app: nginx`, the hijacker pod was gone, and the previously orphaned hijacker ConfigMap was no longer present. Repeating the run immediately afterwards completed cleanly without colliding with leftover state from the previous run, which was the original failure mode described in the issue.

## Tests

The unit tests in `tests/test_service_hijacking_scenario_plugin.py` were updated so that the mocked webservice object exposes a `config_map_name` attribute everywhere `deploy_service_hijacking` is mocked. The existing rollback test now asserts that `delete_namespaced_config_map` is called with the expected name and namespace. A new test, `test_rollback_service_hijacking_legacy_data_without_configmap`, encodes a rollback payload that omits the new field and verifies that rollback still restores the selector and deletes the pod while leaving the ConfigMap deletion path untouched, which guards the backwards compatibility branch on the `.get()` read. All tests pass under `python -m unittest discover -s tests -v`.

In case of screenshots needed, please let me know 😅 .

### DIFF
--- a/krkn/scenario_plugins/service_hijacking/service_hijacking_scenario_plugin.py
+++ b/krkn/scenario_plugins/service_hijacking/service_hijacking_scenario_plugin.py
@@ -101,6 +101,7 @@ class ServiceHijackingScenarioPlugin(AbstractScenarioPlugin):
                 "service_namespace": service_namespace,
                 "original_selectors": original_service["spec"]["selector"],
                 "webservice_pod_name": webservice.pod_name,
+                "webservice_config_map_name": webservice.config_map_name,
             }
             json_str = json.dumps(rollback_data)
             encoded_data = base64.b64encode(json_str.encode("utf-8")).decode("utf-8")
@@ -161,6 +162,7 @@ class ServiceHijackingScenarioPlugin(AbstractScenarioPlugin):
             service_namespace = rollback_data["service_namespace"]
             original_selectors = rollback_data["original_selectors"]
             webservice_pod_name = rollback_data["webservice_pod_name"]
+            webservice_config_map_name = rollback_data.get("webservice_config_map_name")
             
             logging.info(
                 f"Rolling back service hijacking: restoring service {service_name} in namespace {service_namespace}"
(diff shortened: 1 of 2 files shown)


antedotee:
CC @paigerube14 . Please have a look when you are free.

yogananth-subramanian (approved):
LGTM!

antedotee:
Sorry for tagging, but in case you have missed it, Please take a look @paigerube14.

antedotee:
CC @paigerube14 Please have a look at it when free, its been pending for a while 😅😅.

PR #1297, opened 2026-05-12
feat: add wait_until condition support for chaos scenarios

Avilir:
## Summary
- Adds an optional `wait_until` block to scenario configs that polls a user-provided condition (Python function or shell script) after chaos injection and before result collection
- Allows waiting for system recovery (e.g., pods ready, endpoints healthy) instead of relying on a fixed sleep duration
- Integrates into `AbstractScenarioPlugin` between rollback and end-timestamp collection

## Details
- **New utility**: `krkn/utils/wait_until.py` — generic polling loop supporting both `python` (importable `module.function`) and `script` (shell script with exit code) condition types
- **Integration point**: `krkn/scenario_plugins/abstract_scenario_plugin.py` — parses `wait_until` from scenario YAML and invokes the poller after rollback
- **Timeout behavior**: logs a warning and continues if condition is not met within `max_wait_time` — does not fail the scenario

### Example scenario config
```yaml
- id: kill-pods
  config:
    namespace_pattern: ^openshift-etcd$
  wait_until:
    type: python
    target: my_module.check_pods_ready
    max_wait_time: 300
    poll_interval: 5
```

## Test plan
- [x] Unit tests added in `tests/test_wait_until.py`
- [x] Tests cover: immediate success, retry logic, timeout, error handling, invalid configs, script execution
- [ ] Manual validation on a cluster with a real scenario

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -162,6 +162,29 @@ Implement: stop, start, reboot, terminate instances.
     scenario_type: <type>  # Must match plugin's get_scenario_types()
 ```
 
+### Wait Until Condition
+
+Scenarios support an optional `wait_until` block that polls a user-provided condition after chaos injection and before log collection. This allows waiting for the system to recover (e.g., pods ready, endpoints healthy) instead of using a fixed sleep.
+
+```yaml
+- id: kill-pods
+  config:
+    namespace_pattern: ^openshift-etcd$
+  wait_until:
+    type: python                          # "python" or "script"
+    target: my_module.check_pods_ready    # module.function or /path/to/script.sh
+    max_wait_time: 300                    # max seconds to wait (required)
+    poll_interval: 5                      # seconds between checks (default: 10)
+```
+
+**Condition types:**
+- **`python`**: `target` is a `module.function` path. Signature: `(kubeconfig: str, scenario: str) -> bool`. Return `True` when condition is met.
+- **`script`**: `target` is a path to a shell script. Receives kubeconfig and scenario path as arguments. Exit code `0` = condition met.
+
+**Timeout behavior**: If the condition is not met within `max_wait_time`, krkn logs a warning and continues — it does not fail the scenario.
+
+**Implementation**: `krkn/utils/wait_until.py` (polling utility), integrated in `abstract_scenario_plugin.py` after rollback and before `end_timestamp`/log collection.
+
 ## Code Style
 
 - **Import order**: Standard library, third-party, local imports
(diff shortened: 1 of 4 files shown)


kdelee:
@paigerube14 Avi's from my team also working on Ansible chaos testing

paigerube14:
Hi @Avilir, thanks for the PR! I like the idea of making recovery times more visible. However, pod_scenarios already has the timeout of waiting for pods to become ready built-in (using krkn_pod_recovery_time in the scenario yaml). In addition, Krkn also currently tracks and prints pod_rescheduling_time, pod_readiness_time, and total_recovery_time in the [telemetry output](https://krkn-chaos.dev/docs/scenarios/pod-scenario/#recovery-time-metrics-in-krkn-telemetry) and console summary.

Since this is already covered by the existing telemetry logic, we should probably avoid adding redundant timing logic to the scenario itself. Could you check if the existing output meets your needs, or if there's a specific reason we need it formatted differently here? 

I think this could definitely be a useful addition to Krkn in other scenarios, could you maybe remove this from being for pod-scenarios and add this logic to other scenarios you see fit

kdelee:
I think the desire is to measure time until not just whena pod is running again,  but a service is responsing as expected or some other more complex combination of conditions is Truthy. For example, we are running scenarios where we kill a database pod. The database heals faster than the service using it, but the recovery time of the service is what we are interested in.

paigerube14:
@kdelee @Avilir I think I understand the use case better now. Unfortunately though I'm still not sure this exact approach will work. The primary execution lifecycle for Krkn expects scenarios to run entirely within containerized environments. The local/native hooks introduced here won't function as intended in isolated container runs.

To make this feature resilient and aligned with Krkn's architecture, we should shift this responsibility to a custom health check plugin. This plugin can be decoupled from the core scenario logic and safely poll/validate the service recovery latency. I think we have two good paths forward depending on what you're trying to achieve:

1. Create a Pod Recovery Plugin/Scenario: We could implement a dedicated pod recovery check that watches all the pods within a specified namespace to track their respective downtimes, allowing you to validate how long each affected component remains unavailable. And if disrupting one pod has a large blast radius on others causing more chaos on the service as a whole. 

2. Use the HTTP Plugin: If your service exposes an external route or service endpoint, we might be able to leverage the existing HTTP plugin to hit that endpoint and monitor the uptime/response time directly.

For both cases, we would just increase the tunings.wait_duration and have the health check report how long the service was down and come back up within the duration

https://krkn-chaos.dev/docs/developers-guide/health-check-plugins/

PR #1300, opened 2026-05-12
fix: replace bare except with specific Exception handler in K8s client init

### LINKED ISSUE #1298: fix: bare except clause silently swallows errors during Kubernetes client initialization in run_kraken.py
## Summary

In `run_kraken.py` around line 231, a bare `except:` clause is used when initializing
the Kubernetes clients (`KrknKubernetes` and `KrknOpenshift`). This is a well-known
Python anti-pattern that silently catches all exceptions, including `SystemExit`,
`KeyboardInterrupt`, and `NameError` making failures invisible and extremely hard
to debug.

## The Problem

The original code:

```python
try:
    kubeconfig_path = str(kubeconfig_path)
    kubecli = KrknKubernetes(kubeconfig_path=kubeconfig_path)
    ocpcli = KrknOpenshift(kubeconfig_path=kubeconfig_path)
    except:
        kubecli.initialize_clients(None)  # NameError if kubecli was never assigned
```

Four issues with this:

1. Bare `except:` catches everything, including `KeyboardInterrupt` and `SystemExit`,
   which should never be swallowed.
2. Silent failure, if `KrknKubernetes()` raises an exception, the error is swallowed
   with no log message. You get zero feedback about what went wrong.
3. `NameError` in the except block, if the exception is raised before `kubecli` is
   assigned, then `kubecli.initialize_clients(None)` raises a `NameError`. That
   secondary exception is also silently lost.
4. `kubeconfig_path` bare expression `kubeconfig_path` on its own does nothing useful;
   it was likely intended as an existence check but has no effect.

## Expected Behaviour

- Catch only specific, expected exceptions (e.g. `Exception`)
- Log a clear error message when client initialization fails
- Exit gracefully with a non-zero code instead of proceeding with an undefined `kubecli`

## Proposed Fix

```python
try:
    kubeconfig_path = str(kubeconfig_path)
    kubecli = KrknKubernetes(kubeconfig_path=kubeconfig_path)
    ocpcli = KrknOpenshift(kubeconfig_path=kubeconfig_path)
    except Exception as e:
        logging.error(
            "Failed to initialize Kubernetes client with kubeconfig %s: %s",
            kubeconfig_path, e,
        )
        return -1
```

## Steps to Reproduce

1. Provide an invalid or inaccessible kubeconfig path in `config.yaml`
2. Run `python run_kraken.py`
3. No error is logged execution continues with an undefined `kubecli`, leading to
   a confusing `AttributeError` or `NameError` later in the run

## Impact

- Makes debugging initialization failures very difficult
- Can cause misleading secondary errors far from the actual root cause
- Violates PEP 8 and Python best practices

v0idheaven:
Fixes #1298

The bare except: block in 
un_kraken.py around line 231 was silently swallowing everything — including KeyboardInterrupt and SystemExit. Worse, if KrknKubernetes() failed before kubecli was assigned, the except block itself would throw a NameError that also disappeared silently. Bad kubeconfig, zero feedback, confusing crash later.

Replaced it with except Exception:, added logging.exception() with the kubeconfig path, and return 1 on failure. Also removed the no-op bare kubeconfig_path expression that was doing nothing.

Tested by pointing config at a nonexistent kubeconfig path — logs the error clearly and exits cleanly.

### DIFF
--- a/run_kraken.py
+++ b/run_kraken.py
@@ -218,7 +218,6 @@ def main(options, command: Optional[str]) -> int:
         safe_logger = SafeLogger(filename=telemetry_log_file)
 
         try:
-            kubeconfig_path
             os.environ["KUBECONFIG"] = str(kubeconfig_path)
             # krkn-lib-kubernetes init
             kubecli = KrknKubernetes(kubeconfig_path=kubeconfig_path)


run_kraken.py
```diff
@@ -218,13 +223,16 @@ def main(options, command: Optional[str]) -> int:
...
             kubecli = KrknKubernetes(kubeconfig_path=kubeconfig_path)
             ocpcli = KrknOpenshift(kubeconfig_path=kubeconfig_path)
-        except:
-            kubecli.initialize_clients(None)
+        except Exception:
+            logging.exception(
+                "Failed to initialize Kubernetes client with kubeconfig %s, exiting with code 1",
+                kubeconfig_path,
+            )
+            return 1
```
yogananth-subramanian: change to "return -1"
"return 1 " is mostly for  Scenario failure,Chaos scenario failed, post-scenarios still failing.

PR #1310, opened 2026-05-13
fix: use .get() for optional health check config keys to prevent KeyError

### LINKED ISSUE #1309: bug: HealthChecker crashes silently on missing optional config keys (bearer_token, auth, exit_on_failure)
In `krkn/utils/HealthChecker.py`, the `run_health_check` method accesses optional config keys using bare `[]` indexing instead of `.get()`. If a user omits `bearer_token`, `auth`, or `exit_on_failure` from their health check config, the health check thread crashes with a `KeyError`.

Because this runs in a background thread, the exception is never surfaced, the telemetry queue never gets populated, `chaos_telemetry.health_checks` ends up `None`, and the user gets no log message explaining why health checks silently stopped working.

The three affected lines:

```python
if config["bearer_token"]:           # KeyError if key absent
if config["auth"]:                   # KeyError if key absent
if config["exit_on_failure"] is True: # KeyError if key absent
```

The `config.yaml` template shows these as optional fields, so users reasonably omit them. The fix is straightforward replace with `.get()` calls with safe defaults.

v0idheaven:
Fixes #1309

`bearer_token`, `auth`, `exit_on_failure`, and `verify_url` were all accessed with bare `[]` indexing in `run_health_check`. These are optional fields in the config users reasonably omit them. When any one is missing, the health check thread crashes with a `KeyError` that disappears silently because it's running in a background thread. The telemetry queue never gets populated, `chaos_telemetry.health_checks` ends up `None`, and there's no log message to explain why.

Replaced all four with `.get()` calls with safe defaults. No behaviour change when the keys are present.

### DIFF
--- a/krkn/utils/HealthChecker.py
+++ b/krkn/utils/HealthChecker.py
@@ -46,13 +46,15 @@ class HealthChecker:
             while self.current_iterations < self.iterations:
                 for config in health_check_config.get("config"):
                     auth, headers = None, None
-                    verify_url = config["verify_url"] if "verify_url" in config else True
-                    if config["url"]: url = config["url"]
+                    verify_url = config.get("verify_url", True)
+                    if config.get("url"):
+                        url = config["url"]
 
-                    if config["bearer_token"]:
+                    if config.get("bearer_token"):
                         bearer_token = "Bearer " + config["bearer_token"]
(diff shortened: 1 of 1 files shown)


paigerube14 (approved):
/lgtm

PR #1312, opened 2026-05-14
fix(ci): narrow customapp_pod scenario to coredns to avoid etcd disrup…

### LINKED ISSUE #1314: CI: customapp_pod functional test is flaky on KinD — wildcard name_pattern can kill etcd and destabilize the cluster
## Problem

The `test_customapp_pod` CI functional test is flaky on KinD clusters. The test uses `scenarios/openshift/customapp_pod.yaml` which targets pods in `kube-system` on control-plane nodes with `name_pattern: .*`. This wildcard pattern can match **any** pod in `kube-system` — including `etcd` — and killing etcd on a single-node KinD cluster destabilizes the entire cluster, causing cascading failures in subsequent CI tests.

Additionally, the recovery time (`krkn_pod_recovery_time: 60`) is too short for control-plane pods to fully recover on KinD's resource-constrained environment, and there is no explicit `timeout` set.

## Expected Behavior

The `customapp_pod` test should reliably target a non-critical but representative pod (e.g., `coredns`) so that:
- The cluster remains stable after the chaos injection
- Subsequent CI tests are not affected
- The test still validates the pod disruption scenario with `node_label_selector`

## Observed Behavior

Intermittent CI failures where `test_customapp_pod` (or tests running after it) fail because etcd was killed, making the KinD cluster unresponsive or unstable.

## Proposed Fix

In `scenarios/openshift/customapp_pod.yaml`:
- Narrow `name_pattern` from `.*` to `coredns.*`
- Increase `krkn_pod_recovery_time` from `60` to `120`
- Add explicit `timeout: 180`

## Environment

- CI runner: `ubuntu-latest` (GitHub Actions)
- Cluster: KinD (multi-node)
- Test: `CI/tests/test_customapp_pod.sh`

ddjain:
## Summary

- Narrow `name_pattern` from `.*` to `coredns.*` in `customapp_pod.yaml` to prevent accidentally killing etcd on KinD
- Increase `krkn_pod_recovery_time` from 60s to 120s and add explicit `timeout: 180` for reliable recovery on resource-constrained CI runners

## Related Issue

Closes #1314

## Root Cause

The wildcard `name_pattern: .*` could match any pod in `kube-system` on control-plane nodes, including `etcd`. On KinD clusters (used in CI), killing etcd destabilizes the entire cluster and causes cascading failures in subsequent functional tests.

## Changes

**`scenarios/openshift/customapp_pod.yaml`:**
| Field | Before | After |
|-------|--------|-------|
| `name_pattern` | `.*` | `coredns.*` |
| `krkn_pod_recovery_time` | `60` | `120` |
| `timeout` | _(not set)_ | `180` |

## Test Plan

- [x] `test_customapp_pod` still exercises the pod disruption scenario with `node_label_selector`
- [x] CoreDNS pods exist on control-plane nodes in KinD, so the selector remains valid
- [x] No impact on non-KinD (OpenShift) usage — coredns is a safe, representative target

### DIFF
--- a/scenarios/openshift/customapp_pod.yaml
+++ b/scenarios/openshift/customapp_pod.yaml
@@ -2,8 +2,9 @@
 - id: kill-pods
   config:
     namespace_pattern: "kube-system"
-    name_pattern: .*
-    krkn_pod_recovery_time: 60
+    name_pattern: "coredns.*"
+    krkn_pod_recovery_time: 120
+    timeout: 180
     kill: 1     # num of pods to kill
     #Not needed by default, but can be used if you want to target pods on specific nodes
     # Option 1: Target pods on nodes with specific labels [master/worker nodes]


paigerube14 (approved):
/lgtm

PR #1313, opened 2026-05-14
FIX :  ensure all series are evaluated in instant vector Prometheus results

### LINKED ISSUE #1311: [BUG] slo_passed silently ignores all but the first series in instant vector Prometheus results
The `slo_passed` function in `krkn/prometheus/collector.py` incorrectly evaluates 
multi-series results for Prometheus **instant vector** queries (`"value"` key).

The current code uses an immediate `return` inside the loop over series:

```python
elif "value" in series:
    has_samples = True
    try:
        return float(series["value"][1]) == 0
    except (TypeError, ValueError):
        return False
```
Because of the direct return, only the first series is ever evaluated. If the first series passes (value == 0), the function returns True immediately and silently ignores all remaining series even if a later series is failing.

Ao-chuba:
Fixed a bug in `slo_passed()` (`krkn/prometheus/collector.py`) where instant vector  Prometheus query results containing multiple series were only partially evaluated. 
The original code used `return float(series["value"][1]) == 0` inside a `for` loop. This caused the function to exit immediately on the **first** series, silently  ignoring all subsequent series even if they were failing the SLO. so I replaced the immediate `return` with a conditional early exit on failure only:
```python
elif "value" in series:
            has_samples = True
            try:
    # if any series fails the SLO (!=0), return False immediately.
    # if it passes (==0), we must continnue checking other series.
                if float(series["value"][1]) != 0:
                    return False
            except (TypeError, ValueError):
                return False
```
The loop now continues checking all series. If no failure is found, the function correctly falls through to return True. This makes instant vector handling consistent with how range vector queries ("values") are already evaluated.

running the test  >>  pytest tests/test_prometheus_collector.py gives me this : 
===========20 passed in 4.86s ====================
all test cases passes. 
closes #1311


Ao-chuba:
as per `qodo-code-review` i have to `add doc string `and fix` comment indentation` in `slo_passed() `function .

Ao-chuba:
@paigerube14, please do check this PR out when you get time ,  there are 2 failed checks here but i dont think its related to the edited/updated code.

paigerube14:
can you rebase your PR with upstream/main and be sure to squash your commits so you have only 1, TIA

krkn/prometheus/collector.py
```diff
@@ -44,7 +60,10 @@ def slo_passed(prometheus_result: List[Any]) -> Optional[bool]:
         elif "value" in series:
             has_samples = True
             try:
-                return float(series["value"][1]) == 0
+                # If any series fails the SLO (!= 0), return False immediately.
+                # If it passes (== 0), continue checking remaining series.
+                if float(series["value"][1]) != 0:
```
paigerube14, line 65: changes look good to me here, will test after commits gets squashed

paigerube14: if you could check if there are any tests for this code and add in a check for this change that would also be super helpful

Ao-chuba: alright thanks for the review , ill get it squashed, also since currently there is no test for this current situation a new test set would have to be created.

Ao-chuba:
@paigerube14 
do review whenever time premits.  --> commits squashed to 1 , test case added for issue's situation , and rebased .

Ao-chuba:
@paigerube14  rebased again. Please do check in when time permits. thank you !

Ao-chuba:
@paigerube14 The [Functional & Unit Tests](https://github.com/krkn-chaos/krkn/actions/runs/25936357633/job/76242723749?pr=1313) failure is not related to my code changes. It looks like the CI setup step aborted early due to a timeout waiting for the container pod `(error: timed out waiting for the condition on pods/container`), which caused all the tests and coverage checks to fail.

Ao-chuba:
@paigerube14 please do check in this PR. i have updated my branch with upstream.

krkn/prometheus/collector.py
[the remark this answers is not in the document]
Ao-chuba: changed to continue

PR #1322, opened 2026-05-14
Update cron schedule for scorecard analysis

antedotee:
Changed weekly run to daily at 01:00 am ist or 19:30 utc.

### DIFF
--- a/.github/workflows/scorecard-analysis.yml
+++ b/.github/workflows/scorecard-analysis.yml
@@ -7,7 +7,7 @@ on:
 
   # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
   schedule:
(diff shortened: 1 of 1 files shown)


paigerube14:
we need to investigate the actual root of why the badge is not working

antedotee:
@paigerube14 The badge is not working because it expects the workflow to run successfully. The workflow hasn't been ran even once, it was scheduled to run on Monday. If I edit the cron to run it at 1 am ist(which is in 3 minutes), the workflow will run successfully, and then url to which badge is referencing will be valid and the badge will redirect to the correct link.

In case of any further clarifications, please let me know.

antedotee:
image

Working 😀 , thanks for the quick merge

PR #1340, opened 2026-05-18
fix: make inter-scenario sleep interruptible on STOP signal

### LINKED ISSUE #1339: bug: STOP/PAUSE signals are ignored during inter-scenario sleep, causing delayed shutdown
## Summary

When krkn is running multiple scenario files under a single scenario type, the STOP/PAUSE signal sent via the status server is silently ignored for the entire remaining batch. The signal is only checked in the outer loop of `run_kraken.py` (once per scenario *type*), but never inside `run_scenarios()` in `AbstractScenarioPlugin` which is where the actual per-file iteration and sleep happens.

## Affected Code

**`krkn/scenario_plugins/abstract_scenario_plugin.py`** end of `run_scenarios()`:
```python
logging.info(f"waiting {wait_duration} before running the next scenario")
time.sleep(wait_duration)   # ← unconditional, no signal check
```

**`run_kraken.py`** outer loop only checks signal between scenario *types*, not between individual files within a batch:
```python
while int(iteration) < iterations and run_signal != "STOP":
    for scenario in chaos_scenarios:
        if publish_running_status:
            run_signal = server.get_status(address)  # checked here...
        # ...but run_scenarios() is called and runs all files without re-checking
```

## Steps to Reproduce

1. Set `wait_duration: 60` and list 3+ scenario files under one scenario type in `config.yaml`
2. Start krkn: `python run_kraken.py`
3. Wait for the first scenario to finish (log: `waiting 60 before running the next scenario`)
4. Send STOP signal to the status server
5. Observe: krkn sleeps the full 60s and continues running remaining scenarios before honoring STOP

## Impact

- Affects **every user** running multiple scenario files under one scenario type, the most common production config (e.g., `config/config.yaml` lists multiple files under `pod_disruption_scenarios`)
- **CI/CD pipelines** that need to abort on timeout or test failure cannot do so cleanly
- **Daemon mode** users who rely on the signal server to stop the run are stuck for `N_scenarios × wait_duration` seconds after sending STOP
- With `wait_duration: 60` and 5 scenario files, a STOP signal can be delayed up to **5 minutes**
- The PAUSE signal has the same problem, it is never checked mid-batch

## Expected Behavior

After sending a STOP signal, krkn should stop at the earliest safe point at minimum, it should not start the next scenario in the batch and should not sleep through the full `wait_duration`.

## Suggested Fix

Replace the unconditional `time.sleep(wait_duration)` in `run_scenarios()` with an interruptible sleep that polls the signal at short intervals (e.g., every 1s). This requires threading a `get_signal_fn` callable through `run_scenarios()` from `run_kraken.py`, and adding a signal check before starting each scenario in the batch loop.

```python
# Instead of:
time.sleep(wait_duration)

# Use an interruptible sleep:
for _ in range(wait_duration):
    if get_signal_fn() == "STOP":
        logging.info("STOP signal received during wait, aborting remaining scenarios")
        return failed_scenarios, scenario_telemetries
    time.sleep(1)
```

This change touches `AbstractScenarioPlugin.run_scenarios()` (affecting all scenario plugins) and the call sites in `run_kraken.py`.

v0idheaven:
## Summary

Fixes #1339

`time.sleep(wait_duration)` in `AbstractScenarioPlugin.run_scenarios()` was unconditional. A STOP signal sent via the status server was only checked in the outer loop of `run_kraken.py` (once per scenario *type*), so it could be silently ignored for up to `N_scenarios × wait_duration` seconds while the batch slept between individual scenario files.

With `wait_duration: 60` and 5 scenario files, a STOP signal could be delayed up to **5 minutes**.

### `krkn/scenario_plugins/abstract_scenario_plugin.py`
- Added optional `get_signal_fn: Optional[Callable[[], str]] = None` parameter to `run_scenarios()`
- Replaced `time.sleep(wait_duration)` with a 1-second polling loop that checks `get_signal_fn()` on each tick and returns early if STOP is received
- Defaults to `None` for full backward compatibility existing callers without signal support are unaffected

### `run_kraken.py`
- Passes `lambda: server.get_status(address)` as `get_signal_fn` when `publish_running_status` is enabled
- Passes `None` otherwise, preserving existing behaviour for non-signal runs

## Before / After

**Before:**
```python
logging.info(f"waiting {wait_duration} before running the next scenario")
time.sleep(wait_duration)  # STOP signal ignored for full duration
```

**After:**
```python
logging.info(f"waiting {wait_duration} before running the next scenario")
for _ in range(int(wait_duration)):
    if get_signal_fn is not None and get_signal_fn() == "STOP":
        logging.info("STOP signal received during inter-scenario wait, aborting remaining scenarios in this batch")
        return failed_scenarios, scenario_telemetries
    time.sleep(1)
```

## Backward Compatibility

`get_signal_fn` defaults to `None`, so all existing subclasses and any external callers of `run_scenarios()` continue to work without modification.


paigerube14:
can you try to address the comments required and recomended in qodo review

https://github.com/krkn-chaos/krkn/pull/1340#issuecomment-4478770698

v0idheaven:
Addressed all qodo review comments:

- Fixed import group separation
- Added docstring to `run_scenarios`
- Added clarification for intentional sleep behavior change (interruptible via STOP signal)

Ready for re-review.

krkn/scenario_plugins/abstract_scenario_plugin.py
```diff
@@ -96,6 +98,25 @@ def run_scenarios(
                 failed_scenarios.append(scenario_config)
                 break
 
+            # Check signal before starting each scenario so a STOP/PAUSE sent
```
yogananth-subramanian: Could you also add  tests around signal interruption, that would be great

v0idheaven: Added signal interruption tests in test_signal_interruption.py covering STOP/PAUSE during inter-scenario sleep and pre-scenario signal checks.

krkn/scenario_plugins/abstract_scenario_plugin.py
```diff
@@ -96,6 +98,25 @@ def run_scenarios(
...
+                        "aborting remaining scenarios in this batch"
+                    )
+                    return failed_scenarios, scenario_telemetries
+                if signal == "PAUSE":
+                    logging.info(
+                        "PAUSE signal received, waiting for RUN signal before "
+                        "starting next scenario"
+                    )
+                    while get_signal_fn() == "PAUSE":
+                        time.sleep(1)
```
paigerube14: think for Pause we can let it continue to sleep here and the pause functionality at the run_kraken layer will take over

krkn/scenario_plugins/abstract_scenario_plugin.py
```diff
@@ -96,6 +98,25 @@ def run_scenarios(
...
+            # Check signal before starting each scenario so a STOP/PAUSE sent
+            # while the previous scenario was running is honoured immediately.
+            if get_signal_fn is not None:
+                signal = get_signal_fn()
+                if signal == "STOP":
+                    logging.info(
+                        "STOP signal received before starting next scenario, "
+                        "aborting remaining scenarios in this batch"
+                    )
+                    return failed_scenarios, scenario_telemetries
```
paigerube14, line 110: `run_scenarios()` returns early on STOP, but `run_kraken.py` doesn't know *why* it returned — it can't distinguish "completed normally" from "aborted due to STOP." If there are multiple scenario batches or types in the outer loop, the loop will continue into the next batch regardless. `run_kraken.py` needs to check the signal (or inspect the return somehow) after `run_scenarios()` returns and break the outer loop if STOP was received.

paigerube14: I think we'll still want to continue out of the loop that is in here because we might not have the full details of scenario_telemetries

v0idheaven: Fixed in run_kraken.py replaced break with a signal re-check after run_scenarios() returns. run_signal is already set to "STOP" by server.get_status(), so the outer while loop exits cleanly on the next tick while preserving full scenario_telemetries for reporting.

tests/test_signal_interruption.py
```diff
@@ -0,0 +1,207 @@
...
+    def _setup_signal_ctx(self, mock_signal_ctx):
+        mock_signal_ctx.return_value.__enter__ = Mock()
+        mock_signal_ctx.return_value.__exit__ = Mock(return_value=False)
+
+    # ------------------------------------------------------------------
+    # STOP during inter-scenario sleep
+    # ------------------------------------------------------------------
+
+    @_apply_patches
+    def test_stop_during_sleep_aborts_remaining_scenarios(
```
paigerube14, line 76: this test is failing, please take a look

v0idheaven: Fixed in the latest commit, patched time.monotonic and corrected the signal sequences so STOP arrives inside the sleep loop rather than being consumed by the pre-scenario check.

krkn/scenario_plugins/abstract_scenario_plugin.py
```diff
@@ -172,7 +185,18 @@ def run_scenarios(
...
+            # PAUSE is handled by the outer loop in run_kraken.py.
+            end = time.monotonic() + wait_duration
+            while time.monotonic() < end:
+                if get_signal_fn is not None and get_signal_fn() == "STOP":
+                    logging.info(
+                        "STOP signal received during inter-scenario wait, "
+                        "aborting remaining scenarios in this batch"
+                    )
+                    return failed_scenarios, scenario_telemetries
+                time.sleep(1)
```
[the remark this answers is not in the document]
v0idheaven: Fixed in commit ee10baf: replaced time.sleep(1) with time.sleep(min(1, max(0, end - time.monotonic()))) so the loop sleeps only the remaining slice of the deadline. Sub-second wait_duration values now work correctly and the loop can never overshoot by a full second.

krkn/scenario_plugins/abstract_scenario_plugin.py
```diff
@@ -172,7 +185,18 @@ def run_scenarios(
             scenario_telemetries.append(scenario_telemetry)
             cerberus.publish_kraken_status(start_time,end_time)
             logging.info(f"waiting {wait_duration} before running the next scenario")
-            time.sleep(wait_duration)
+            # Interruptible sleep: use a monotonic deadline to preserve
+            # sub-second wait_duration semantics (int() would floor 0.5 → 0).
+            # PAUSE is handled by the outer loop in run_kraken.py.
+            end = time.monotonic() + wait_duration
+            while time.monotonic() < end:
+                if get_signal_fn is not None and get_signal_fn() == "STOP":
```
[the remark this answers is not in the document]
v0idheaven: Fixed in commit ee10baf: downgraded both logging.info() calls in server.get_status() (lines 91 and 95) to logging.debug(). The 'connection set up' and 'response ...' messages are now silent at the default log level, eliminating the hundreds of log entries produced across scenarios with large wait_duration values.

PR #1355, opened 2026-05-23
added stale workflow from actions repo

### LINKED ISSUE #1343: Add stale.yml from the actions repository here to use
There exists stale.yml, but since we are centralizing everything in the actions repository, it would be nice to have to replace it by referencing from there.

antedotee:
Replaced old stale workflow by using actions repo.

Fixes #1343

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -2,9 +2,8 @@ name: Manage Stale Issues and Pull Requests
 
 on:
   schedule:
-    # Run daily at 1:00 AM UTC
-    - cron: '0 1 * * *'
-  workflow_dispatch: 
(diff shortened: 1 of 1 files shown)


antedotee:
CC @paigerube14 . Please have a look.

.github/workflows/stale.yml
```diff
@@ -2,51 +2,13 @@ name: Manage Stale Issues and Pull Requests
 
 on:
   schedule:
-    # Run daily at 1:00 AM UTC
```
paigerube14, line 5: lets leave this comment

paigerube14 (commented):
let's leave the one comment, but otherwise LGTM

PR #1357, opened 2026-05-23
fix: replace bare except with proper exception handling in client init

### LINKED ISSUE #1356: run_kraken.py: bare except with non-existent method call masks real initialization error
## Bug Description

In `run_kraken.py` lines 220-227, the `except` block calls `kubecli.initialize_clients(None)` as a fallback when Kubernetes client initialization fails. This method does not exist on `KrknKubernetes` in krkn-lib 6.1.0 (verified against the published wheel). So any failure in the `try` block causes the `except` block to throw a second exception, completely hiding the original error.

There are two failure modes depending on where the `try` block fails:

1. If `KrknKubernetes(kubeconfig_path=kubeconfig_path)` raises before completing, `kubecli` is never assigned. The `except` block then raises `UnboundLocalError: local variable 'kubecli' referenced before assignment`.

2. If `kubecli` was assigned but the call still fails later, the `except` block raises `AttributeError: 'KrknKubernetes' object has no attribute 'initialize_clients'`.

In both cases the original exception is lost and the user has no idea what actually failed.

The bare `except:` also catches `KeyboardInterrupt` and `SystemExit`, which is unintended.

**Affected code (`run_kraken.py` lines 220-227):**

```python
try:
    kubeconfig_path
    os.environ["KUBECONFIG"] = str(kubeconfig_path)
    kubecli = KrknKubernetes(kubeconfig_path=kubeconfig_path)
    ocpcli = KrknOpenshift(kubeconfig_path=kubeconfig_path)
except:
    kubecli.initialize_clients(None)
```

### Impact
When krkn fails to connect to the cluster, the real error (invalid kubeconfig, unreachable cluster, missing credentials) is never surfaced. The user sees a misleading secondary exception with no useful diagnostic information. This makes failures in CI and production environments significantly harder to debug.

### Expected Fix
```
try:
    os.environ["KUBECONFIG"] = str(kubeconfig_path)
    kubecli = KrknKubernetes(kubeconfig_path=kubeconfig_path)
    ocpcli = KrknOpenshift(kubeconfig_path=kubeconfig_path)
except Exception as e:
    logging.exception("Failed to initialize Kubernetes/OpenShift clients")
    raise
```

The initialize_clients fallback call is not needed. KrknKubernetes.__initialize_config already handles the in-cluster service account token fallback internally when no kubeconfig file is present.

### Environment
run_kraken.py, lines 220-227
Library: krkn-lib==6.1.0

netram75:
The except block called `kubecli.initialize_clients(None)` which does not exist on KrknKubernetes in krkn-lib 6.1.0. If the try block raised before kubecli was assigned, this caused an UnboundLocalError masking the original exception. If kubecli was assigned, it caused AttributeError for the same reason. Either way the real error was lost.

Also removed the bare `except:` which was catching KeyboardInterrupt and SystemExit unintentionally, and removed the no-op `kubeconfig_path` statement.

Closes #1356

# Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization

# Description

The except block in the Kubernetes client initialization block called `kubecli.initialize_clients(None)`, a method that does not exist on KrknKubernetes in krkn-lib 6.1.0. This caused a secondary exception inside the except block, completely masking the original error. Replaced with proper exception logging and re-raise so the real failure is always surfaced.

- Related Issue #: 1356
- Closes #: 1356

# Documentation

Not needed. This is a bug fix with no user-facing behavior change.

*REQUIRED*:

This fix removes a broken except block that called a non-existent method. A live Kubernetes or OpenShift cluster is required to run krkn end-to-end. The fix is verifiable by code inspection: `initialize_clients` does not exist in any `.py` file in the krkn-lib 6.1.0 published wheel, verified locally by extracting the wheel archive.

### DIFF
--- a/run_kraken.py
+++ b/run_kraken.py
@@ -218,15 +218,12 @@ def main(options, command: Optional[str]) -> int:
         safe_logger = SafeLogger(filename=telemetry_log_file)
 
         try:
-            kubeconfig_path
             os.environ["KUBECONFIG"] = str(kubeconfig_path)
-            # krkn-lib-kubernetes init
             kubecli = KrknKubernetes(kubeconfig_path=kubeconfig_path)
             ocpcli = KrknOpenshift(kubeconfig_path=kubeconfig_path)
-        except Exception as e:
-            logging.error("Failed to initialize Kubernetes clients: %s" % e)
-            kubecli = KrknKubernetes(kubeconfig_path=None)
-            ocpcli = KrknOpenshift(kubeconfig_path=None)
+        except Exception:
+            logging.exception("Failed to initialize Kubernetes/OpenShift clients")
+            return -1
 
         distribution = "kubernetes"
         if ocpcli.is_openshift():


chaitanyaenr:
cosmos review

netram75:
Hi @chaitanyaenr @paigerube14 @tsebastiani, just checking in - happy to make any changes if needed. Please let me know if you have any feedback.

Issue #1359, opened 2026-05-28
[Bug] check_license.py crashes on Windows during directory traversal


Aastha-spec-tech:
While running `scripts/check_license.py` on Windows, the script crashes with a path-related error instead of completing the license check.

The issue seems to come from using `Path.rglob("*.py")` for recursive traversal. On Windows, this can run into problems when it encounters unresolved symlinks or restricted directories, causing the entire script to fail with an `os error 2`.

Example error:

```bash
d:\krkn\krkn\.antigravityignore: The system cannot find the file specified. (os error 2)
```

### Steps to reproduce

1. Clone the repository on a Windows machine.
2. Open PowerShell or Command Prompt in the repository root.
3. Run:

   ```bash
   python scripts/check_license.py
   ```

<img width="1320" height="590" alt="Image" src="https://github.com/user-attachments/assets/dc5cc994-38eb-4d46-8dab-8846fe473a7a" />

### Proposed solution

A possible fix would be to replace the current `Path.rglob()` traversal inside `collect_source_files()` with `os.walk(..., followlinks=False)`.

PR #1360, opened 2026-05-28
fix(license): resolve check_license.py symlink traversal crash on Windows

Aastha-spec-tech:
#1359
hi @paigerube14 
Type of Change:
Bug fix

Problem: In Windows environments, scripts/check_license.py used Path.rglob("*.py") to search for python source files. On Windows systems, this method attempts to traverse directories in a way that triggers path-not-found or permission issues (such as with symlinks), causing the license checker script to crash with os error 2 during local runs.

Solution: Replaced Path.rglob with os.walk(..., followlinks=False) to safely traverse the repository tree and ignore excluded directories in-place.

Impact: Prevents the license checker from crashing on Windows, ensuring a consistent local development and CI experience across all operating systems.

### DIFF
--- a/scripts/check_license.py
+++ b/scripts/check_license.py
@@ -32,6 +32,7 @@ Usage:
 import argparse
 import re
 import sys
+import os
 from datetime import datetime
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


paigerube14 (approved):
/lgtm 

tried on my mac and it also works

PR #1368, opened 2026-05-30
fix(node-actions): stop_start_kubelet_scenario calls restart_kubelet_scenario instead of node_reboot_scenario

### LINKED ISSUE #1367: fix(node-actions): stop_start_kubelet_scenario calls node_reboot_scenario instead of restart_kubelet_scenario
## What's wrong

`stop_start_kubelet_scenario` in `krkn/scenario_plugins/node_actions/abstract_node_scenarios.py` calls `node_reboot_scenario` as its second step instead of `restart_kubelet_scenario`.

```python
# Line 103-108
def stop_start_kubelet_scenario(self, instance_kill_count, node, timeout):
    logging.info("Starting stop_start_kubelet_scenario injection")
    self.stop_kubelet_scenario(instance_kill_count, node, timeout)
    self.node_reboot_scenario(instance_kill_count, node, timeout)  # wrong call
    self.affected_nodes_status.merge_affected_nodes()
    logging.info("stop_start_kubelet_scenario has been successfully injected!")
```

The intended behavior is: stop the kubelet, then start it back up. What actually happens is: stop the kubelet, then reboot the entire node.

### Why it matters
A kubelet restart and a full node reboot are very different operations. A kubelet restart is a targeted process-level recovery test. A node reboot evicts all pods on that node, unmounts storage volumes, and takes the node offline for several minutes.

Users running this scenario believe they are testing whether their cluster handles kubelet recovery gracefully. They are actually triggering a full node reboot without knowing it. The scenario logs "successfully injected" either way, so there is no indication anything went wrong.

### Fix
restart_kubelet_scenario already exists in the same file at line 111. The fix is a one-line change:
```
def stop_start_kubelet_scenario(self, instance_kill_count, node, timeout):
    logging.info("Starting stop_start_kubelet_scenario injection")
    self.stop_kubelet_scenario(instance_kill_count, node, timeout)
    self.restart_kubelet_scenario(instance_kill_count, node, timeout)  # correct call
    self.affected_nodes_status.merge_affected_nodes()
    logging.info("stop_start_kubelet_scenario has been successfully injected!")
```

netram75:
`stop_start_kubelet_scenario` was calling `node_reboot_scenario` as its second step instead of `restart_kubelet_scenario`, causing a full node reboot instead of a targeted kubelet restart.

The node reboot would evict all pods on the node and take it offline for several minutes, which is not what users expect when running this scenario. Since the node comes back up after a reboot with the kubelet running, the scenario would log success either way and the wrong behavior went unnoticed.

`restart_kubelet_scenario` already exists in the same file. Replaced the incorrect call with it.

Fixes #1367

# Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization

# Description

`stop_start_kubelet_scenario` in `krkn/scenario_plugins/node_actions/abstract_node_scenarios.py` was calling `node_reboot_scenario` instead of `restart_kubelet_scenario`. Replaced with the correct call.

- Related Issue #: 1367
- Closes #: 1367

# Documentation

- [ ] I have performed a self-review of my code by running krkn and specific scenario

### DIFF
--- a/krkn/scenario_plugins/node_actions/abstract_node_scenarios.py
+++ b/krkn/scenario_plugins/node_actions/abstract_node_scenarios.py
@@ -103,6 +103,9 @@ class abstract_node_scenarios:
     def stop_start_kubelet_scenario(self, instance_kill_count, node, timeout):
         logging.info("Starting stop_start_kubelet_scenario injection")
         self.stop_kubelet_scenario(instance_kill_count, node, timeout)
+        # node_reboot_scenario is used intentionally: stopping the kubelet leaves the node NotReady,
+        # and oc debug cannot connect to a NotReady node, so restart_kubelet_scenario would fail.
+        # A hard reboot restarts the node and brings it back to Ready state.
         self.node_reboot_scenario(instance_kill_count, node, timeout)
         self.affected_nodes_status.merge_affected_nodes()
         logging.info("stop_start_kubelet_scenario has been successfully injected!")


krkn/scenario_plugins/node_actions/abstract_node_scenarios.py
```diff
@@ -103,7 +103,7 @@ def stop_kubelet_scenario(self, instance_kill_count, node, timeout):
     def stop_start_kubelet_scenario(self, instance_kill_count, node, timeout):
         logging.info("Starting stop_start_kubelet_scenario injection")
         self.stop_kubelet_scenario(instance_kill_count, node, timeout)
-        self.node_reboot_scenario(instance_kill_count, node, timeout)
+        self.restart_kubelet_scenario(instance_kill_count, node, timeout)
```
paigerube14: I think we might want to keep the node_reboot functionality here, when we stop the kubelet the node goes NotReady and needs a hard reboot to be able to restart and go ready. The restart_kubelet function does and `oc debug` and won't be able to form a connection to the node

netram75: Thanks for the context @paigerube14. That makes sense on the oc debug limitation. Should I close this PR then, or would it be worth adding a comment in the code explaining why node_reboot_scenario is used here instead of restart_kubelet_scenario?

paigerube14: a comment would be helpful, yes if you could add that would be great

netram75: @paigerube14  Added the comment explaining why node_reboot_scenario is used here. Let me know if the wording looks good.

paigerube14 (approved):

PR #1376, opened 2026-06-02
fix(hogs): guard against ZeroDivisionError when samples list is empty

### LINKED ISSUE #1375: fix(hogs): ZeroDivisionError crash in run_scenario_worker when duration is too short to collect samples
In `run_scenario_worker()` (`krkn/scenario_plugins/hogs/hogs_scenario_plugin.py`, lines 137-139), the function divides by `len(samples)` to calculate average resource usage after the hog pod finishes.

`samples` will be empty when `config.duration` is too short. The sampling loop on line 115 is:

```python
while time.time() - start < config.duration - 1:
    samples.append(lib_k8s.get_node_resources_info(node))
```

However, line 109 has a hardcoded `time.sleep(3)` before this loop. By the time the loop is reached, 3 seconds have already elapsed. If `duration` is 4 or less, the condition is already False before the first iteration and `samples` stays empty.

Lines 137-139 then crash with `ZeroDivisionError`:

```python
avg_node_resources.cpu = avg_node_resources.cpu / len(samples)
avg_node_resources.memory = avg_node_resources.memory / len(samples)
avg_node_resources.disk_space = avg_node_resources.disk_space / len(samples)
```

This affects all three hog types: cpu, memory, and io.

## Steps to reproduce
Set `duration: 4` or less in any hogs scenario config and run krkn. The scenario crashes with `ZeroDivisionError`.

## Expected behavior
When no samples are collected, the function logs a warning and skips the average calculation instead of crashing.

## Fix
```python
if len(samples) == 0:
    logging.warning(f"[{node}] no samples collected, duration too short to measure resource usage")
else:
    avg_node_resources.cpu = avg_node_resources.cpu / len(samples)
    avg_node_resources.memory = avg_node_resources.memory / len(samples)
    avg_node_resources.disk_space = avg_node_resources.disk_space / len(samples)
```

netram75:
# Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization

# Description

`run_scenario_worker()` in `krkn/scenario_plugins/hogs/hogs_scenario_plugin.py` crashes with `ZeroDivisionError` when `config.duration` is too short to collect any samples.

Line 109 has a hardcoded `time.sleep(3)` before the sampling loop. The loop condition on line 115 is:

```python
while time.time() - start < config.duration - 1:
```

If `duration` is 4 or less, 3 seconds have already elapsed by the time the loop starts and `samples` stays empty. Lines 137-139 then crash dividing by `len(samples)` which is 0.

Added a guard that logs a warning and skips the average calculation when no samples were collected. This affects all three hog types: cpu, memory, and io.

- Related Issue #1375
- Closes #1375

# Documentation

- [ ] I have performed a self-review of my code by running krkn and specific scenario

## Test results

This fix adds a guard before the division on lines 137-139. The crash is triggered by setting `duration: 4` or less in any hogs scenario config. The fix logs a warning and skips the average calculation when no samples were collected instead of crashing.

### DIFF
--- a/krkn/scenario_plugins/hogs/hogs_scenario_plugin.py
+++ b/krkn/scenario_plugins/hogs/hogs_scenario_plugin.py
@@ -134,9 +134,13 @@ class HogsScenarioPlugin(AbstractScenarioPlugin):
                 avg_node_resources.memory += resource.memory
                 avg_node_resources.disk_space += resource.disk_space
 
-            avg_node_resources.cpu = avg_node_resources.cpu/len(samples)
-            avg_node_resources.memory = avg_node_resources.memory / len(samples)
-            avg_node_resources.disk_space = avg_node_resources.disk_space / len(samples)
+            if len(samples) == 0:
+                logging.warning(f"[{node}] no samples collected, duration too short to measure resource usage")
+                return
+            else:
+                avg_node_resources.cpu = avg_node_resources.cpu / len(samples)
+                avg_node_resources.memory = avg_node_resources.memory / len(samples)
+                avg_node_resources.disk_space = avg_node_resources.disk_space / len(samples)
 
             if config.type == HogType.cpu:
                 logging.info(f"[{node}] detected cpu consumption: "
--- a/tests/test_hogs_scenario_plugin.py
+++ b/tests/test_hogs_scenario_plugin.py
@@ -6,14 +6,14 @@ Test suite for HogsScenarioPlugin class
 Usage:
     python -m coverage run -a -m unittest tests/test_hogs_scenario_plugin.py -v
 
-Assisted By: Claude Code
 """
 
+import queue
 import unittest
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
-from krkn_lib.k8s import KrknKubernetes
-from krkn_lib.telemetry.ocp import KrknTelemetryOpenshift
+from krkn_lib.models.krkn import HogConfig, HogType
+from krkn_lib.models.k8s import NodeResources
 
 from krkn.scenario_plugins.hogs.hogs_scenario_plugin import HogsScenarioPlugin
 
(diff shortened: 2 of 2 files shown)


krkn/scenario_plugins/hogs/hogs_scenario_plugin.py
```diff
@@ -134,9 +134,12 @@ def run_scenario_worker(self, config: HogConfig,
                 avg_node_resources.memory += resource.memory
                 avg_node_resources.disk_space += resource.disk_space
 
-            avg_node_resources.cpu = avg_node_resources.cpu/len(samples)
-            avg_node_resources.memory = avg_node_resources.memory / len(samples)
-            avg_node_resources.disk_space = avg_node_resources.disk_space / len(samples)
+            if len(samples) == 0:
+                logging.warning(f"[{node}] no samples collected, duration too short to measure resource usage")
```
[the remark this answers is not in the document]
netram75: Good catch. Added return after the warning so execution doesn't fall through to the logging blocks with zero values. Fixed in the latest commit.

paigerube14 (approved):

krkn/scenario_plugins/hogs/hogs_scenario_plugin.py
```diff
@@ -134,9 +134,13 @@ def run_scenario_worker(self, config: HogConfig,
                 avg_node_resources.memory += resource.memory
                 avg_node_resources.disk_space += resource.disk_space
 
-            avg_node_resources.cpu = avg_node_resources.cpu/len(samples)
-            avg_node_resources.memory = avg_node_resources.memory / len(samples)
-            avg_node_resources.disk_space = avg_node_resources.disk_space / len(samples)
+            if len(samples) == 0:
```
paigerube14, line 137: could be nice to add a unit test around this

netram75: @paigerube14 Thanks for the suggestion! Added a unit test in tests/test_hogs_scenario_plugin.py that covers the empty samples case.

paigerube14:
@netram75 something seems off with the test you added, its just hanging when running the unit tests. Please try running locally and be sure it runs before commiting

netram75:
@paigerube14  Sorry about that! Fixed the time.time mock to use side_effect=[0, 1000, 1000, 1000] so the sampling loop exits immediately. Ran locally and both tests pass:


test_get_scenario_types ... ok
test_run_scenario_worker_no_samples_no_zero_division ... ok
Ran 2 tests in 0.039s OK

netram75:
@paigerube14 while going through the docs I noticed memory-hog mentions "Validate that memory-based horizontal pod autoscaling works correctly" as a use case, but krkn has no chaos scenario that actually targets HPAs.

I am thinking of adding a new autoscaler scenario plugin that patches an HPA's minReplicas and maxReplicas to simulate autoscaler misconfiguration or failure, holds that state for a configurable duration, then restores the original values. This would let users test whether their workloads degrade gracefully when autoscaling is broken.

Let me know if this is worth pursuing.

paigerube14:
@netram75 your functional test you added I think is failing, are you able to take a look, TIA

```
======================================================================
ERROR: test_run_scenario_worker_no_samples_no_zero_division (test_hogs_scenario_plugin.TestHogsScenarioPlugin.test_run_scenario_worker_no_samples_no_zero_division)
  File "/home/runner/work/krkn/krkn/tests/test_hogs_scenario_plugin.py", line 49, in test_run_scenario_worker_no_samples_no_zero_division
    raise exception_queue.get_nowait()
  File "/home/runner/work/krkn/krkn/krkn/scenario_plugins/hogs/hogs_scenario_plugin.py", line 115, in run_scenario_worker
    while time.time() - start < config.duration-1:
          ^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.15/x64/lib/python3.11/unittest/mock.py", line 1124, in __call__
    return self._mock_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.15/x64/lib/python3.11/unittest/mock.py", line 1128, in _mock_call
    return self._execute_mock_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.15/x64/lib/python3.11/unittest/mock.py", line 1185, in _execute_mock_call
    result = next(effect)
             ^^^^^^^^^^^^
StopIteration

```

PR #1387, opened 2026-06-09
hogs: store resource utilization and recovery time in scenario telemetry

### LINKED ISSUE #1385: hogs: store resource utilization and recovery time in scenario telemetry
## Summary

The hog scenario samples node resource utilization throughout the chaos duration and computes averages — but never written to `scenario_telemetry`. Adding these as telemetry metrics would give meaningful injection impact data and a genuine resource-level recovery metric (time for the node to return to pre-chaos baseline), which no other scenario currently measures.

## What Already Exists

In `run_scenario_worker()` ([hogs_scenario_plugin.py#L84](https://github.com/krkn-chaos/krkn/blob/main/krkn/scenario_plugins/hogs/hogs_scenario_plugin.py#L84)):

- `node_resources_start` — baseline captured before hog via `get_node_resources_info()` (line 98)
- `samples` loop — polls `get_node_resources_info()` throughout the hog duration (line 116)
- `avg_node_resources` — avg CPU/memory/disk computed from samples (lines 132–139)
- All metrics are computed and **only logged**, never stored in `scenario_telemetry`

## Reference Implementation (pod_disruption)

- Config param: [`krkn_pod_recovery_time`](https://github.com/krkn-chaos/krkn/blob/main/krkn/scenario_plugins/pod_disruption/models/models.py#L23)
- Recovery timing + telemetry write: [`pod_disruption_scenario_plugin.py#L100–L132`](https://github.com/krkn-chaos/krkn/blob/main/krkn/scenario_plugins/pod_disruption/pod_disruption_scenario_plugin.py#L100-L132)

### 1. Injection impact metrics (what was consumed during chaos)
Already computed, just needs storing:
- `avg_cpu_utilization` — average CPU consumed as a percentage during hog
- `avg_memory_increase` — average memory increase relative to baseline
- `avg_disk_allocated` — average disk space allocated during hog

### 2. Resource recovery time (how long after hog stops until node returns to baseline)
After pod deletion, poll `get_node_resources_info()` comparing against `node_resources_start` until the delta is within a threshold — this is the recovery time. No other scenario currently measures resource-level recovery.

### krkn-lib — `krkn_lib/models/k8s/models.py`
- Add `HogNodeMetrics` model with:
  - `node_name: str`
  - `avg_cpu: float`
  - `avg_memory: float`
  - `avg_disk: float`
  - `resource_recovery_time: float` — seconds until node resources return to pre-chaos baseline
- `NodeResources` already exists with `cpu`, `memory`, `disk_space` — no changes needed there

### krkn-lib — `krkn_lib/models/telemetry/models.py`
- Add `affected_nodes_resources: list[HogNodeMetrics]` to `ScenarioTelemetry`

### krkn-lib — `krkn_lib/models/elastic/models.py`
- Add `ElasticHogNodeMetrics(InnerDoc)` with `node_name`, `avg_cpu = Float()`, `avg_memory = Float()`, `avg_disk = Float()`, `resource_recovery_time = Float()` — mirrors `ElasticAffectedNodes` pattern
- Add `affected_nodes_resources = Nested(ElasticHogNodeMetrics, multi=True)` to `ElasticScenarioTelemetry`
- Wire mapping in `ElasticChaosRunTelemetry` from `sc.affected_nodes_resources`

### krkn — `hogs_scenario_plugin.py`
- After pod deletion, add a polling loop comparing `get_node_resources_info()` against `node_resources_start` until within threshold — record elapsed time as `resource_recovery_time`
- Build a `HogNodeMetrics` object per node with the computed averages and recovery time
- Write to `scenario_telemetry.affected_nodes_resources`

v0idheaven:
## Summary

Closes #1385

The hog scenario samples node resource utilization throughout the chaos duration and computes averages — but the computed values were only logged and never written to `scenario_telemetry`. This PR stores them plus adds a genuine resource-level recovery metric.

### `scenario_plugins/hogs/hogs_scenario_plugin.py`

**Thread `scenario_telemetry` + lock through the call stack**

`run()` → `run_scenario()` → `run_scenario_worker()` now pass a `ScenarioTelemetry` instance and a `threading.Lock` so each per-node worker can write results safely without races.

**Injection impact metrics (metric 1)**

After computing `avg_node_resources` from samples collected during the hog duration, the following are written into `scenario_telemetry.parameters["node_resource_metrics"][<node_name>]`:

| Key | Description |
|---|---|
| `avg_cpu_nanocores` | Average CPU usage (nanocores) during chaos |
| `avg_memory_available_bytes` | Average available memory during chaos |
| `avg_disk_available_bytes` | Average available disk during chaos |
| `baseline_cpu_nanocores` | Pre-chaos CPU baseline |
| `baseline_memory_available_bytes` | Pre-chaos memory baseline |
| `baseline_disk_available_bytes` | Pre-chaos disk baseline |

**Resource-level recovery time (metric 2)**

After the hog pod is deleted, a polling loop calls `get_node_resources_info()` every 5 s until the metric relevant to the hog type (cpu / memory / disk) returns within **5 % of the pre-chaos baseline**, up to `krkn_node_recovery_timeout` seconds.

- Default timeout: **120 s** (`_DEFAULT_RECOVERY_TIMEOUT`)
- Configurable per scenario YAML via `node-recovery-timeout:` key — mirrors `krkn_pod_recovery_time` from the pod_disruption scenario
- Result stored as `resource_recovery_time_seconds` in the per-node dict; value is `-1` if the node did not recover within the timeout

**`_is_recovered()` helper**

Extracted as a `@staticmethod` so the type-specific tolerance check is self-contained and independently testable.

## Reference Implementation

Follows the same pattern as `pod_disruption_scenario_plugin.py` which writes `scenario_telemetry.affected_pods` and uses `krkn_pod_recovery_time` as the recovery timeout cap.

## Testing

- `python -m py_compile` passes on the modified file.
- No pre-existing unit tests for this plugin; the logic is integration-testable by running a hog scenario and inspecting the emitted telemetry JSON.

### DIFF
--- a/krkn/scenario_plugins/hogs/hogs_scenario_plugin.py
+++ b/krkn/scenario_plugins/hogs/hogs_scenario_plugin.py
@@ -23,7 +23,7 @@ import time
 import yaml
 from krkn_lib.models.telemetry import ScenarioTelemetry
 from krkn_lib.telemetry.ocp import KrknTelemetryOpenshift
-from krkn_lib.models.krkn import  HogConfig, HogType
+from krkn_lib.models.krkn import HogConfig, HogType
 from krkn_lib.models.k8s import NodeResources
 from krkn_lib.k8s import KrknKubernetes
 from krkn_lib.utils import get_random_string
@@ -32,9 +32,16 @@ from krkn.scenario_plugins.abstract_scenario_plugin import AbstractScenarioPlugi
 from krkn.rollback.config import RollbackContent
 from krkn.rollback.handler import set_rollback_context_decorator
 
+# How close a metric must be to the pre-chaos baseline to be considered
+# "recovered" (5 % tolerance).
+_RECOVERY_TOLERANCE = 0.05
+
+# Default maximum seconds to wait for node resources to return to baseline.
+_DEFAULT_RECOVERY_TIMEOUT = 120
+
 
 class HogsScenarioPlugin(AbstractScenarioPlugin):
-    
+
     @set_rollback_context_decorator
     def run(self, run_uuid: str, scenario: str, lib_telemetry: KrknTelemetryOpenshift,
             scenario_telemetry: ScenarioTelemetry) -> int:
@@ -42,10 +49,17 @@ class HogsScenarioPlugin(AbstractScenarioPlugin):
             with open(scenario, "r") as f:
                 scenario = yaml.safe_load(f)
             scenario_config = HogConfig.from_yaml_dict(scenario)
-            
+
+            # Optional: maximum seconds to wait for node resources to recover
+            # after the hog completes (mirrors krkn_pod_recovery_time in
+            # pod_disruption scenario).  Defaults to _DEFAULT_RECOVERY_TIMEOUT.
+            scenario_config.krkn_node_recovery_timeout = int(
+                scenario.get("node-recovery-timeout", _DEFAULT_RECOVERY_TIMEOUT)
+            )
+
             # Get node-name if provided
             node_name = scenario.get('node-name')
-            
+
             has_selector = True
             if not scenario_config.node_selector or not re.match("^.+=.*$", scenario_config.node_selector):
                 if scenario_config.node_selector:
@@ -71,8 +85,21 @@ class HogsScenarioPlugin(AbstractScenarioPlugin):
             if scenario_config.number_of_nodes and len(available_nodes) > scenario_config.number_of_nodes:
                 available_nodes = random.sample(available_nodes, scenario_config.number_of_nodes)
 
+            # Ensure scenario_telemetry.parameters is a dict so per-node
+            # metrics can be written into it thread-safely.
+            if not isinstance(scenario_telemetry.parameters, dict):
+                scenario_telemetry.parameters = {}
+
+            telemetry_lock = threading.Lock()
             exception_queue = queue.Queue()
-            self.run_scenario(scenario_config, lib_telemetry.get_lib_kubernetes(), available_nodes, exception_queue)
+            self.run_scenario(
+                scenario_config,
+                lib_telemetry.get_lib_kubernetes(),
+                available_nodes,
+                exception_queue,
+                scenario_telemetry,
+                telemetry_lock,
+            )
             return 0
         except Exception as e:
             logging.error(f"scenario exception: {e}")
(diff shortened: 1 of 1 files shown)


SK8-infi (commented):
Hi Varun recently made changes to the same file.

I tested this PR live on an **Azure AKS cluster**  using a CPU hog scenario. The injection was successful (CPU utilization spiked from ~450M nanocores to ~3.6B nanocores), and all telemetry fields (`avg_*`, `baseline_*`, `resource_recovery_time_seconds`) were correctly populated under `scenario_telemetry.parameters["node_resource_metrics"]`.

However, there are a few logic and implementation items I would like to point out:

### 1. Logical Flaw in `_is_recovered()` for Memory, Disk, and CPU
The helper currently checks `abs(current - baseline)` against a 5% tolerance for all resource types. This leads to false-positive recovery failures in realistic cluster environments:
- **Memory & Disk Space (`available_bytes`)**: Memory and disk space metrics represent *available* capacity (higher is better). If OS background processes or container teardown free up memory/disk after chaos such that post-chaos free space is *greater* than baseline (e.g., 11 GB free vs 10 GB baseline), `abs(11 - 10) / 10 = 10% > 5% tolerance`. `_is_recovered()` returns `False`, causing the polling loop to block until the 120s timeout and log a false failure (`resource_recovery_time_seconds: -1`).
- **CPU Usage (`nanocores`)**: CPU usage represents load (lower is better). If post-chaos CPU load drops below the pre-chaos baseline, `abs(current.cpu - baseline.cpu)` can also exceed the tolerance threshold and return `False`.

I would suggest you use one-sided threshold checks.

### 2. Dynamic Attribute Assignment on `HogConfig`
In `run()`, line 56 dynamically assigns `scenario_config.krkn_node_recovery_timeout = ...` on the `HogConfig` dataclass, which is later fetched via `getattr(config, "krkn_node_recovery_timeout", ...)`. 
Injecting un-declared attributes onto dataclass instances is fragile in Python. Consider either adding `krkn_node_recovery_timeout` officially to `HogConfig` in `krkn_lib`, or passing `node_recovery_timeout` directly as a parameter through `run_scenario()` and `run_scenario_worker()`.

This ones optional...if works out in your scope

### 3. Unit Tests Request
Since `_is_recovered()` is extracted as a `@staticmethod`, adding unit tests in `tests/` covering CPU, memory, and disk recovery checks (including cases where post-chaos resources exceed baseline) would prevent future regressions.

PR #1427, opened 2026-06-27
protect config.telemetry access with fallback to empty dict

Sahil-u07:
## Type of change

Bug fix

## Files changed

- `run_kraken.py`
- `tests/test_run_kraken.py`

## Description

Three spots in `main()` access `config["telemetry"]` directly. If the YAML configuration does not contain a `telemetry:` section, all three crash with `KeyError` before the program even gets a chance to run any scenarios.

This is inconsistent with the rest of the file, which uses `get_yaml_item_value(config, key, default)` everywhere else to handle missing keys gracefully — for example when reading `kraken`, `tunings`, `performance_monitoring`, and `elastic` sections.

The three vulnerable lines:

| Line | Code |
|------|------|
| 173 | `config["telemetry"].get("api_url", "")` |
| 174 | `config["telemetry"].get("enabled", True)` |
| 199 | `config["telemetry"].get("archive_path")` |

All three assume `config["telemetry"]` already exists. A config file that omits the telemetry block (which is optional for local or CI runs) would bring down the entire run with a `KeyError`.

## What changed

Added a single line before the first telemetry access:

```python
config["telemetry"] = get_yaml_item_value(config, "telemetry", {})
```
This ensures config["telemetry"] always exists as at least an empty dictionary. All downstream .get() calls and bracket assignments (archive_path, run_tag, telemetry_group, prometheus_backup, etc.) work correctly whether or not the config includes the telemetry block.

Testing

Wrote a test that feeds main() a config with everything except the telemetry section. Before the fix it raises KeyError. After the fix it returns -1 (no kubeconfig — expected early exit) without crashing.
Existing zone_outage and shut_down tests also pass with the change.

- Self-review
- Unit tests added
- Existing tests pass


paigerube14 (commented):
this change looks good to me, are you able to address the copilot comments

Sahil-u07:
> this change looks good to me, are you able to address the copilot comments

yes.

PR #1459, opened 2026-07-09
fix: ingress network chaos ip commands broken by bash -c argument splitting

### LINKED ISSUE #1380: [BUG] Ingress network chaos: `ip -br addr show` returns help text in debug pod, causing interface verification to fail on ROSA/OVNKubernetes
## Summary

When running an ingress network chaos scenario on ROSA (OpenShift 4.19.32) with OVNKubernetes, the interface verification step in `ingress_shaping.py` fails because `ip -br addr show` executed inside the krkn debug pod returns the `ip` help/usage text instead of the expected interface list.

## Environment

| Field | Value |
|---|---|
| OpenShift version | 4.19.32 |
| Cloud | AWS ROSA |
| CNI | OVNKubernetes |
| krkn image | `quay.io/krkn-chaos/krkn:tools` |
| Node OS | Red Hat Enterprise Linux CoreOS 9.6.20260520-0 (Plow) |
| Node arch | amd64 / m5.xlarge |

## Steps to Reproduce

1. Configure an ingress network chaos scenario targeting a specific node interface:
   ```yaml
   - id: network_chaos
     config:
       node_interface_name: {ip-10-0-1-136.ec2.internal: [ens5]}
       execution_type: parallel
       network_params: {latency: 418ms, bandwidth: 973mbit}
       test_duration: 120
       wait_duration: 30
       image: quay.io/krkn-chaos/krkn:tools
   ```
2. Run the scenario against the cluster.

## Observed Behavior

The scenario fails with:

```
ERROR Interface ens5 not found in node ip-10-0-1-136.ec2.internal interface list
['Usage:', 'ip', 'where', 'tunnel', 'netns', 'OPTIONS', '-h[uman-readable]',
 '-f[amily]', '-4', '-l[oops]', '-o[neline]', '-rc[vbuf]']
```

Full traceback from the scenario log:
```
Exception: Interface ens5 not found in node ip-10-0-1-136.ec2.internal interface list
['Usage:', 'ip', 'where', 'tunnel', 'netns', 'OPTIONS', '-h[uman-readable]',
 '-f[amily]', '-4', '-l[oops]', '-o[neline]', '-rc[vbuf]']

  File "ingress_shaping.py", line 716, in network_chaos
    node_interface_dict = get_node_interfaces(...)
  File "ingress_shaping.py", line 345, in get_node_interfaces
    node_interface_dict[node] = verify_interface(...)
  File "ingress_shaping.py", line 275, in verify_interface
    raise Exception(...)
```

## Root Cause

In `ingress_shaping.py`, `verify_interface()` (line ~258) runs:

```python
cmd = ["ip", "-br", "addr", "show"]
output = kubecli.exec_cmd_in_pod(cmd, pod_name, "default")

interface_ip = output.split("\n")
node_interface_list = [
    interface.split()[0] for interface in interface_ip[:-1]
]
```

On this cluster, `ip -br addr show` inside the debug pod returns the `ip` command's help/usage text rather than the expected brief-format interface list. The parser then splits that help text by newline and takes the first token of each line, producing the nonsensical list `['Usage:', 'ip', 'where', ...]` instead of `['lo', 'ens5', ...]`.

## Likely Causes

Two possible explanations:

1. **The debug pod does not have `hostNetwork` access** on ROSA/OVNKubernetes (likely due to SCC restrictions in the `default` namespace), so `ip -br addr show` only sees the pod's own network namespace — or fails entirely and falls back to printing help text.

2. **The `ip` binary in `quay.io/krkn-chaos/krkn:tools` does not support the `-br` flag** on this RHEL CoreOS 9.6 host environment, causing it to print usage instead.

The `if not output` guard at line ~244 only catches a completely empty response — it does not detect the case where `ip` exits 0 but prints help text, so the bad output silently flows into the parser.

## Expected Behavior

`ip -br addr show` should return something like:
```
lo               UNKNOWN        127.0.0.1/8
ens5             UP             10.0.1.136/24
```

And the parsed list should be `['lo', 'ens5']`, allowing verification to succeed.

## Suggested Fix

- Add a sanity check after parsing: if the resulting list contains tokens like `Usage:` or starts with `-`, log a clear error and exit early rather than raising a confusing "interface not found" exception.
- Investigate whether the debug pod spec needs explicit `hostNetwork: true` and the required SCC/SecurityContext to function on ROSA.
- Consider falling back to `ip addr show` (without `-br`) if the brief flag is unavailable, or detect the `-br` support before using it.

rh-rahulshetty:
## Summary

Fixes the root cause of `ip -br addr show` returning help/usage text in the ingress network chaos scenario on ROSA/OVNKubernetes.

### Root cause

`exec_cmd_in_pod` wraps commands with `bash -c` when no `base_command` is specified. Passing `["ip", "-br", "addr", "show"]` produces `["bash", "-c", "ip", "-br", "addr", "show"]` — due to `bash -c` semantics, only the first argument after `-c` (`"ip"`) is treated as the command string; the rest become unused positional parameters. This caused bare `ip` to run with no arguments, printing help/usage text instead of interface data.

The egress scenario was unaffected because it uses `base_command="chroot"` which bypasses `bash -c` wrapping.

## Changes

Passes all 3 `ip` commands in `ingress_shaping.py` as single strings so `bash -c` treats each as one complete command:

- `get_default_interface`: `["ip", "r"]` → `["ip r"]`
- `verify_interface` (empty list path): `["ip", "r"]` → `["ip r"]`
- `verify_interface` (interface check): `["ip", "-br", "addr", "show"]` → `["ip -br addr show"]`

## Test plan

- [x] Reproduced issue on ROSA 4.19.36 / OVNKubernetes / m5.xlarge with `ens5` interface — confirmed `ip` help text in error on `main` branch (exit code 1)
- [x] Confirmed PR #1388 alone still fails — help text detected, but fallback `ip addr show` has the same `bash -c` bug (exit code 1)
- [x] Verified this fix resolves the issue — scenario runs to completion with `output_id: "success"` (exit code 0)

Closes #1380
Supersedes #1388

### DIFF
--- a/krkn/scenario_plugins/native/network/ingress_shaping.py
+++ b/krkn/scenario_plugins/native/network/ingress_shaping.py
@@ -185,7 +185,7 @@ def get_default_interface(node: str, pod_template, kubecli: KrknKubernetes, imag
     kubecli.create_pod(pod_body, "default", 300)
     pod_name = f"fedtools-{pod_name_regex}"
     try:
-        cmd = ["ip", "r"]
+        cmd = ["ip r"]
         output = kubecli.exec_cmd_in_pod(cmd, pod_name, "default")
 
         if not output:
@@ -238,7 +238,7 @@ def verify_interface(
     pod_name = f"fedtools-{pod_name_regex}"
     try:
         if input_interface_list == []:
-            cmd = ["ip", "r"]
+            cmd = ["ip r"]
             output = kubecli.exec_cmd_in_pod(cmd, pod_name, "default")
 
             if not output:
@@ -254,7 +254,7 @@ def verify_interface(
             input_interface_list = [default_route.split()[4]]
 
         else:
-            cmd = ["ip", "-br", "addr", "show"]
+            cmd = ["ip -br addr show"]
             output = kubecli.exec_cmd_in_pod(cmd, pod_name, "default")
 
             if not output:


rh-rahulshetty:
## Verification on ROSA 4.19.35 / OVNKubernetes

Tested all 3 states against a live ROSA cluster (2x m5.xlarge workers, OVNKubernetes CNI, RHEL CoreOS 9.6).

### Scenario config used

```yaml
- id: network_chaos
  config:
    node_interface_name:
      ip-10-0-1-142.ec2.internal:
        - ens5
    execution_type: parallel
    network_params:
      latency: 100ms
      bandwidth: 500mbit
    test_duration: 30
    wait_duration: 60
    image: quay.io/krkn-chaos/krkn:tools
```

### Results comparison

| Run | Key output | Exit code |
|---|---|---|
| **main branch** | \`ERROR Interface ens5 not found in node ... interface list ['Usage:', 'ip', 'where', ...]\` | **1** (fail) |
| **PR #1388 only** | \`WARNING ip -br addr show did not return a valid interface list... falling back to ip addr show\` → \`ERROR Could not determine the interface list\` | **1** (fail) |
| **This PR** | Interface verified, tc filters applied, \`output_id: "success"\` | **0** (pass) |

### Root cause

\`exec_cmd_in_pod\` wraps commands with \`bash -c\` when \`base_command\` is not specified. Passing \`["ip", "-br", "addr", "show"]\` produces \`["bash", "-c", "ip", "-br", "addr", "show"]\`. Due to \`bash -c\` semantics, only the first argument after \`-c\` (\`"ip"\`) is treated as the command string — the rest become unused positional parameters. So bare \`ip\` runs with no arguments, printing help text.

PR #1388's fallback to \`ip addr show\` had the same bug (\`["ip", "addr", "show"]\` → also only runs \`ip\`).

The egress scenario was never affected because it uses \`base_command="chroot"\` which bypasses \`bash -c\` wrapping entirely.

### Successful run log (this PR)

```
2026-07-09 18:14:22,501 [INFO] Running NativeScenarioPlugin: ['pod_network_scenarios', 'ingress_node_scenarios'] -> build/issue_1380_scenario.yml
2026-07-09 18:14:22,665 [INFO] Starting Ingress Network Chaos
2026-07-09 18:14:23,968 [INFO] Creating pod to query interface on node ip-10-0-1-142.ec2.internal
2026-07-09 18:14:29,095 [INFO] Deleting pod to query interface on node
2026-07-09 18:14:35,101 [INFO] Creating 1 virtual interfaces on node ip-10-0-1-142.ec2.internal using a pod
2026-07-09 18:14:38,496 [INFO] Deleting pod used to create virtual interfaces
2026-07-09 18:14:41,058 [INFO] Executing sleep 30;tc qdisc add dev ens5 handle ffff: ingress;tc filter add dev ens5 parent ffff: protocol ip u32 match u32 0 0 action mirred egress redirect dev ifb0; tc qdisc add dev ifb0 root netem delay 100ms  rate 500mbit ;  tc qdisc ls dev ifb0 ; sleep 30; tc qdisc del dev ifb0 root ;tc qdisc del dev ens5 handle ffff: ingress; sleep 20; tc qdisc ls dev ifb0 ; on node ip-10-0-1-142.ec2.internal
2026-07-09 18:14:41,878 [INFO] Waiting for parallel job to finish
2026-07-09 18:16:17,217 [INFO] Deleting all virtual interfaces on node ip-10-0-1-142.ec2.internal
2026-07-09 18:16:21,410 [INFO] Deleting jobs(if any)
2026-07-09 18:16:23,070 [INFO] {"output_id": "success", "output_data": {"filter_direction": "ingress", "test_interfaces": {"ip-10-0-1-142.ec2.internal": ["ens5"]}, "network_parameters": {"latency": "100ms", "bandwidth": "500mbit"}, "execution_type": "parallel"}}
2026-07-09 18:16:31,035 [INFO] No error logs collected during chaos run
2026-07-09 18:16:31,040 [INFO] Successfully finished running Kraken. UUID for the run: 092af3bf-b385-4565-b5e1-5903fa03ee91. Report generated at kraken.report. Exiting
```

ddjain:
@rh-rahulshetty Thanks for the clear root-cause analysis, the bash -c splitting issue looks right, and I could reproduce it on both AWS and GCP with cluster-bot.

Would you mind keeping this to the minimal fix (["ip -br addr show"] / ["ip r"] as single strings) and dropping the help-text parser + fallback? The help text is just bare ip output from the split args, so once the full command is passed correctly those extras aren’t needed.

@paigerube14 thoughts?

ddjain (approved):

rh-rahulshetty:
@ddjain Makes sense, let me update it.

rh-rahulshetty:
@ddjain Updated and tested successfully on ROSA clusterbot.

paigerube14 (approved):

PR #1478, opened 2026-07-13
fix: deterministic pod kills with exclude_label in pod disruption scenario

### LINKED ISSUE #1477: [BUG]: Pod Disruption randomly kills fewer pods than config.kill when exclude_label is used
# Bug Description

## **Describe the bug**

The `PodDisruptionScenarioPlugin` randomly fails to kill the requested number of pods when `exclude_label` is used. 
Currently, the plugin fetches all pods, checks if `len(pods) < config.kill`, and then shuffles the list. During the `for i in range(config.kill)` loop, if the randomly selected pod happens to have the `exclude_label`, the script logs "Excluding... from chaos" and skips it. However, it **does not** advance the loop or try to find a replacement pod. 
As a result, the chaos run is completely non-deterministic. If you request `kill: 5` and there are excluded pods in the pool, the script might only kill 2 or 3 pods depending on how the array was randomly shuffled, but it will still exit successfully as if it processed all 5. Furthermore, the `len(pods) < config.kill` check happens *before* exclusions are applied, which can falsely pass even if there aren't enough valid pods left.

## **To Reproduce**

1. Deploy 5 pods that match a common label selector (e.g., `app=my-app`).
2. Add a special label `chaos-exclude=true` to 3 of those 5 pods.
3. Setup a scenario config with `kill: 3` and `exclude_label: "chaos-exclude=true"`.
4. Run Kraken. You will notice that it almost never kills 3 pods. It will usually only kill 1 or 2 pods based on random chance.

### Scenario File
Scenario file(s) that were specified in your config file (can be starred (*) with confidential information )
```yaml
pod_disruption:
  - config:
      name_pattern: '.*'
      namespace_pattern: 'my-test-namespace'
      label_selector: 'app=my-app'
      exclude_label: 'chaos-exclude=true'
      kill: 3
      duration: 10
      timeout: 120
```

### Config File 
-  (Using standard config/config.yaml for testing pod disruption scenario)

## **Expected behavior**

The scenario should guarantee that exactly config.kill pods are deleted (in this example, exactly 3 pods). To achieve this, the script should filter out any pods matching exclude_label from the initial pods array before checking the array bounds and before entering the target selection loop.

## **Krkn Output**

INFO: 2026-07-14 00:00:00,000 - pod_disruption_scenario_plugin - Excluding my-app-pod-1 from chaos
INFO: 2026-07-14 00:00:00,100 - pod_disruption_scenario_plugin - Deleting pod my-app-pod-2
INFO: 2026-07-14 00:00:00,200 - pod_disruption_scenario_plugin - Excluding my-app-pod-3 from chaos
(In this run where `kill: 3` was requested, only successfully killed 1 pod because 2 iterations landed on excluded pods).

### **Why this happens (Step-by-Step Example):**
Assume we have 5 pods (A, B, C, D, E), `exclude_label` matches pods A and B, and we set `kill: 3`.
**1. Finding Pods & Exclusions:**
```python
pods = self.get_pods(...) # Gets all 5 pods [A, B, C, D, E]
# ... 
exclude_pods.add(pod[0]) # Identifies [A, B] as excluded

**2. The First Mistake (Validation bypass):**
```
python
if len(pods) < config.kill: # 5 < 3 is False. 

#The system currently checks the array before removing excluded pods, falsely believing there are plenty of valid targets.
```
**3. The Second Mistake (Target Selection):**

```
python

random.shuffle(pods) # Example: [B, D, A, C, E]
for i in range(3):
    pod = pods[i]
    if pod[0] in exclude_pods:
        logging.info(f"Excluding {pod[0]}...") 
    else:
        kubecli.delete_pod(pod[0], pod[1])
```

Because excluded pods are still mixed into the shuffled array, the loop consumes its iterations on them without replacing them.

Iteration 1 grabs Pod B (Excluded -> Skipped)
Iteration 2 grabs Pod D (Valid -> Killed)
Iteration 3 grabs Pod A (Excluded -> Skipped)
Result: The scenario successfully finishes, but instead of the requested 3 pods, only 1 pod was actually killed.

### PR DESCRIPTION
**Pull request by varun-ai69**
# Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization

# Description  
Fixes a non-deterministic logic bug in the `PodDisruptionScenarioPlugin`. Previously, if an `exclude_label` was provided, the plugin identified the excluded pods but failed to remove them from the target `pods` array. This caused the random target selector to occasionally land on excluded pods, log a skip, and fail to fulfill the requested `config.kill` count, silently resulting in fewer dead pods than the scenario demanded. 
This PR applies a strict list-comprehension filter to strip `exclude_pods` right after identifying them, natively guaranteeing that `len(pods) < config.kill` works accurately and that the disruption loop targets the exact number of valid pods.

- Related Issue #: Fixes #1477
- Closes #: Fixes #1477

# Documentation  

- N/A

- [x] Ensure the changes and proposed solution have been discussed in the relevant issue and have received acknowledgment from the community or maintainers. See [contributing guidelines](https://krkn-chaos.dev/docs/contribution-guidelines/)
See [testing your changes](https://krkn-chaos.dev/docs/developers-guide/testing-changes/) and run on any Kubernetes or OpenShift cluster to validate your changes

*REQUIRED*:

```
```bash
python -m coverage run -a -m unittest tests/test_pod_disruption_scenario_plugin.py -v
...

test_get_scenario_types (tests.test_pod_disruption_scenario_plugin.TestPodDisruptionScenarioPlugin.test_get_scenario_types)
Test get_scenario_types returns correct scenario type ... ok
test_killing_pods_not_enough_pods (tests.test_pod_disruption_scenario_plugin.TestPodDisruptionScenarioPlugin.test_killing_pods_not_enough_pods) ... ERROR:root:Not enough pods match the criteria, expected 4 but found only 3 pods
ok
test_killing_pods_with_exclude_label (tests.test_pod_disruption_scenario_plugin.TestPodDisruptionScenarioPlugin.test_killing_pods_with_exclude_label) ... ok

Ran 3 tests in 0.015s

OK 


```

### DIFF
--- a/krkn/scenario_plugins/pod_disruption/pod_disruption_scenario_plugin.py
+++ b/krkn/scenario_plugins/pod_disruption/pod_disruption_scenario_plugin.py
@@ -227,6 +227,7 @@ class PodDisruptionScenarioPlugin(AbstractScenarioPlugin):
                 _exclude_pods = self.get_pods("",config.exclude_label,config.namespace_pattern, kubecli, field_selector="status.phase=Running", node_label_selector=config.node_label_selector, node_names=config.node_names)
                 for pod in _exclude_pods:
                     exclude_pods.add(pod[0])
+                    logging.info(f"Excluding pod {pod[0]} from chaos")
 
             expected_pod_count = len(pods)
             pods = [pod for pod in pods if pod[0] not in exclude_pods]
--- a/krkn/scenario_plugins/pod_disruption/pod_disruption_scenario_plugin.py
+++ b/krkn/scenario_plugins/pod_disruption/pod_disruption_scenario_plugin.py
@@ -228,9 +228,9 @@ class PodDisruptionScenarioPlugin(AbstractScenarioPlugin):
                 for pod in _exclude_pods:
                     exclude_pods.add(pod[0])
 
+            expected_pod_count = len(pods)
             pods = [pod for pod in pods if pod[0] not in exclude_pods]
 
-            pods_count = len(pods)
             if len(pods) < config.kill:
                 logging.error("Not enough pods match the criteria, expected {} but found only {} pods".format(
                         config.kill, len(pods)))
@@ -243,7 +243,7 @@ class PodDisruptionScenarioPlugin(AbstractScenarioPlugin):
                 logging.info(f'Deleting pod {pod[0]}')
                 kubecli.delete_pod(pod[0], pod[1])
             
-            return_val = self.wait_for_pods(config.label_selector,config.name_pattern,config.namespace_pattern, pods_count, config.duration, config.timeout, kubecli, config.node_label_selector, config.node_names)
+            return_val = self.wait_for_pods(config.label_selector,config.name_pattern,config.namespace_pattern, expected_pod_count, config.duration, config.timeout, kubecli, config.node_label_selector, config.node_names)
         except Exception as e:
             raise(e)
 
--- a/krkn/scenario_plugins/pod_disruption/pod_disruption_scenario_plugin.py
+++ b/krkn/scenario_plugins/pod_disruption/pod_disruption_scenario_plugin.py
@@ -228,6 +228,7 @@ class PodDisruptionScenarioPlugin(AbstractScenarioPlugin):
                 for pod in _exclude_pods:
                     exclude_pods.add(pod[0])
 
+            pods = [pod for pod in pods if pod[0] not in exclude_pods]
 
             pods_count = len(pods)
             if len(pods) < config.kill:
@@ -239,11 +240,8 @@ class PodDisruptionScenarioPlugin(AbstractScenarioPlugin):
             for i in range(config.kill):
                 pod = pods[i]
                 logging.info(pod)
-                if pod[0] in exclude_pods:
-                    logging.info(f"Excluding {pod[0]} from chaos")
-                else:
-                    logging.info(f'Deleting pod {pod[0]}')
-                    kubecli.delete_pod(pod[0], pod[1])
+                logging.info(f'Deleting pod {pod[0]}')
+                kubecli.delete_pod(pod[0], pod[1])
             
             return_val = self.wait_for_pods(config.label_selector,config.name_pattern,config.namespace_pattern, pods_count, config.duration, config.timeout, kubecli, config.node_label_selector, config.node_names)
         except Exception as e:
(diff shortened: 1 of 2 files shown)


krkn/scenario_plugins/pod_disruption/pod_disruption_scenario_plugin.py
```diff
@@ -228,6 +228,7 @@ def killing_pods(self, config: InputParams, kubecli: KrknKubernetes):
                 for pod in _exclude_pods:
                     exclude_pods.add(pod[0])
 
+            pods = [pod for pod in pods if pod[0] not in exclude_pods]
```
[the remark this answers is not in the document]
varun-ai69: Thanks for the review. You're absolutely right `wait_for_pods` queries against `label_selector` directly, which returns the full un-excluded population, so passing the post-exclusion length would cause a timeout deadlock when the cluster restored.

I've just pushed a fix for this. The logic now captures `expected_pod_count` = `len(pods)` before filtering the exclusions, while the kill-bounds check correctly remains on the filtered subset array. I also updated the unit tests with mock assertions to strictly ensure `wait_for_pods` receives the original cluster parameter size.

varun-ai69:
hey @paigerube14 @tsebastiani @chaitanyaenr @ddjain can u please review the pr whenever you are free

varun-ai69:
@paigerube14 Hi! I looked into the failing CI. It seems the failure is in the `test_customapp_pod `functional test. From what I can see, that scenario doesn't use `exclude_label`, so it doesn't appear to exercise the code path modified in this PR. The unit tests, including the new test for the exclusion logic, pass locally. Could you please confirm whether this looks like an unrelated CI failure or if there's something in my change that I should investigate further? Thanks!

varun-ai69:
Hi @paigerube14! Just following up on this. I wanted to check if you've had a chance to determine whether this is an unrelated CI failure or if there's anything in my changes that you'd like me to investigate further. Happy to make any updates if needed. Thanks!

krkn/scenario_plugins/pod_disruption/pod_disruption_scenario_plugin.py
```diff
@@ -239,13 +240,10 @@ def killing_pods(self, config: InputParams, kubecli: KrknKubernetes):
             for i in range(config.kill):
                 pod = pods[i]
                 logging.info(pod)
-                if pod[0] in exclude_pods:
-                    logging.info(f"Excluding {pod[0]} from chaos")
```
paigerube14, line 243: it would be nice if we could still somehow list pods that were excluded, maybe in the for loop on line 228-229

varun-ai69: Thanks for the feedback! I've updated the implementation to log each pod excluded by exclude_label during the exclusion loop.

PR #1493, opened 2026-07-20
Retry other PVC-mounted pods when selected pod is unusable

### LINKED ISSUE #1489: PvcScenarioPlugin exits on first unusable pod
## Description

`PvcScenarioPlugin` currently selects a single pod associated with a PVC using `random.choice(pvc.podNames)`.

If the selected pod is not usable (for example, it is in `ImagePullBackOff`, `CrashLoopBackOff`, or its container cannot execute commands), the plugin immediately exits with an error without attempting any of the other pods that are mounted to the same PVC.

As a result, the chaos scenario fails even when another healthy pod attached to the same PVC could have successfully executed the experiment.

## Steps to Reproduce

1. Create a PVC.
2. Attach the PVC to multiple pods:
   - One healthy/running pod.
   - One unusable pod (e.g. `ImagePullBackOff`).
3. Configure `PvcScenarioPlugin` with the PVC name.
4. Run the PVC scenario.
5. If `random.choice(pvc.podNames)` selects the unusable pod first, the plugin exits immediately.

## Expected Behavior

The plugin should attempt to execute the chaos scenario using another pod mounted to the same PVC before failing.

The scenario should only fail if **all** pods associated with the PVC are unusable.

## Actual Behavior

The plugin selects one pod at random.

If the selected pod is unusable, the scenario fails immediately without trying the remaining pods associated with the PVC.

Example error:

```text
PvcScenarioPlugin container 'app' in pod 'broken-pod' has no usable shell (/bin/bash or /bin/sh); cannot execute commands inside the container
```

### PVC configuration

Image

### Pods attached to PVC

Image

### Scenario config

```yaml
pvc_scenario:
  pvc_name: kraken-test-pvc
  pod_name: kraken-test-pod
  namespace: kraken
  fill_percentage: 98
  duration: 10s
```
### Logs showing the plugin selecting the unusable pod and exiting without attempting the other available pod

Image

rohIta-k:
# Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization

# Description

This PR improves the reliability of the PVC scenario plugin. Instead of picking a single random pod and failing immediately if it's unusable, the plugin now loops through the `pods_to_try` list until it finds a healthy pod with a working shell to execute the scenario

## Proposed Changes

- **Looping Through Pods:** Refactored the pod selection into a `for candidate_pod_name in pods_to_try:` loop. 
- **Handling Shell Failures:** If `get_pod_shell()` returns `None`, the code logs a warning and runs `continue` to try the next available pod instead of returning `1` right away.
- **Fixed PVC Flag Tracking:** Replaced  `if "pvc" not in locals():` check with a clean boolean flag named `found_valid_pvc`. This properly tracks the volume state for the current pod iteration and prevents stale loop data from causing runtime crashes.

### Test Functions Added
* Extended `TestPvcScenarioPluginRun` by introducing structural test coverage targeting loop mechanics:
  - `test_run_retries_pods_and_succeeds`: Checks that if one or more candidate pods are unusable, the code skips them and successfully uses the first available usable pod.
  - `test_run_fails_when_all_pods_lack_shell`: Makes sure the plugin returns an error code if none of the available pods have a working shell.
  - `test_run_skips_pod_without_pvc_and_continues`: Guarantees the loop safely skips over any candidate pod that is missing a valid PVC mapping without breaking.
  - `test_run_pod_name_only_fails_when_unusable`: Ensures that when only a `pod_name` is provided, the plugin returns an error code if that single targeted pod is unusable.

### PVC Scenario Logs

Screenshot 2026-07-20 at 10 36 01 PM

### Unit Test Results
Screenshot 2026-07-20 at 10 36 55 PM

## Related Issues
- Related Issue #1489 
- Closes #1489   

- [x] Ensure the changes and proposed solution have been discussed in the relevant issue and have received acknowledgment from the community or maintainers. See [contributing guidelines](https://krkn-chaos.dev/docs/contribution-guidelines/)
See [testing your changes](https://krkn-chaos.dev/docs/developers-guide/testing-changes/) and run on any Kubernetes or OpenShift cluster to validate your changes

### DIFF
--- a/krkn/scenario_plugins/pvc/pvc_scenario_plugin.py
+++ b/krkn/scenario_plugins/pvc/pvc_scenario_plugin.py
@@ -110,6 +110,7 @@ class PvcScenarioPlugin(AbstractScenarioPlugin):
                     pods_to_try = [pod_name]
 
                 shell = None
+                configured_pvc_name = (pvc_name or "").strip()
                 for candidate_pod_name in pods_to_try:
                     logging.info("Attempting to use pod '%s'",candidate_pod_name)
                     # Get volume name
@@ -125,14 +126,20 @@ class PvcScenarioPlugin(AbstractScenarioPlugin):
 
                     found_valid_pvc = False
                     for volume in pod.volumes:
-                        if volume.pvcName is not None:
-                            volume_name = volume.name
-                            pvc_name = volume.pvcName
-                            pvc = lib_telemetry.get_lib_kubernetes().get_pvc_info(
-                                pvc_name, namespace
-                            )
-                            found_valid_pvc = True
-                            break
+                        
+                        if volume.pvcName is None:
+                            continue
+                        # If a PVC was configured, only match that PVC
+                        if configured_pvc_name and volume.pvcName != configured_pvc_name:
+                            continue
+                        
+                        volume_name = volume.name
+                        pvc_name = volume.pvcName
+                        pvc = lib_telemetry.get_lib_kubernetes().get_pvc_info(
+                            pvc_name, namespace
+                        )
+                        found_valid_pvc = True
+                        break
                     if not found_valid_pvc:
                         logging.warning(
                             "PvcScenarioPlugin Pod '%s' in namespace '%s' does not use a pvc"
--- a/krkn/scenario_plugins/pvc/pvc_scenario_plugin.py
+++ b/krkn/scenario_plugins/pvc/pvc_scenario_plugin.py
@@ -74,7 +74,7 @@ class PvcScenarioPlugin(AbstractScenarioPlugin):
                         "PvcScenarioPlugin You must specify the namespace where the PVC is"
                     )
                     return 1
-                if pvc_name is None and pod_name is None:
+                if not pvc_name and not pod_name:
                     logging.error(
                         "PvcScenarioPlugin You must specify the pvc_name or the pod_name"
                     )
@@ -95,7 +95,7 @@ class PvcScenarioPlugin(AbstractScenarioPlugin):
                     pvc = lib_telemetry.get_lib_kubernetes().get_pvc_info(
                         pvc_name, namespace
                     )
-                    if not pvc.podNames:
+                    if pvc is None or not pvc.podNames:
                         logging.error(
                             "PvcScenarioPlugin Pod associated with %s PVC, on namespace %s, "
                             "not found" % (str(pvc_name), str(namespace))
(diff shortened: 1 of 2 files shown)


rohIta-k:
Hi @paigerube14 ! I’ve updated the PR to address all the feedback from augment, specifically making the final error messages much more accurate for both pod and PVC paths and adding a new test case to test single pod fallback path.

Aryanbhargava18:
Great refactor for handling pod shell failures and retrying appropriately! 

While looking through the changes, I did notice a subtle edge-case bug that was technically present in the original code, but becomes much more dangerous with this retry loop.

In the volume iteration block, we overwrite the `pvc_name` variable with the first PVC found on the pod:
```python
                    found_valid_pvc = False
                    for volume in pod.volumes:
                        if volume.pvcName is not None:
                            volume_name = volume.name
                            pvc_name = volume.pvcName  # Overwrites the configured pvc_name!
                            pvc = lib_telemetry.get_lib_kubernetes().get_pvc_info(
```
**The Problem:**
If the scenario explicitly targets a specific PVC via `pvc_name: "my-target-pvc"` in the config, but the candidate pod happens to mount multiple PVCs (e.g. `my-log-pvc` and `my-target-pvc`), this loop blindly matches the *first* PVC in `pod.volumes`. It will overwrite `pvc_name` to `my-log-pvc`, effectively pointing the chaos injection to the wrong volume!

**Recommendation:**
Since you're already refactoring this validation block, I highly recommend saving the original `pvc_name` from the config and checking against it to ensure we don't accidentally fill the wrong volume:

```python
                configured_pvc_name = get_yaml_item_value(scenario_config, "pvc_name", "")
                # ...
                
                # Inside the loop:
                    found_valid_pvc = False
                    for volume in pod.volumes:
                        if volume.pvcName is not None:
                            # If we were given a specific PVC, only match that one
                            if configured_pvc_name and volume.pvcName != configured_pvc_name:
                                continue
                                
                            volume_name = volume.name
                            pvc_name = volume.pvcName
                            pvc = lib_telemetry.get_lib_kubernetes().get_pvc_info(
                                pvc_name, namespace
                            )
                            found_valid_pvc = True
                            break
```
This ensures that when a pod mounts multiple PVCs, the chaos is injected into the exact PVC the user requested!

rohIta-k:
@Aryanbhargava18 Thanks for pointing this out! I updated the logic to handle both cases while preserving the existing behavior.

- **When `pvc_name` is provided:** the configured `pod_name` is ignored, and `pods_to_try` is populated with the pods associated with the configured PVC. For each candidate pod, the plugin now selects the **configured PVC** instead of simply picking the first mounted PVC.

- **When `pvc_name` is not provided:** the configured `pod_name` is used as the only candidate in `pods_to_try`. Since there is no configured PVC to match, the plugin falls back to selecting the **first PVC mounted on that pod**, which is the same behavior as before.

This fixes the issue where a pod mounting multiple PVCs could previously result in the scenario operating on the wrong PVC.

rohIta-k:
@paigerube14 Can you review the pr?

paigerube14:
Few blockers before merging: 

1. pvc_scenario_plugin.py:103 — Regression introduced by the refactor: if not pvc.podNames: has no guard for get_pvc_info returning None. The original try/except caught the AttributeError and logged a meaningful "PVC not found" message. Now it falls through to the generic handler. Fix: check if pvc is None or not pvc.podNames:.
2. pvc_scenario_plugin.py:144 — The break inside the inner for vol in container.volumeMounts loop doesn't exit the outer for container in pod.containers loop. Last matching container wins instead of first, so exec_cmd_in_pod targets the wrong container. Fix: use a flag or next()+generator to break both loops.
3. pvc_scenario_plugin.py:118 — Validation checks if pvc_name is None and pod_name is None, but get_yaml_item_value returns '' as default (not None). Both empty strings pass the check, and pods_to_try = [''] sends a blank pod name to the API instead of the explicit config error. Fix: check if not pvc_name and not pod_name.

rohIta-k:
@paigerube14 I've made the requested changes.

PR #1499, opened 2026-07-21
add label via slash workflow call

tejugang:
# Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization

# Description  
This calls the workflow that allows users to add any label that exists in the repo to an issue or PR using /label-name.
- ensures that only admins or users with "write" privileges can add labels.
- if triage/accepted label is added via '/triage/accepted' comment, then needs-triage label is automatically removed


.github/workflows/slash-to-add-label.yml
```diff
@@ -0,0 +1,13 @@
+name: Label via slash
```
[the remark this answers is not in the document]
paigerube14: @tejugang does this need .yml/.yaml extension?

.github/workflows/slash-to-add-label.yml
```diff
@@ -0,0 +1,13 @@
...
+
+on:
+  issue_comment:
```
paigerube14, line 4: do we need a pull-request path as well?

tejugang: don't think that is necessary, this call works for PR comments too!

PR #1530, opened 2026-07-29
changing setuptools version

tejugang:
# Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization

# Description  
The pkg_resources module was removed in setuptools version 82.0.0 and above, so a change was made in requirements.txt to install a version of setuptools<82.0.0, and the force install of 81.0.0 is removed in the test workflow.


paigerube14 (approved):
/lgtm

PR #694, opened 2024-09-12
Core Refactoring, Krkn Scenario Plugin API

tsebastiani:
# main goals:

- Introduce an API for the Krkn Scenario Plugins
- Introduce a dynamic plugin loader for all the classes implementing the API
- Convert all the existing scenarios in classes that implement the API
- Define a strict naming convention for Scenario Plugins validated by the loader to give to the comunity a clear guideline on how to implement new scenarios (MVC Frameworks approach)
- reduce the run_krkn.py line numbers and complexity
- minimize the code changes in the run_krkn.py to introduce a new scenario (plugins are loaded dynamically)
- Produce a clear documentation on how to develop new Krkn scenarios
- Unit test the scenarios

# main benefits:
- speed up the on-boarding on scenario development (community and new team members)
- create more traction around the project
- increase the test coverage
- streamline the codebase to increase readability

PR #1540, opened 2026-08-05
add back the manifestwork template the managed cluster scenarios render

Sahil-u07:
## Type of change

Bug fix

## Files changed

- `krkn/scenario_plugins/managed_cluster/manifestwork.j2`
- `tests/test_managed_cluster_scenario_plugin.py`

## Description

All four managed cluster scenarios (start, stop, start_klusterlet, stop_klusterlet) crash at runtime with `jinja2.exceptions.TemplateNotFound: manifestwork.j2` before anything is created in the cluster.

The scenarios render the template through `FileSystemLoader` pointing at the plugin directory, but the file was dropped when the plugin moved from `kraken/managedcluster_scenarios` into `krkn/scenario_plugins` during the refactor in #694. It was never re-added in the new location.

The template is restored from the old layout with the same content, so the rendered ManifestWork (ClusterRole, RoleBindings, ServiceAccount, Job) is unchanged from before the refactor.

## Testing

Added a test that the template file exists in the plugin directory and that rendering it with the variables the scenarios pass produces a valid ManifestWork document. All 18 tests in the managed cluster test file pass.

### DIFF
--- a/krkn/scenario_plugins/managed_cluster/manifestwork.j2
+++ b/krkn/scenario_plugins/managed_cluster/manifestwork.j2
@@ -1,0 +1,67 @@
+apiVersion: work.open-cluster-management.io/v1
+kind: ManifestWork
+metadata:
+  namespace: {{managedcluster_name}}
+  name: managedcluster-scenarios-template
+spec:
+  workload:
+    manifests:
+      - apiVersion: rbac.authorization.k8s.io/v1
+        kind: ClusterRole
+        metadata:
+          name: scale-deploy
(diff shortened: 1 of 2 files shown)


Sahil-u07:
Hi, this is a small regression fix. The manifestwork.j2 template was dropped when the managed cluster plugin moved during the refactor in #694, so all managed cluster scenarios have been failing at runtime since. The PR description references that. Also fixed an invalid namespace on the ClusterRole while restoring it.

PR #1549, opened 2026-08-11
fix: do not report a skipped health check plugin as started

### LINKED ISSUE #1537: Improve health check startup logging
## Description

Currently, `HealthCheckFactory.start_all()` creates the plugin, calls `run_health_check()`, and then always logs that the plugin has started.

However, the validation to determine whether a health check should actually run is currently performed inside run_health_check(). If the required configuration is missing, the plugin logs that it is skipping and returns immediately, but the factory still logs that the health check plugin has started, making the startup logs misleading.

For example, the logs currently look like this:

```text
2026-08-04 18:43:32,200 [INFO] HTTP health check config is not defined, skipping
2026-08-04 18:43:32,200 [INFO] Started health check plugin 'http_health_check' reading from config key 'health_checks'

2026-08-04 18:43:32,201 [INFO] kubevirt checks config namespace is not defined, skipping them
2026-08-04 18:43:32,201 [INFO] Started health check plugin 'virt_health_check' reading from config key 'kubevirt_checks'
```

This is misleading because the plugin has already decided not to run, but the factory still reports that it has started successfully.

## Proposed Solution

Move the configuration validation currently performed at the beginning of each `run_health_check()` implementation into a new abstract method in `AbstractHealthCheckPlugin`.

For example:

```python
@abstractmethod
def can_run(self, config: dict[str, Any]) -> bool:
    """Return True if the plugin has sufficient configuration to start."""
```

Each health check plugin would implement `can_run()` using the validation logic that currently exists in `run_health_check()`, logging the reason before returning `False` if the plugin should be skipped.

Then `HealthCheckFactory.start_all()` can validate the plugin before attempting to start it:

```python
plugin = self.create_plugin(...)

if not plugin.can_run(plugin_config):
    continue

if plugin.manages_own_threads():
    ...
else:
    ...

logging.info(
    f"Started health check plugin '{plugin_type}' "
    f"reading from config key '{config_key}'"
)
```

dchaudhari7177:
Closes #1537.

## What

`start_all()` created each plugin, called `run_health_check()`, then always logged that it had started. Because the "can I run?" decision lived *inside* `run_health_check()`, a plugin that bailed out immediately still got reported as started.

`can_run(config)` is added to `AbstractHealthCheckPlugin` and `start_all()` gates on it, so an unrunnable plugin is never started and never reported as started. The skip reason still gets logged, from `can_run()`.

## Before / after

Same config both times (a `health_checks:` section present but with no URL), captured by running `start_all()`:

**Before (`main`):**
```
2026-08-11 21:56:59,912 [INFO] HTTP health check config is not defined, skipping
2026-08-11 21:56:59,912 [INFO] Started health check plugin 'http_health_check' reading from config key 'health_checks'
--> start_all returned 1 checker(s)
```

**After (this PR):**
```
2026-08-11 21:56:58,355 [INFO] HTTP health check config is not defined, skipping
--> start_all returned 0 checker(s)
```

Note the second line is gone *and* the plugin is no longer in the returned checker list, so nothing downstream waits on a thread that did nothing.

## One deviation from the issue

The issue proposes `can_run()` as an `@abstractmethod`. **I made it a concrete method returning `True`** instead, because this is a dynamically-loaded plugin architecture: an abstract method breaks every out-of-tree plugin at load time, and `SimpleHealthCheckPlugin` has no configuration prerequisites to implement. `manages_own_threads()` on the same class already uses exactly this shape — concrete, sensible default, overridden only where it matters.

Say the word if you'd rather have it abstract and I'll switch it.

`VirtHealthCheckPlugin._initialize_from_config` keeps its own early returns. It's reachable independently of the factory, and it validates further than `can_run()` should (cluster connectivity, VMI discovery) — work that shouldn't happen before the plugin starts. `can_run()` only mirrors the namespace check, which is the one that produced the misleading log.

## Tests

5 added to `tests/test_health_check_factory.py`: the default is `True`; the HTTP plugin's yes/no cases; `start_all` returns no checkers and logs no "Started" line for a skipped plugin; and a runnable plugin is still started and still logged.

With `krkn/health_checks/` reverted to `main`: **1 failure + 3 errors**. With the change: all 15 pass.

## Checks run locally

| gate | before | after |
| --- | --- | --- |
| `unittest tests.test_health_check_factory` | 10 passed | **15 passed** |
| `unittest discover -s tests` | 885 tests, 1 failure + 150 errors | 890 tests, **same** 1 failure + 150 errors |

The 150 errors and the `test_junit_utils` failure are identical on pristine `main` here — they're missing cloud SDKs (`requirements.txt` pulls AWS/Azure/GCP/IBM/Aliyun, which I couldn't install on this machine). I installed `krkn-lib` alone into a clean venv to run the health-check tests, so this change is exercised; the rest is CI's.

## Unrelated bug found while testing

Worth a separate issue, I think. **A second `HealthCheckFactory()` in the same process gets an empty `config_key_map`**, so its `start_all()` silently starts nothing:

```python
a = HealthCheckFactory(); print(a.config_key_map)
# {'health_checks': 'http_health_check', 'simple_health_checks': 'simple_health_check'}
b = HealthCheckFactory(); print(b.config_key_map)
# {}
```

`loaded_plugins` is a class attribute, so `__load_plugins` short-circuits on the second construction and never repopulates the per-instance `config_key_map`. My tests set `config_key_map` explicitly rather than depend on construction order, which is why they don't trip over it. Happy to file it, or fix it in a follow-up if you'd prefer.

### DIFF
--- a/krkn/health_checks/abstract_health_check_plugin.py
+++ b/krkn/health_checks/abstract_health_check_plugin.py
@@ -106,6 +106,24 @@ class AbstractHealthCheckPlugin(ABC):
         """
         pass
 
+    def can_run(self, config: dict[str, Any]) -> bool:
+        """
+        Indicates whether this plugin has sufficient configuration to start.
+
+        The factory calls this before starting the plugin, so a plugin that would
+        immediately return from ``run_health_check()`` is never started and never
+        reported as started. A plugin that decides not to run should log the reason
+        here before returning ``False``.
+
+        Returns ``True`` by default, so plugins with no configuration prerequisites
+        (and any out-of-tree plugin written before this method existed) keep working
+        unchanged.
+
+        :param config: the plugin's own section of config.yaml
+        :return: True if the plugin should be started; False to skip it
+        """
+        return True
+
     def manages_own_threads(self) -> bool:
         """
         Indicates whether this plugin spawns and manages its own worker threads internally.
--- a/krkn/health_checks/health_check_factory.py
+++ b/krkn/health_checks/health_check_factory.py
@@ -137,7 +137,7 @@ class HealthCheckFactory:
                     iterations=iterations,
                     **kwargs
                 )
-                if not plugin.can_run(plugin_config):
+                if not self.__can_run(plugin, plugin_config, plugin_type):
                     continue
                 if plugin.manages_own_threads():
                     tq = queue.SimpleQueue()
@@ -160,6 +160,30 @@ class HealthCheckFactory:
                     f"Health check plugin '{plugin_type}' not found, skipping"
                 )
         return checkers
+
+    @staticmethod
+    def __can_run(plugin, plugin_config: dict, plugin_type: str) -> bool:
+        """
+        Asks a plugin whether it can run, treating a raising ``can_run()`` as "no".
+
+        ``can_run()`` inspects raw config, so a malformed section (wrong type, or
+        wrong entry types within a list) can raise out of it. It is called here on
+        the main thread, so an uncaught exception would abort Kraken startup for
+        every scenario rather than skipping the one misconfigured plugin.
+
+        :param plugin: the instantiated health check plugin
+        :param plugin_config: the plugin's section of config.yaml
+        :param plugin_type: the plugin's health check type, for logging
+        :return: True if the plugin should be started; False otherwise
+        """
+        try:
+            return plugin.can_run(plugin_config)
+        except Exception as e:
+            logging.warning(
+                f"Health check plugin '{plugin_type}' failed to validate its "
+                f"configuration, skipping: {e}"
+            )
+            return False
 
     def __load_plugins(self, base_class: Type):
         """
--- a/krkn/health_checks/health_check_factory.py
+++ b/krkn/health_checks/health_check_factory.py
@@ -137,6 +137,8 @@ class HealthCheckFactory:
                     iterations=iterations,
                     **kwargs
                 )
+                if not plugin.can_run(plugin_config):
+                    continue
                 if plugin.manages_own_threads():
                     tq = queue.SimpleQueue()
                     plugin.run_health_check(plugin_config, tq)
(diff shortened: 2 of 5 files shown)


rohIta-k:
@dchaudhari7177  I noticed that can_run() currently doesn't catch exceptions. If validation itself fails, the plugin shouldn't be treated as runnable or let the exception abort the entire Kraken startup. This should be handled by catching the exception at the factory level.

dchaudhari7177:
@rohIta-k good catch — you're right, and it's a regression this PR introduced rather than a pre-existing gap. Fixed at the factory level as you suggested.

Before this change the url check lived inside `run_health_check()`, which for non-self-threading plugins runs on a worker thread, so a raise there killed one thread. Moving it into `can_run()` put it on the main thread inside `start_all()`, where only `HealthCheckPluginNotFound` was caught — so a malformed `health_checks` section (a list of strings instead of dicts, say) would `AttributeError` out of `start_all()` and abort Kraken startup for every scenario.

`HealthCheckFactory.__can_run()` now wraps the call and treats a raising `can_run()` as "cannot run": the plugin is skipped and the reason logged, same shape as the existing `HealthCheckPluginNotFound` handling. I kept the catch at the factory rather than inside each plugin so out-of-tree plugins get it too.

Also addressed the second qodo/augment finding on the same commit: `test_start_all_still_starts_a_runnable_plugin` was starting the real polling loop, doing live DNS/HTTP against `example.com` and only exiting on the next interval tick. `run_health_check` is patched out there now — the factory's gate is what that test covers, not the plugin's request loop.

New test `test_start_all_survives_a_plugin_whose_can_run_raises` is red before the fix (`AttributeError: 'str' object has no attribute 'get'`) and green after. Full file: 16/16 passing locally against `krkn-lib`.

On the assignment bot: #1537 is currently unassigned — happy to be assigned it, or to close this if someone else has picked it up.

dchaudhari7177:
Rebased onto current `main` and added the missing `Signed-off-by` trailer to both commits, which should clear the DCO check. No code changed — the diff for `krkn/health_checks/` and `tests/test_health_check_factory.py` is byte-identical to the reviewed version; the rebase only picked up upstream commits landed since.

`tests/test_health_check_factory.py`: 16 passed.

The remaining red check is FOSSA **License Compliance** ("6 issues found"), which reports against `refs/branch/main` and flags nothing this PR touches — it's showing the same 6 issues on main. Not something I can act on from a fork, so flagging rather than guessing.

PR #1571, opened 2026-08-16
fix(rollback): evaluate Version fields per-instance (fixes #1488)

### LINKED ISSUE #1488: [BUG] Version dataclass evaluates timestamp and hash_suffix once at import time, not per instance
## Describe the bug

In `krkn/rollback/config.py`, the `Version` dataclass defines `timestamp` and `hash_suffix` as plain class-level defaults:

```python
@dataclass(frozen=True)
class Version:
    scenario_type: str
    rollback_context: RollbackContext
    timestamp: int = time.time_ns()         # ← runs once at module import
    hash_suffix: str = get_random_string(8) # ← runs once at module import
```

In Python, dataclass field defaults that are not wrapped in field(default_factory=...) are evaluated exactly once — when the class body is first executed (at import time). Every subsequent call to Version.new_version(...) returns an instance with the same frozen timestamp and hash_suffix that were computed when the module was first imported.

This means all rollback filenames generated within a single krkn process are identical:
```python
hog_scenarios_1719000000000000000_abcdefgh.py  ← first registration
hog_scenarios_1719000000000000000_abcdefgh.py  ← second registration (same file)
```
This directly contradicts the field docstrings, which say "Get current timestamp in nanoseconds" and "Generate a random string" — the clear intent is per-instance freshness.

**To Reproduce:**
```python
from krkn.rollback.config import Version, RollbackContext

ctx = RollbackContext("run-uuid-abc")

v1 = Version.new_version("hog_scenarios", ctx)
v2 = Version.new_version("hog_scenarios", ctx)

print(v1.timestamp == v2.timestamp)     # True — should be False
print(v1.hash_suffix == v2.hash_suffix) # True — should be False
print(v1.version_file_name == v2.version_file_name)  # True — collision
```
Likely impact hogs_scenario_plugin.py spawns one thread per target node, and each thread calls self.rollback_handler.set_rollback_callable(...). Because all threads share the same RollbackHandler and Version.new_version() always returns the same filename, every thread is likely writing to the same rollback file path — each overwriting the previous registration.

If krkn is interrupted (SIGTERM, Ctrl+C, crash) while multiple hog pods are running, only the last thread to write has its rollback data on disk. Earlier pods lose their rollback registration silently — no error is logged, the system reports successful serialization, but those pods cannot be rolled back.

Note: whether the full file path actually collides depends on path construction in version_file_full_path. However, the default values are provably incorrect regardless — a stale import-time timestamp is never the right default for a versioning field.

Expected behavior: Each Version.new_version() call should evaluate time.time_ns() and get_random_string(8) fresh, producing unique identifiers per registration.

**Proposed fix:**
```python
from dataclasses import dataclass, field

@dataclass(frozen=True)
class Version:
    scenario_type: str
    rollback_context: RollbackContext
    timestamp: int = field(default_factory=time.time_ns)
    hash_suffix: str = field(default_factory=lambda: get_random_string(8))
```

**Additional context:**
1. Affected file: krkn/rollback/config.py lines 254–259
2. Caller: krkn/rollback/handler.py line 268 (set_rollback_callable)
3. Multi-call trigger: krkn/scenario_plugins/hogs/hogs_scenario_plugin.py lines 99, 160–166
4. Existing tests call set_rollback_callable only once, so the collision path is not currently tested
5. PR #1204 mentioned fixing dataclass default_factory issues but was never merged (needs-rebase, needs-dco labels, still open)
6. This is separate from issue #1404 (which covers silent exception swallowing in serialization)

I would be happy to open a PR with this fix and the associated tests if this approach looks good to you!!!

swagatobauri:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization

## Description

This PR fixes the issue where the `timestamp` and `hash_suffix` fields in the `Version` dataclass (`krkn/rollback/config.py`) were evaluated exactly once at module import time. By wrapping these assignments in `dataclasses.field(default_factory=...)`, the rollback identifiers are now dynamically evaluated per instance. This prevents silent collisions and rollback overwrites when `Version.new_version()` is called multiple times concurrently (e.g., by the hogs scenario plugin).

- Closes #1488

## Documentation

*(No documentation needed as this is an internal engine bug fix)*

N/A

*REQUIRED:*

```text
python -m unittest discover -s tests -v

...
test_version_uniqueness (test_rollback.TestVersionDataclass.test_version_uniqueness) ... ok
...
Ran 79 tests in 8.423s

OK

### DIFF
--- a/tests/test_rollback.py
+++ b/tests/test_rollback.py
@@ -397,3 +397,21 @@ class TestSecureTempDirectories:
         """When user provides an explicit path, no fallback is triggered."""
         explicit_path = "/some/user/chosen/path"
         assert explicit_path  # truthy, so no fallback
+
+import unittest
+
+class TestVersionDataclass(unittest.TestCase):
+    """Tests for the Version dataclass in rollback configuration."""
+
+    def test_version_uniqueness(self):
+        """Verify that Version instances evaluate dynamic fields per-instance."""
+        from krkn.rollback.config import Version, RollbackContext
+        import time
+        ctx = RollbackContext("test-uuid")
+        v1 = Version.new_version("test_scenario", ctx)
+        time.sleep(0.001)
+        v2 = Version.new_version("test_scenario", ctx)
+
+        self.assertNotEqual(v1.timestamp, v2.timestamp)
+        self.assertNotEqual(v1.hash_suffix, v2.hash_suffix)
+        self.assertNotEqual(v1.version_file_name, v2.version_file_name)
--- a/tests/test_rollback.py
+++ b/tests/test_rollback.py
@@ -3,6 +3,8 @@ import logging
 import os
 import sys
 import tempfile
+import time
+import unittest
 import uuid
 import subprocess
 
@@ -11,7 +13,7 @@ from krkn_lib.ocp import KrknOpenshift
 from krkn_lib.telemetry.ocp import KrknTelemetryOpenshift
 from krkn_lib.models.telemetry import ScenarioTelemetry
 from krkn_lib.utils import SafeLogger
-from krkn.rollback.config import RollbackConfig
+from krkn.rollback.config import RollbackConfig, Version, RollbackContext
 
 sys.path.append(
     os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by swagatobauri** [comment]
Hi @paigerube14! The PR is ready for review. 

I've fixed the bug by using `field(default_factory=...)` so the timestamps and hashes generate uniquely for every instance instead of getting stuck at import time. I also added a unit test to verify it, and resolved all of the AI bot's feedback regarding docstrings and inheritance.

Could you please approve the workflows so the CI tests can run? Thanks!

File: tests/test_rollback.py [hunk 1]
```diff
@@ -397,3 +397,21 @@ def test_explicit_path_is_respected(self):
...
         assert explicit_path  # truthy, so no fallback
+
+import unittest
+
+class TestVersionDataclass(unittest.TestCase):
+    """Tests for the Version dataclass in rollback configuration."""
+
+    def test_version_uniqueness(self):
+        """Verify that Version instances evaluate dynamic fields per-instance."""
+        from krkn.rollback.config import Version, RollbackContext
```
paigerube14: lets put all imports at the top of the file

swagatobauri:
Hi @paigerube14 ,
Just checking in on this PR to keep it active before the bot closes it next week. I've incorporated the feedback from the last review. 
Please let me know if there is anything else needed from my end. 
Thank you.

paigerube14 (approved):

PR #1588, opened 2026-09-02
ci: sign released krkn images with cosign

tsebastiani:
## Summary

Signs the released **krkn base image** with cosign on every tag build, so the images can be cryptographically verified at runtime by **krknctl** and **krkn-operator** through the shared `pkg/verify` package.

Part of krkn-chaos/krknctl#180 (the verification side — shared `pkg/verify` + mandatory enforcement in the krknctl run path — is in krkn-chaos/krknctl#182). This PR delivers the **signing** half of that end-to-end trust chain.

## What changed (`.github/workflows/docker-image.yml`, `manifest` job)

- **Install cosign** (pinned `sigstore/cosign-installer@6f9f177` / v4.1.2).
- After pushing the multi-arch manifest lists, **sign each by its manifest-list digest** — for both `quay.io/krkn-chaos/krkn` and the `quay.io/redhat-chaos/krkn` mirror.
- Signing is folded into the same step as the manifest push (per registry) so it runs **while that registry's credentials are still active** — both accounts live on `quay.io`, so the second `docker login` overwrites the first in the docker config; the cosign signature artifact (`sha256-<digest>.sig`) must be pushed to the same repo as the image.
- **Sign by digest, not tag** (anti-TOCTOU): consumers resolve the tag → this digest and find the signature. `latest` and `${TAG}` reference the same per-platform digests, so they share one manifest-list digest — signing it covers both tags.
- **Key-based, `--tlog-upload=false`** — matches the ecosystem's fully-offline verification policy (no Fulcio/Rekor dependency, air-gap friendly).
- **Self-verification**: each signature is immediately verified against the public half of the signing key (derived on the fly via `cosign public-key`, no extra secret), so a broken signing setup fails the release instead of shipping unverifiable images.

## Required repository secrets (must be added before the next tag)

| Secret | Value |
|--------|-------|
| `COSIGN_PRIVATE_KEY` | the `cosign.key` produced by `cosign generate-key-pair` (encrypted PEM) |
| `COSIGN_PASSWORD` | the password protecting that private key |

The matching **public key** (`cosign.pub`) must replace the placeholder currently embedded in krknctl's `pkg/verify/cosign.pub`, so both consumers trust this signing key.

## Scope / follow-ups
- This PR signs the **krkn base image** only. The **krkn-hub scenario images** are built by a separate action/repo (`redhat-chaos/actions/krkn-hub`) and will need the same signing step there.
- Generate the ecosystem key pair, store the secrets here, and open the krknctl PR that swaps the placeholder public key for the real one.

## Testing
- `docker-image.yml` validated with `actionlint` (clean) and YAML-parsed.
- End-to-end signing exercises only on a real tag push (needs the secrets); the in-step `cosign verify` self-check gates that.


tsebastiani:
## Addressed qodo finding: "Unsigned releases survive failures" (High)

Fixed in the latest commit by reordering the manifest job to a **stage → sign → verify → promote** flow, so consumer-facing tags are never published pointing at an unsigned/unverifiable digest.

For both `quay.io/krkn-chaos/krkn` and the `quay.io/redhat-chaos/krkn` mirror:

1. The multi-arch manifest list is pushed to a **staging tag only** (`${TAG}-signing`) — the public `${TAG}` and `latest` do not exist yet.
2. `cosign sign` + `cosign verify` run against that **digest** (anti-TOCTOU, key-based, `--tlog-upload=false`).
3. Only **after** verification succeeds is the exact signed digest **promoted** to `${TAG}` and `latest` via `docker buildx imagetools create --tag <repo>:<tag> <repo>@<digest>` (retag by digest, no rebuild).

If the key is missing, the password is wrong, the registry errors, or verification fails, the job fails **before** any public release tag is created — so no unsigned release is ever exposed. A `Set up Docker Buildx` step was added to the `manifest` job for `imagetools`.

Note: the `${TAG}-signing` staging tag is intentionally left behind; it points to the same signed digest, so it is harmless and itself verifiable.

paigerube14 (approved):

